PR #71, opened 2015-04-28
Added optional timeout to WebRequest

Bomberlt:
I have added timeout to WebRequestExtensions.WriteBody() to abort sending request if there is no response after timeout.

### DIFF
--- a/src/Hl7.Fhir.Core/Rest/WebRequestExtensions.cs
+++ b/src/Hl7.Fhir.Core/Rest/WebRequestExtensions.cs
@@ -17,8 +17,6 @@ using System.Threading.Tasks;
 
 namespace Hl7.Fhir.Rest
 {
-    using System.Diagnostics;
-
     public static class WebRequestExtensions
     {
         public static void WriteBody(this WebRequest request, byte[] data)
@@ -122,10 +120,12 @@ namespace Hl7.Fhir.Rest
                 });
 
             var async = req.BeginGetResponse(callback, null);
-            
+
             if (!async.IsCompleted)
             {
+#if !PORTABLE45
                 ThreadPool.RegisterWaitForSingleObject(async.AsyncWaitHandle, new WaitOrTimerCallback(TimeoutCallback), req, req.Timeout, true);
+#endif
 
                 //async.AsyncWaitHandle.WaitOne();
                 // Not having thread affinity seems to work better with ManualResetEvent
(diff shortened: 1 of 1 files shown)


ewoutkramer:
Hi Bomberlt, can you make this patch work under PCL4.5 too?  If you open the Portable45.sln file, you'll see this causes compiler issues under the Portable Class Library. Need this to work before I can merge this change...

Bomberlt:
Added #if !PORTABLE45 directive for fixing the issue with PCL4.5 build. I hope it's ok.

ewoutkramer:
Ok, merged. This means mobile clients won't have the timeout support, but we'll not invest too much time since this may all change for the new Win10 codebase anyway.

Bomberlt:
Good :)
By the way, when the next version of DSTU1 nuget is planned? Maybe you can build new package and release it?

ewoutkramer:
Oops….so busy with DSTU1 that I have not put any thought into releasing something new for DSTU1.  Should not be too hard I guess

Bomberlt:
@ewoutkramer sorry to nag you so much, but can you release new nuget for DSTU1?

ewoutkramer:
Pushed as 0.11.4!

Bomberlt:
@ewoutkramer are you sure you pushed right (DSTU1) brach to nuget? It seems that some files are missing from DSTU1 in nuget package.

Also are you going to merge develop-dstu1 to master-dstu1?

ewoutkramer:
Done!

Bomberlt:
Seems like newest DSTU1 nuget have DSTU2 sources. Are you sure you made nuget from master-dstu1 branch and NOT from master branch?

ewoutkramer:
I actually built from the develop-dstu1 branche…. What’s wrong exactly?

Bomberlt:
After upgrading package to version 0.11.4 we receive these errors:
>The type or namespace name 'FhirRequest' could not be found (are you missing a using directive or an assembly reference?)	
>The type or namespace name 'UriLinkEntry' could not be found (are you missing a using directive or an assembly reference?)
>The type or namespace name 'UriLinkList' could not be found (are you missing a using directive or an assembly reference?)

FhirRequest was pressent at DSTU1 version of lib, but at DSTU2 it has disapeared as with UriLinkEntry and UriLinkList.

ewoutkramer:
Found the problem. I used the right sources but then I did a DEBUG build, so nuget packed the wrong assemblies ☹

ewoutkramer:
Ok, updated again. The version of the library should be 0.11.5….

PR #152, opened 2016-01-07
Remove hard-coded c:/temp path from unit tests

tiloc:
Two of the unit tests included hard-coded references to c:/temp. As my machine has its temp directory elsewhere and there is a good .Net API to find the correct location, I have replaced the hard-coded references with code to correctly obtain the temp directory (and also use the Path.Combine API to append the filenames to it).


tiloc:
Did additional reading on proper Git working model. Will properly branch, etc., then resubmit.

ewoutkramer:
Yes, please!

PR #155, opened 2016-01-14
Accept unknown enums when AcceptUnknownLiterals is configured.

tiloc:
Due to a small mismatch in specification age between FHIR .Net API and a commercial FHIR implementation there arose the need to make FHIR client a bit more tolerant when unknown enums were encountered. I have opted to enable/disable this behavior through the existing AcceptUnknownMembers switch.

### DIFF
--- a/src/Hl7.Fhir.Core/Serialization/PrimitiveValueReader.cs
+++ b/src/Hl7.Fhir.Core/Serialization/PrimitiveValueReader.cs
@@ -36,9 +36,9 @@ namespace Hl7.Fhir.Serialization
         internal object Deserialize(Type nativeType)
         {
             if (nativeType == null) throw Error.ArgumentNull("nativeType");
-                 
+
             object primitiveValue = _current.GetPrimitiveValue();
-            
+
             if (nativeType.IsEnum() && primitiveValue.GetType() == typeof(string))
             {
                 var enumMapping = _inspector.FindEnumMappingByType(nativeType);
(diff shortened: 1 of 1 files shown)


ewoutkramer:
That's subtle. Do we expand the notion of "expand unknown members" to "treat a known enumerated member with an incorrect value as unknown" ?

tiloc:
True, subtle. I was seeing the "AcceptUnknownMembers" switch more as a "be tolerant" switch, as in Postel's law. I think it would be overreaching a bit to introduce additional "AcceptSomething" switches for each scenario where more tolerance is required in real-world integrations. And renaming the switch to be more generic in naming would break an existing API. So based on these considerations, I finally opted for expanding the scope of the existing switch a bit.

ewoutkramer:
I have started a discussion about this on the FHIR Skype Implementers Chat. Are you following that chat?

tiloc:
Not yet. Where would I find that chat? I have a Skype client.

magohl:
I think this is the chat mentioned

https://chats.fhir.me/feeds/skype/implementers.html

I am also interested in a way to handle unknown enums.

ewoutkramer:
Yes, this is the read-only log of the Skype chat, but if you want to participate you'd need to join that chat. Unfortunately, it's full (Microsoft limits the number of participants to 299)... But just in case, send me a Skype invite (user ewoutkramer) and I'll add you when we have capacity again.

tiloc:
I am closing this pull request. I assume the solution for issue #167 will look considerably different.

ewoutkramer:
Indeed, see #203!

PR #156, opened 2016-01-15
More complete checking of HTTP response codes.

tiloc:
In Requester, there is an additional place where HTTP status code needs to be checked. Otherwise, Format Exceptions will occur later. Since there are now 2 places where the status code will have to be checked, I am proposing to also carve the conversion from HTTP status code to an exception into its own method.

### DIFF
--- a/src/Hl7.Fhir.Core/Rest/Requester.cs
+++ b/src/Hl7.Fhir.Core/Rest/Requester.cs
@@ -57,39 +57,20 @@ namespace Hl7.Fhir.Rest
 
             LastResult = doRequest(interaction);
             var status = LastResult.Response.Status;
-            HttpStatusCode statusCode;
-            if (!Enum.TryParse<HttpStatusCode>(status, out statusCode))
-            {
-                // If the status code is unable to be parsed, then report
-                // in internal server error
-                statusCode = HttpStatusCode.InternalServerError;
-            }
+            var outcome = LastResult.Resource as OperationOutcome;
 
-            if (status.StartsWith("2"))      // 2xx codes - success
-            {
-                return LastResult;
-            }
-            else if (status.StartsWith("3") || status.StartsWith("1"))      // 3xx codes - we don't handle them, unless the .NET API did it for us
-            {
-                throw Error.NotSupported("Server returned a status code '{0}', which is not supported by the FhirClient".FormatWith(status));
-            }
-            else if (status.StartsWith("4") || status.StartsWith("5"))      // 4xx/5xx codes - client or server error.
-            {
-                var message = String.Format("Operation was unsuccessful, and returned status {0}.", status);
+            Exception httpException = HttpStatusToException(status, outcome);
 
-                var outcome = LastResult.Resource as OperationOutcome;
-                if (outcome != null)
-                {
-                    throw new FhirOperationException(message + " OperationOutcome: " + outcome.ToString(), statusCode, outcome);
-                }
-                else
-                    throw new FhirOperationException(message, statusCode);
+            if (httpException == null)
+            {
+                return LastResult; // Everything good, return result
             }
             else
             {
-                throw Error.NotSupported("Server returned an illegal http status code '{0}', which is not defined by the Http standard".FormatWith(status));
+                throw httpException;   // Http request failed, throw exception
             }
         }
+
 
         private Bundle.EntryComponent doRequest(Bundle.EntryComponent interaction)
         {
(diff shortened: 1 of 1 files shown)


ewoutkramer:
I do not completely understand the change: the checking is done after doRequest() is called, and Execute() is the only method calling doRequest(). So, I don't see why there are two places where the code needs to be checked?  In fact, Execute() will now do the check twice, once in doRequest() and once in Execute(). Moreover, the semantics of doRequest() have been changed: it's important that the AfterResponse hook is called *even if there's a non 2xx statuscode*, and you now throw an exception before the AfterResponse is called...

tiloc:
I have to provoke the situation again and look into the call trace. It made sense the day I introduced the code. I had a situation where responses from a FHIR server which clearly had a status code 500 were still passed into the FHIR parser and thus resulted in an exception, as the response contained HTML coded error messages from the server. I analyzed the call trace back then and chose the place in the code based on my analysis. But that was 3 weeks ago (had to learn GIT and pull requests prior to submitting this :-))
I agree with your point about changed AfterResponse behavior. That was definitely unintentional and I would fix this once I reestablish why the duplicate check is required and works where I placed it.

tiloc:
I have setup a faulty FHIR server to always return status 500 and recreated the situation, using the code as found in your repository (i.e. prior to my patch). The offending line in Requester.cs is:

*// Do this call after AfterResponse, so this will be called, even if exceptions are thrown by ToBundleEntry()*
                    return webResponse.**ToBundleEntry(inBody);**

This calls ToBundleEntry on a response which is status 500 with a body of type text/html. ToBundleEntry chokes on the HTML error text and will throw a FormatException which is never caught. I.e. while Requester.Execute seems to be designed to throw a FhirOperationException on error 500, in reality it passes through an uncaught FormatException. 

I still think my code should not be pulled by you, for the reasons which you stated. But I think it would still be better if the response code was checked prior to attempting to create a bundle from it.

tiloc:
Next try. I have amended the pull request with a new commit. Now it does the error checking only once. After the AfterResponse has been invoked, and before trying to convert the response into Bundle. Any error will cleanly throw FhirOperationException, a success will create a Bundle.

ewoutkramer:
Nice, thanks. Will merge.

ewoutkramer:
Tilo, I was responding to a question from Sean Moore from Epic on the  mailinglist. I noticed that in our new code, if a status 4xx or 5xx is returned, but the body contains a valid (OperationOutcome) resource, the operation outcome is lost. So, while we can now safely handle returned bodies with HTML on a 4xx and 5xx, a 4xx/5xx with details in an OperationOutcome will get lost :-(  Need to fix this!

brianpos:
And should add a unit test to verify that we catch this next time.

ewoutkramer:
It's very hard to unit-test. Indeed, Tilo had to create a faulty server to test!  Who wants to take a look at it?
You want to do it Tilo, or shall I?

tiloc:
I break it, I fix it :-)  I wasn't aware of that particular possibility, as none of the sandbox servers I have tried so far return that combination. I wouldn't know how to provide a unit-test due to lack of a server which exhibits this behavior. Any recommendation for a URL that returns a 4xx/5xx error with an OperationOutcome?

brianpos:
I'm happy to contribute the unit test.

ewoutkramer:
Yes, most servers do that when you query for a non existing resource,

Grahame: http://fhir2.healthintersections.com.au/open/Patient/doesnotexist
Mine: http://spark-dstu2.furore.com/fhir/Patient/doesnotexist

You'll get a 404 with an OperationOutcome!

tiloc:
Excellent, I can work with that. So integration test-wise we can have a test case with 404 with OperationOutcome, and 500 with HTML.

tiloc:
Using Grahame's server for testing now. Ironically, Ewout's server gives me XML body in Internet Explorer and on Curl, and an HTML body on FhirClient. Weird...

tiloc:
See latest Pull Request. Code fix + Integration test.

ewoutkramer:
>Ewout's server gives me XML body in Internet Explorer and on Curl, and an HTML body on FhirClient. Weird...

so, I think this is because you used http://spark.furore.com as the endpoint (which returns a HTML page saying there's nothing at that address), instead of http://spark.furore.com/fhir?

tiloc:
Hi Ewout, actually I later found I had accidentally placed the "Patient/doesnotexist" into both the fhirclient constructor and into the patient retrieval, which put it into the URL twice with the mentioned results. 
The line which you commented gives me exactly what I want: an error 500 with Html body. Error 500 due to misconfigured webserver. 
All this is considered in my pull request. The servers I use are returning the results I need for the tests. Will eventually add another variety I found with Cerner: error 404 with json mime type and empty result.

PR #163, opened 2016-01-30
Fix for pull request #156

tiloc:
Unfortunately, I had overlooked that due to the rebuilt code, the LastResult was not being properly set anymore.
I have fixed the issue and added a dedicated integration test for status 500 response handling.

P.S.: I am wondering whether the Lastxxx properties in Requester.cs should exist? One could safely share the FhirClient between multiple threads otherwise, but it seems the Lastxxx properties will overwrite each other when FhirClient is invoked from multiple threads. But I guess that's a discussion for another day...


brianpos:
Agree with the threading issue.
Would need to have another way to handle the last result  where it  isn't the resource and am operation outcome is returned. I.e. Prefer minimal

PR #164, opened 2016-01-31
Amendment to PR #162: Reenable VB case ambiguity warnings.

tiloc:
Hl7.Fhir.Specification is fully CLSCompliant. I.e. assembly flagged as CLSCompliant true and no need to suppress any warnings.

Hl7.Fhir.Core: Mark the 2 offending classes specifically as non-compliant. Only suppress warning CS3016, as CS3015 no longer occurs. Mark overall assembly as CLSCompliant to receive warnings if VB case ambiguity is accidentally introduced.


brianpos:
Thanks, I found it as we use the project from some VB projects in .net 4.0 so no real choice.

PR #165, opened 2016-02-01
Make Portable45 target compilable and testable.

tiloc:
This includes:
- Include Portable.DataAnnotations package
- Remove obsolete ValidationContextFactory workaround
- Reworking the DescriptionAttribute ~~(new EnumDocumentationAttribute)~~
- Updating to JSON 8.0.2 (in an attempt to fix 64 bit integer limit, as release notes suggested this might be fixed).

Known limitations:
- portable JSON parser can only parse numbers up to 64 bit limit (EdgecaseRoundtrip test fails)
- Xhtml validation is not available in portable library (TestXhtmlValidation fails)


tiloc:
See my comment in the corresponding issue. **Look, but don't pull yet :-)**

tiloc:
I have amended the pull request based on feedback from Brian. Reverted back to a (better) version of DescriptionAttribute.

PR #173, opened 2016-02-11
Build scripts

tiloc:
This is a **look, but don't pull yet!** Pull Request for review. It has no conflicts with the current state of the develop branch.

**It does the following:**

- Consistent naming for Visual Studio Projects and configurations. All configurations are consolidated into a single solution called "Hl7.Fhir.sln" 
- Introduces a build script. Commands like "runbuild Test" can be used to build, unit test, and package NuGet packages with strong named assemblies in one fully automated go.
- Test case tagging to enable filtering of tests which rely on external servers (IntegrationTest), or run exceedingly long (LongRunner)
- Strong-naming for test binaries allows unit testing of release packages.
- AppVeyor.yml configuration file allows 5 minute setup of AppVeyor for CI


ewoutkramer:
So.....when shall we pull?  HIMSS is over...

tiloc:
Let's get it started. Have responded on Skype.

IDispose:
Does this mean that I can now switch back to using the official nuget package for .net core support? I have successfully used @tiloc 's private source and would like to switch to official channel.

ewoutkramer:
Mostly, but I think you still need the project.json file as mentioned in issue #178

tiloc:
I am not seeing the official NuGet package for 0.9.5-alpha**2** being pushed to NuGet.org yet. That will be the first package to contain my changes. Once that is out there you could remove my private channel from your NuGet.config. 
The project.json will always be required, as it is technically part of your application. You would have to change the dependency on the Fhir.Net library from "0.9.6" to "0.9.5-alpha*". Maybe later the project.json could be shorter. I have to examine whether it is possible to express some of the dependencies in the NuGet package, rather than in the project.json.

IDispose:
@ewoutkramer @tiloc Thanks. As soon as the package is available on the alpha channel, I'll switch and report back.

PR #176, opened 2016-03-02
Search parameters parsing improvements

michelemottini:
All search parameter parsing errors cause a `FormatException`. The exception message specifies the name of the parameter that caused the error.

Specifying more than once a search parameter that can have only a single value now cause a parsing error – previously there was no error and all values except the last one were ignored.

Sort orders just starting with `asc` or `desc` – e.g. `desczzz` or `ascdesc` were considered valid, now they cause an error. `ascending` and `descending` and their truncation down to `asc` and `desc` are considered valid.

Empty generic search parameters no longer cause an error, and are added ‘as is’ to the `SearchParams.Parameters` collection. This allow the caller to ignore them or generate appropriate error messages depending on their meaning.

(Based on conversations at the FHIR Connectathon 2016 in Orlando a FHIR server should not fail when a search URL contains unknown parameters with arbitrary values, and this is impossible without this last change)

Fixes https://github.com/ewoutkramer/fhir-net-api/issues/148

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/Rest/SearchParamsTests.cs
+++ b/src/Hl7.Fhir.Core.Tests/Rest/SearchParamsTests.cs
@@ -207,5 +207,167 @@ namespace Hl7.Fhir.Test.Rest
             CollectionAssert.AreEquivalent(q.Parameters.ToList(), q2.Parameters.ToList());
             CollectionAssert.AreEquivalent(q.Elements.ToList(), q2.Elements.ToList());
         }
+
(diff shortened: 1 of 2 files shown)


michelemottini:
Tests are failing because `http://fhir2.healthintersections.com.au/open/` appear to be down or unreachable

ewoutkramer:
Hi Michele, thanks for the improvements! Can you redo the pull request on develop, instead of Master?

michelemottini:
Sure, I'll do that

PR #182, opened 2016-03-17
My first pull request

mvdzel:
Hi,
The code is not 100% yet. The roundtrip test doesnot fully work yet.
Looking forward to your comments.
Michael

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ This is the support API for working with the DSTU version of [HL7 FHIR][1] on th
 
 This library provides:
 * Class models for working with the FHIR data model using POCO's
-* Xml and Json parsers and serializers
+* Xml, Json and Turtle parsers and serializers
 * A REST client for working with FHIR-compliant servers
 * Helper classes to work with the specification metadata, most notably `StructureDefinition` and generation of differentials
 
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ This library provides:
 * Xml, Json and Turtle parsers and serializers
 * A REST client for working with FHIR-compliant servers
 * Helper classes to work with the specification metadata, most notably `StructureDefinition` and generation of differentials
+DISCLAIMER: The Turtle implementation and specs are still a WIP.
 
 ## Getting Started ##
 Get started by reading the [online documentation][3] or downloading the NuGet packages:
--- a/contributors.md
+++ b/contributors.md
@@ -7,5 +7,6 @@ Ewout Kramer (Furore)
 Marten Smits (Furore)
 Christiaan Knaap (Furore)
 Martijn Harthoorn (Furore)
+Michael van der Zel (UMCG)
 
 
--- a/src/Hl7.Fhir.Core.Tests/Hl7.Fhir.Core.Tests.csproj
+++ b/src/Hl7.Fhir.Core.Tests/Hl7.Fhir.Core.Tests.csproj
@@ -110,6 +110,7 @@
     <None Include="TestData\json-edge-cases.json">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </None>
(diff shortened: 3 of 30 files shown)


tiloc:
A few things I noticed:
- use the develop branch as your baseline. It is much newer than master, and where all development is based off. 
- HtmlAgilityPack dependency hasn't been maintained in 4 years. It is not very up to date with regards to supported target platforms. 
- there is a console application included in the pull request. Maybe convert it into a unit test?

ewoutkramer:
Hi Michael,

We are using GitFlow (http://nvie.com/posts/a-successful-git-branching-model/) for working on this project, so we work on "develop" not on "master". 

And as @tiloc said, I'd love to have unit-tests, but not a separate console application. As well, I will only add new dependencies on 3rd party libraries very cautiously, since by now the library has a large user-base and I don't want to force everyone using this library to include yet more dependencies...

mvdzel:
Ok. I'll try and move to develop. Is there an easy way to make my current fork be a fork on develop?
You can ignore the console app. That is just "pruts".
As for the dependency, can we make the dependency configurable? So that you only include it when you want to experiment with turtle?

ewoutkramer:
I don't know how, but maybe @tiloc knows ;-)

tiloc:
.Net can load DLLs dynamically, based on their name. While that would technically remove the static dependency, the resulting code would be nasty (lots of reflection). The DLLs could also not be acquired through NuGet then. If I had to solve the problem cleanly, I would probably refactor the FHIR .Net library to allow "registration" of serializers for different formats (XML, JSON, Turtle, Google ProtoBuf) and then have the Turtle support and other experimental support sit in their own libraries which could be registered with the Core library.
Since I already barely have a need for 2 different formats (XML, JSON; I always configure the library to JSON as first line in my code) I would not get myself involved in such an effort.

mvdzel:
There is another option. We just keep the turtle fork and make sure it is in sync with the main release. 
I just need a pointer or steps on creating the nuget package.

tiloc:
In the build/ directory there is a PowerShell script called "runbuild.ps1". If you go to that directory and do "runbuild.ps1 Package" you will see a script execute a build and then drop several NuGet packages into a directory called "working". These would however have the exact same identifiers and version names as Ewout's official NuGets, therefore you would have to consider how to disseminate them without confusing anybody. You could also sign up to the AppVeyor build platform (free, simple) and register your fork with it. If you do this, everytime you do a commit, AppVeyor will build your fork and post the resulting NuGet packages to a private NuGet stream which your users can subscribe to.

Looking at the statements from [https://www.hl7.org/fhir/rdf.html](https://www.hl7.org/fhir/rdf.html) on the purpose of the RDF 
_"Implementer Note: The FHIR RDF format is defined to assist the process of bridging between operational data exchange and formal knowledge processing systems. While the RDF form offers a fully functional representation of FHIR resources, it has very different operational characteristics from the JSON and XML, and would be implemented for different reasons. Systems focused on operational exchange of data would not generally use RDF."_ I am wondering though, what would be the best structure for the libraries. As a commercial user of the Fhir .Net library for the purpose which is described here as "operational exchange" (i.e. mostly pulling data from commercial EMRs), I appreciate the Fhir .Net Core library being very lean, i.e. small foot-print and few dependencies. I think it is therefore already a good idea to have the support for working with the specification in its own library which depends on Core, but not vice-versa. Maybe ultimately instead of forking, you might want to have an add-on library to Core, for folks who want to build a bridge to knowledge-processing systems. Just thinking out loud.

ewoutkramer:
@tiloc, yes, agree. I think we should put this serializer in a separate assembly and distribute it separately on NuGet. Note that this is true for many extensions we are going to be making to the library (I have already split off the FhirPath stuff too), which is in line with how the .NET Core stuff seems to be organized: much more granular NuGet packages.

PR #188, opened 2016-04-06
DSTU2 -> STU3

tiloc:
This pull request touches all places where DSTU2 had to be changed to STU3.
The resulting NuGet packages will be identified as **Hl7.Fhir.STU3-0.9.5-ConnectathonMay2016**.

In order to build release NuGet packages do this:

- Place your official strong-naming key at src/FhirNetApi.snk
- In a PowerShell, change to the build/ directory, then run ".\runbuild.ps1 Package"
- NuGet files end up in working/ directory and can be published as usual


ewoutkramer:
Cool, thanks!

ewoutkramer:
That thing you did with "$dependencyId -like " we should have that in the main branch too, no?

tiloc:
The "$dependencyId -like" thing is a small improvement which makes the script less dependent on the STU-version. As long as the main branch stays on DSTU2, it will not make a difference whether you change it. And as you merge the Connectathon branch over to main branch this change will come along with the merge. So making the change on the mainline branch right now would IMHO only give you a good feeling, but no tangible benefit :-)

ewoutkramer:
Ah, right. Let's safe ourselves some work then.

PR #194, opened 2016-04-28
Modified SpecificationDirectory to start looking in  …

Gjoll:
Modified SpecificationDirectory to start looking in
Assembly.GetExecutingAssembly().Location rather than
AppDomain.CurrentDomain.BaseDirectory

AppDomain.CurrentDomain.BaseDirectory is set to a windows directory when running
visual studio unit tests, which causes validation.zip file to not be found.

### DIFF
--- a/src/Hl7.Fhir.Specification/Specification/Source/FileDirectoryArtifactSource.cs
+++ b/src/Hl7.Fhir.Specification/Specification/Source/FileDirectoryArtifactSource.cs
@@ -19,6 +19,7 @@ using System.Xml;
 using System.Diagnostics;
 using Newtonsoft.Json;
 using Hl7.Fhir.Rest;
+using System.Reflection;
 
 namespace Hl7.Fhir.Specification.Source
 {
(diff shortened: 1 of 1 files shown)


ewoutkramer:
Hi Kurt, there was probably a good reason I put it that way, i.e. to make it work while running in a web application. Bit hesitant to pull it because of that. Did you try it in a web application as well?  Somehow my VS unit test have always worked the way it is now. Strange?!

Gjoll:
Maybe things work differently East of the Atlantic ocean. ☺
I understand your hesitancy in changing it. Leave it as it is unless someone else complains.

I wonder why your unit tests worked though. Did you hand copy the validation*.zip file to the windows dir?

Oh well…

See you in Montreal?

Best,
Kurt A.

ewoutkramer:
Sure I will be in Montreal!

Let’s pull it after Montreal then, at least we have some more time to find out the consequences ;-)

tstolker:
Hi Ewout,

To chime into the same, we are running a web application with FHIR server but found the unpacking of the zip file at runtime into a windows temp directory a rather strange solution. So instead, we unpacked the zip file and deploy the files from the zip as part of our website. It was actually hard as part of DSTU1 to make the code deal with these files in our own deployment; it required copying source code and re-implementing some classes.

Best regards,

Theo.

Issue #196, opened 2016-05-03
JSON vs XML serialization differences


bnantz:
There seems to be a performance difference between XML and JSON serialization that I do not understand.  See attached unit test which shows anecdotally that JSON is about 10x slower.  The Hl7.Fhir.Core.Serialization.ComplexTypeWriter code at line 98 seems to serialize things in JSON differently to support extensions?

Any guidance would be great as we are using this in active development now.  We would be happy to investigate or help with this if we can.

[FhirSerializationLoadTest.zip](https://github.com/ewoutkramer/fhir-net-api/files/247712/FhirSerializationLoadTest.zip)

ewoutkramer:
I will soon rewrite the serializer, based on the new IElementNavigator interface, which will probably fix this. Stay tuned!

ewoutkramer:
Fixed as part of #356 (PR #646).

PR #199, opened 2016-05-06
Performance improvement for JSON serialization of FHIR primitive types.

bnantz:
This is a potential fix for Issue #196 that I filed.  For arrays of primitive types this improves serialization performance.

### DIFF
--- a/src/Hl7.Fhir.Core/Serialization/ComplexTypeWriter.cs
+++ b/src/Hl7.Fhir.Core/Serialization/ComplexTypeWriter.cs
@@ -9,15 +9,9 @@
 using Hl7.Fhir.Introspection;
 using Hl7.Fhir.Model;
 using Hl7.Fhir.Support;
-using Newtonsoft.Json.Linq;
 using System;
 using System.Collections;
-using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
-using System.Reflection;
-using System.Text;
-
 
 namespace Hl7.Fhir.Serialization
 {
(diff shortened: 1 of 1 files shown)


tiloc:
Just wondering: in isFhirPrimitive would instance have to be checked for being an empty list prior to accessing its 0 element?
And all pull requests should be against the /develop branch.

src/Hl7.Fhir.Core/Serialization/ComplexTypeWriter.cs
```diff
@@ -87,15 +81,20 @@ private void write(ClassMapping mapping, object instance, bool summary, Property
...
                 {
                     // else split up between two properties, name and _name
-                    writer.Serialize(prop,value, summary, SerializationMode.ValueElement);
+                    writer.Serialize(prop, value, summary, SerializationMode.ValueElement);
                     _writer.WriteEndProperty();
                     _writer.WriteStartProperty("_" + memberName);
-                    writer.Serialize(prop, value, summary, SerializationMode.NonValueElements);
+                    if (!isFhirPrimitive(value))
+                    {
+                        writer.Serialize(prop, value, summary, SerializationMode.NonValueElements);
```
brianpos: Without this call, extensions on primitive types will not be serialized.

bnantz: What would an extension to a primitive (e.g. boolean) mean?  Can you direct me to where in the FHIR spec I can find out more about this?

brianpos: An example use could be to indicate a reason for an Organization being inactive
 <Organization>
      <active value=false>
             <extension url="http://example.org/inactivereason">
                      <valueString value="Organization was de-registered by AHPRA"/>
             </extension>
      </active>

ewoutkramer: Yes, or specifying a precision for decimals or dates, a language for strings. The usecases are endless!

ewoutkramer:
I admit this code is confusing, and these differences should be hidden in the json writer. Tried to do that before, will look at it again.

ewoutkramer:
I will soon rewrite the serializer, based on the new IElementNavigator interface, which will probably fix this. So, reopening this one, closing the related issue.

PR #219, opened 2016-06-15
feature: implement common generic IValue<T> interface on primitive types

wmrutten:
Implement a common IValue<T> interface for primitive types, in order to allow polymorphic calls to Value getters and setters.


ewoutkramer:
Done.

PR #235, opened 2016-08-22
Feature/xmlnavigator

mharthoorn:
New XmlDom Navigator and PathNavigator
XMlDomNavigator is TypeProvider just returns "Element" etc. because that interface is still WIP.


ewoutkramer:
Has been implemented on ISerializedRepresentation instead

PR #236, opened 2016-08-22
Develop - Implement Base.Children on all FHIR model classes

wmrutten:
The initial concept implemented a Properties collection member of type IList<Base>.
However IList is not a suitable return type for PoCo classes, as the non-empty child nodes may change dynamically. So the List would need to be regenerated on each call... inefficient.

The updated implementation now implements a IEnumerable<Base> Children method.
The method only returns non-null elements.
This implies that a new, empty resource instance returns an empty Children sequence.
List members are flattened.

Also implement Children on nested child component classes.

Suppress child elements marked with representation = "xmlAttr" (Element.Id), as these members don't implement Base.


brianpos:
Assume this is just for review, and not intended to actually merge in right?

wmrutten:
Currently this is up for review, but we're planning to merge this into the API.

wmrutten:
Additionally:
- Explicitly prevent inheriting Changed extensions from external profiles
- Ensure that Base element references always return the same instance

PR #252, opened 2016-10-12
Add support Compression/Decompression using gzip/deflate

brianpos:
This adds support for the compression/decompression using gzip/deflate as discussed.
(only thing might want to not include is the change to test against my server)

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/Rest/FhirClientTests.cs
+++ b/src/Hl7.Fhir.Core.Tests/Rest/FhirClientTests.cs
@@ -33,11 +33,11 @@ namespace Hl7.Fhir.Tests.Rest
         //public static Uri testEndpoint = new Uri("http://localhost.fiddler:1396/fhir");
         //public static Uri testEndpoint = new Uri("https://localhost:44346/fhir");
         //public static Uri testEndpoint = new Uri("http://localhost:1396/fhir");
-        public static Uri testEndpoint = new Uri("http://fhir2.healthintersections.com.au/open");
+        //public static Uri testEndpoint = new Uri("http://fhir2.healthintersections.com.au/open");
         //public static Uri testEndpoint = new Uri("https://api.fhir.me");
         //public static Uri testEndpoint = new Uri("http://fhirtest.uhn.ca/baseDstu2");
         //public static Uri testEndpoint = new Uri("http://localhost:49911/fhir");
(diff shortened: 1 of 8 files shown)


src/Hl7.Fhir.Core/Rest/Requester.cs
```diff
@@ -113,12 +135,33 @@ public Bundle.EntryComponent Execute(Bundle.EntryComponent interaction)
...
         {
             if (response.ContentLength != 0)
             {
-                var body = HttpUtil.ReadAllFromStream(response.GetResponseStream());
+                byte[] body = null;
+                var respStream = response.GetResponseStream();
+#if !PORTABLE45
+                if (response.ContentEncoding == "gzip")
+                {
+                    var decompressed = new GZipStream(respStream, CompressionMode.Decompress, true);
```
ewoutkramer: I think we should use a using() statement here to do the Dispose()?

src/Hl7.Fhir.Core/Rest/WebRequestExtensions.cs
```diff
@@ -67,10 +67,23 @@ internal static void WriteBody(this HttpWebRequest request, byte[] data)
...
+            }
+            else
+            {
+                outs = request.GetRequestStream();
+            }
             outs.Write(data, 0, (int)data.Length);
             outs.Flush();
             outs.Dispose();
+            if (compressor != null)
+                compressor.Dispose();
```
ewoutkramer: We should introduce the using {} block here too

src/Hl7.Fhir.Core.Tests/Rest/FhirClientTests.cs
```diff
@@ -198,19 +198,56 @@ public void ReadRelativeAsync()
 			Assert.AreEqual("Den Burg", loc.Resource.Address.City);
 		}
 #endif
+        public static void Compression_OnBeforeRequestGZip(object sender, BeforeRequestEventArgs e)
```
ewoutkramer: What are these events doing?  Isn't the client already setting these headers?

brianpos: This is the manual way of setting the headers without using the property.
This way I can test each of the encryption techniques independently.
(otherwise always goes gzip)
Also wanted to ensure that it would decode the content returned, even if it wasn't asked for by the client code.

PR #262, opened 2016-10-24
Attribute changes for serialization/deserialization with DataContractSerializer

gtakov:
Added the needed attributes for serialization/deserialization with DataContractSerializer.


brianpos:
Can you please add some unit tests to show it in action working so we can understand what is happening, and that we can be sure as we make other changes don't break it.

ewoutkramer:
My main problem with this PR is that it alters generated code OR adds hand-written code that should be generated, especially all the KnownType attributes, they should ideally be generated at that spot, based on information in the profiles (data definitions) in Source/.

On the one hand, the DSTU2 model won't change much anymore, but if we port this to STU3 or even Revision 4, we really need the generated code to maintain this list.

Brian, how hard would it to do that?  Georgi is not the first one who'd like Serialization to work....

brianpos:
I can do the code generation no problems.
Just want to be sure from unit testing that we can ensure it stays working as models change, and that it will handle extensions correctly.

ewoutkramer:
I understand that @gtakov has written a unit test now that does a full roundtrip of all resources. Which takes a considerable amount of time, of course.

gtakov:
I have added the unit test.

ewoutkramer:
@gtakov @brianpos  Is this still worth doing?

ewoutkramer:
According to @brianpos, this change will not function well + we need changes to the T4 templates. Closing it.

ewoutkramer:
I want to take another look at this.

PR #271, opened 2016-11-30
Annotation support for non class types.

brianpos:
includes HasAnnotation and SetAnnotation helper methods


src/Hl7.Fhir.Core/Model/Base.cs
```diff
@@ -124,15 +125,28 @@ public void AddAnnotation(object annotation)
             annotations.Add(annotation);
         }
 
+        public bool HasAnnotation<A>()
+        {
+            return annotations.OfType(typeof(A)).Count() > 0;
```
ewoutkramer: Please change this to an Any(), instead op Count() > 0!

PR #282, opened 2016-12-23
Fix inconsistent URI encoding for Bundle links

tanvdo:
The Bundle class is currently using the `Uri.ToString()` method to store the various links. The `Uri.ToString()` method does preserve URL encoding. According to the documentation, `Uri.ToString()` unescapes all characters except #, ?, and %.

For example if one were to set the SelfLink property to `http://foo/bar?p1=a%2Fb&p2=c%3Ad` and subsquently retrieved the value, the resulting `SelfLink.AbsoluteUri` and `SelfLink.OriginalString` property values would both then be `http://foo/bar?p1=a%2Fb&p2=c:d`. Note that the `%3A` gets decoded. We do not believe this to be the desired behavior.

The fix for this issue to avoid the `Uri.ToString()` method in storing the links and leverage either the `Uri.AbsoluteUri` property for absolute `Uri`s or the `Uri.OriginalString` property for relative `Uri`s.

- https://msdn.microsoft.com/en-us/library/system.uri.tostring(v=vs.110).aspx
- http://stackoverflow.com/questions/7624987/whats-the-difference-between-uri-tostring-and-uri-absoluteuri
- https://code.logos.com/blog/2010/08/uritostring_must_die.html

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/Model/ModelTests.cs
+++ b/src/Hl7.Fhir.Core.Tests/Model/ModelTests.cs
@@ -82,6 +82,34 @@ namespace Hl7.Fhir.Tests.Model
        
 
         [TestMethod]
+        public void TestBundleLinkEncoding()
+        {
+            Action<string> test = (urlFormat) =>
+            {
+                var param1 = "baz/123";
+                var param2 = "qux:456";
+                var manuallyEncodedUrl = string.Format(urlFormat, "baz%2F123", "qux%3A456");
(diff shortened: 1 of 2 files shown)


ewoutkramer:
Yes, Uri.ToString() must die, I have tried to use OriginalString mostly, but there must be places where ToString() is used.

ewoutkramer:
I cherry-picked this commit into the develop branche. Thanks.

PR #310, opened 2017-03-22
Updated IConformanceResource for STU3

mmsmits:


brianpos:
I'm hoping to rename that backbone element class so that its ElementDefinition.BindingComponent, not the dud ElementDefinition.ElementDefinitionBindingComponent

mmsmits:
I agree, this way it's terrible, but the only way to make the project build currently..

PR #311, opened 2017-03-22
config: update AssemblyCopyright attributes, bump year from 2016 to 2017

wmrutten:


ewoutkramer:
I have updated the year in the "new" develop (based on vs2017/netstandard) by hand, this would now give merge conflicts.

PR #324, opened 2017-04-11
Null error checking when no context is passed to the navigate

brianpos:
(useful when testing an expression like today() with no context)


brianpos:
my use case is that I have a call that has no resource to work on, but will draw on environment variables only, (or today()) to extract data from. (Not as validate, but data extraction).
If this can be supported in other ways, then I'd be happy.

ewoutkramer:
(useful when testing an expression like today() with no context)

You should be able to do   

`{}.today()`

Could you try that?

ewoutkramer:
I think my alternative

`{}.today()` 

works,

or we could add a helper method to create an empty navigator.....

PR #326, opened 2017-04-19
FHIRPATH MoveToNext with nameFilter and other optimizations

brianpos:
Big change here, nameFilter added to MoveToNext and several performance updates (narrated)
All unit tests are still passing, however I haven't personally checked that the JsonDomFhirNavigator or XmlDomFhirNavigator implementations are ok.


ewoutkramer:
Hi Brian, the solution does not build after the PR on the CI server. Can you check what's going on?

See https://ci.appveyor.com/project/ewoutkramer/fhir-net-api-7cq0g/build/1.1.63

brianpos:
That last commit only altered a method from internal to public, so I expect this to be a transient failure, I'll double check though.

ewoutkramer:
Done most of the work - changing the signature for MoveToNext() to have a parameter. This in turn caused quite some rewrites to the navigators.....

PR #327, opened 2017-04-20
bugfix: SnapshotGenerator - also raise OnPrepareElement event for named slices

wmrutten:
The generated snapshot was correct, but due to the missing event notifications Forge would incorrectly determine the base element for named slices. This fixes it.


ewoutkramer:
Should this change be propagated to STU3 too?

PR #330, opened 2017-04-21
feature: implement ElementIdGenerator; fix Chris Grenz examples for unit testing

wmrutten:


ewoutkramer:
We have determined that this change deeply influences the way element id's are carried forward from base definition to profile - the impact is currently too big to integrate this change.

PR #341, opened 2017-05-07
udpate project settings for net45

brianpos:
debugpdb and documentation files (in debug only)


ewoutkramer:
Okay, I applied these changes by hand, using the .props file.

PR #357, opened 2017-05-25
Ft turtle stu3

mvdzel:
How about this?
I only need to add the RoundTripTests, this is working code, back ported from DSTU2.


mvdzel:
Now pull request on develop-stu3.

ewoutkramer:
I'll look at it this or next week!  Sorry, still working on my Madrid backlog ;-)

ewoutkramer:
We now - finally - have the right extensible parsing system in place to make this work really well with the .NET API.

It is documented here: http://docs.simplifier.net/fhirnetapi/parsing/intro-to-elementmodel.html and will be available starting from release 1.0 of the API.

Can you please rewrite this PR to work with these new interfaces?  I think a class like https://github.com/ewoutkramer/fhir-net-api/blob/develop/src/Hl7.Fhir.Serialization/FhirXmlNode.cs shows really well how this can be done.

Contact me if you need help porting this to the new methodology, @mvdzel !

PR #362, opened 2017-06-06
Fix for delete returning a status code of 200 being treated as unexpected

eug48:
STU3 is more clear that HTTP DELETE can return status code 200 - see http://hl7.org/fhir/http.html#delete

Deletes are currently failing for me on the HAPI 2.4 CLI test server with a FhirOperationException: 'Operation concluded successfully, but the return status 200 was unexpected'.


ewoutkramer:
Also cherry picked this into DSTU2 version

PR #368, opened 2017-06-16
bugfix: ElementIdGenerator - never inherit IDs from base profile

wmrutten:


wmrutten:
Never mind, needs more work.

PR #380, opened 2017-06-28
Added .NET 4.0 support

lstratman:
The changes to pack.bat were made so that I could easily pass `/p:FhirApiVersion` flags on the command line to support the fact that I have a custom version of this package until all of this makes it into the mainline.


ewoutkramer:
Minimal changes indeed!  I think I like it ;-)

brianpos:
Me too

ewoutkramer:
@ricardopinedathen - I cannot make the async stuff work for .NET4 - is there a way to fix that?

lstratman:
https://stackoverflow.com/questions/19421878/how-can-i-use-the-async-keywords-in-a-project-targeting-net-4-0/19421907

I think you may just need to add a reference to the `Microsoft.Bcl.Async` package for the `net40` target.

ewoutkramer:
I had already tried that, but that gave errors about https://www.nuget.org/packages/system.threading.tasks
not being available for 4.0....

lstratman:
If you move that `<PackageReference>` to a `netstandard` and `net45`-only section and add a `<Reference Include="System" />` to a `<ItemGroup>` for `net40`, that should pull in all of the package and assembly references that you need for each of the targets.

ricardopinedathen:
Where you able to solve this?

ewoutkramer:
No, I followed @lstratman's  instructions but I just keep getting compile errors. 

If I add the <PackageReference> without conditions, I'll get errors as stated above.  If I move them to a netstandard section - the same thing happens.  I don't understand why I would only move it there, cause I think we need that Bcl.Asyn package in the non-netstandard (and net45) settings.  If I only include it in net40 I get:

"warning MSB3268: The primary reference "C:\Users\ewout\.nuget\packages\microsoft.bcl.async\1.0.168\lib\net4
0\Microsoft.Threading.Tasks.dll" could not be resolved because it has an indirect dependency on the framework assembly
"System.Threading.Tasks, Version=1.5.11.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a" which could not be resolve
d in the currently targeted framework. ".NETFramework,Version=v4.0". To resolve this problem, either remove the referen
ce "C:\Users\ewout\.nuget\packages\microsoft.bcl.async\1.0.168\lib\net40\Microsoft.Threading.Tasks.dll" or retarget you
r application to a framework version which contains "System.Threading.Tasks, Version=1.5.11.0, Culture=neutral, PublicK
eyToken=b03f5f7f11d50a3a". [c:\git\fhir-net-api\src\Hl7.Fhir.Core\Hl7.Fhir.Core.csproj]"

I am lost at what to do, so no .net 4 for now still :-(

PR #396, opened 2017-07-14
[Develop-stu3] Fix Element Id attribute (string!); fix Snapshot generator base element annotations

wmrutten:


ewoutkramer:
Merged by cherry picking

PR #398, opened 2017-07-17
Update the EnumLiteralAttribute to support system

brianpos:
And some other changes too.
(not sure on the null error checking from April - that may want to be removed)
And you'll want to cherry pick this forward to STU3, and regenerate the t4 files (you can skip that part in the cherry-pick, as that commit was all code generated parts).


ewoutkramer:
Cherry-picked this into develop

brianpos:
Thanks

PR #403, opened 2017-07-27
Include the XML code comments and debug info

brianpos:
Include in the build the xml documentation file (which has the code comments in it) and also the PDB to ease debugging


ewoutkramer:
I have done this already some time ago - just not per project but in the shared fhir-net-api.props file...

ewoutkramer:
@brianpos - Brian you were going to check whether this was already in place (in the .prop files)? I'd like to close this PR!

PR #405, opened 2017-07-27
Mark types as serializable to enable Orleans serialization of a StructureDefinition

alexmarchis:
I added the serializable attribute on some Fhir.Model classes that are required by the Orleans framework in order to send a StructureDefinition serialized to clients

### DIFF
--- a/src/Hl7.Fhir.Core/Model/Base.cs
+++ b/src/Hl7.Fhir.Core/Model/Base.cs
@@ -40,6 +40,9 @@ using Hl7.Fhir.Utility;
 
 namespace Hl7.Fhir.Model
 {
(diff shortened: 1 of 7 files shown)


brianpos:
Am i right ok understanding that you're interesting a for server with Orleans?
https://dotnet.github.io/orleans/Documentation/Introduction.html
That's way cool.

brianpos:
Would you be willing to share your experience with it?

ewoutkramer:
@brianpos: Alex works for us, we're using Orleans in Simplifier to offload expensive operations!

brianpos:
Awesome, I'll hit you up in San Diego for some details

PR #406, opened 2017-07-28
Refactored the serializer and wrote more accurate unit tests

emanuelmd:


src/Hl7.Fhir.Core/Serialization/ComplexTypeWriter.cs
```diff
@@ -47,120 +49,104 @@ internal void Serialize(ClassMapping mapping, object instance, Rest.SummaryType
...
-
-            // Then emit the rest
-            foreach (var prop in mapping.PropertyMappings.Where(pm => pm.SerializationHint != XmlSerializationHint.Attribute))
+            else if (summary == Rest.SummaryType.Text)
             {
-                writeProperty(mapping, instance, summary, mode, prop);
+                propertiesToWrite = propertiesToWrite.Where(property =>
+                       summaryTextProperties.Contains(property.Name.ToLower())
+                    || property.IsMandatoryElement
+                    || isMetaTextOrIdElementInstance(instance));
```
ewoutkramer: This is an excellent case where isMetaTextOrIdElementInstance() could be a C#7 local function, it's a one-liner that is not used elsewhere

src/Hl7.Fhir.Core/Serialization/ComplexTypeWriter.cs
```diff
@@ -47,120 +49,104 @@ internal void Serialize(ClassMapping mapping, object instance, Rest.SummaryType
...
+                propertiesToWrite = propertiesToWrite.Where(property => property.InSummary);
             }
-
-            // Then emit the rest
-            foreach (var prop in mapping.PropertyMappings.Where(pm => pm.SerializationHint != XmlSerializationHint.Attribute))
+            else if (summary == Rest.SummaryType.Text)
             {
-                writeProperty(mapping, instance, summary, mode, prop);
+                propertiesToWrite = propertiesToWrite.Where(property =>
+                       summaryTextProperties.Contains(property.Name.ToLower())
```
ewoutkramer: The original code also checked the type of the property, to try to avoid serializing properties that have accidentally the same name as the intended properties.  It's still not bullet-proof, since we would need the parent path to be sure, but it helps.

src/Hl7.Fhir.Core/Serialization/ComplexTypeWriter.cs
```diff
@@ -47,120 +49,104 @@ internal void Serialize(ClassMapping mapping, object instance, Rest.SummaryType
 
             _writer.WriteStartComplexContent();
 
-            // Emit members that need xml /attributes/ first (to facilitate stream writer API)
-            foreach (var prop in mapping.PropertyMappings.Where(pm => pm.SerializationHint == XmlSerializationHint.Attribute))
+            // Emit members that need xml attributes / first (to facilitate stream writer API)
+            // attributes first (xml) and order maintained for the rest
+            var propertiesToWrite = mapping.PropertyMappings
```
ewoutkramer: The original code *always* wrote the "id" property but I think the net result is the same: id is marked as "in summary" (for 'true'), explicitly included (for 'text' and 'count'), not excluded (for 'data'), or included (for 'false').

src/Hl7.Fhir.Core/Serialization/ComplexTypeWriter.cs
```diff
@@ -47,120 +49,104 @@ internal void Serialize(ClassMapping mapping, object instance, Rest.SummaryType
...
-            if (mode == SerializationMode.NonValueElements && prop.RepresentsValueElement) return;
+            if (mode == SerializationMode.ValueElement && !property.RepresentsValueElement) return;
+            if (mode == SerializationMode.NonValueElements && property.RepresentsValueElement) return;
 
-            object value = prop.GetValue(instance);
-            var isEmptyArray = (value as IList) != null && ((IList)value).Count == 0;
+            object value = property.GetValue(instance);
 
-         //   Message.Info("Handling member {0}.{1}", mapping.Name, prop.Name);
+            if (value is IList && ((IList)value).Count == 0) return;
```
ewoutkramer: We could turn this into c#7  if(value is IList ilist && ilist.Count == 0 now.

src/Hl7.Fhir.Core/Serialization/ComplexTypeWriter.cs
```diff
@@ -47,120 +49,104 @@ internal void Serialize(ClassMapping mapping, object instance, Rest.SummaryType
...
 
-            if ((value != null || prop.RepresentsValueElement && prop.ElementType.IsEnum() && !string.IsNullOrEmpty(((Primitive)instance).ObjectValue as string)) && !isEmptyArray)
+            bool isEnum = property.ElementType.IsEnum(),
+                 isValueElement = property.RepresentsValueElement,
+                 isEmptyPrimitive = instance is Primitive && string.IsNullOrEmpty(((Primitive)instance).ObjectValue as string);
```
ewoutkramer: Another c#7 "is" candidate

wmrutten:
FYI the exact format is defined by the FHIR specification:
https://www.hl7.org/fhir/datatypes.html#datetime

emanuelmd:
Well, the spec describes date search values here https://www.hl7.org/fhir/search.html#date and it gives this example "lt2013-01-14T10:00" as a valid date value, the fhir-net-api doesn't recognize this as valid.

wmrutten:
Emmanuel, the regex is defined by FHIR in the offical StructureDefinition that defines the datetime datatype, via an extension on the dateTime.value element:
https://www.hl7.org/fhir/examples/datetime.profile.xml
The API picks up the regex via code generation from the definitions in the offical FHIR specification.zip, so you shouldn't try to manually override it.

If you find the regex to be incorrect, then the proper way is to submit a change request to gforge (FHIR issue tracker). Once the specification has been updated, we should then re-run code generation on the API to pick up the change.

I suggest you add a comment in the source code with a link to the change request on gForge.

emanuelmd:
@wmrutten All right, I wasn't really sure where to address this issue. Definitely not here. Thanks for your suggestion. I will submit the time-zone issue to gForge and keep this PR time-validating-regex-modifications free :)

wmrutten:
Great! You could post a link to the new gForge issue here as a comment.

PR #414, opened 2017-08-23
Develop stu3 - improve dependency configuration; bugfix for snapshot generator

wmrutten:
- Update package dependencies to use the lowest required versions, i.e. System.IO.Compression, so the API does not force client's to also use the latest versions of .NET Core components. Especially useful for clients on .NET Framework, i.e. Forge.
- Fix a bug in the snapshot generator, try to match type profile constraints with any type in the base element. Previously the snapshot generator only tried to match the first base type and would emit a warning on mismatch. However we should verify if a type profile matches any base type, and only emit a warning if the profile is incompatible with *all* base types.


ewoutkramer:
Integated by merge into STU3, by cherry picking into DSTU2

PR #441, opened 2017-10-17
Add support for parsing POCOs from IElementNavigator

ewoutkramer:
* Adds overloads to XmlFhirParser and JsonFhirParser to accept IElementNavigator
* Internally, converts all inputs (stream, reader, string) to IElementNavigator, then uses the ElementNavFhirReader to wrap it and implement the original IFhirReader
* Then calls existing parser code based on IFhirReader


brianpos:
What's the performance of this like?

ewoutkramer:
It's as fast, or faster. The IElementNavigator implementations reading xml or json are written quite efficiently - and more efficient than the current IFhirReader implementations.  E.g. parsing an "average" size patient example in xml, and extracting 5 key elements from it using the Xml implementation of IElementNavigator takes 7 microseconds on my machine.

You will not see a difference in the Poco parsing though - setting/getting the poco values using reflection is much slower, and will dwarf any performance gain or loss in the IElementNavigators....

Issue #446, opened 2017-10-23
Custom Serializer processor


mharthoorn:
Make the serializer extensible, so that serialization and deserialization can be post processed.

Specific implementation need: adding a light weight method to serialize annotations (IAnnotatable).

brianpos:
want to be very careful that this doesn't get into the base serializers, and unit tests to ensure that it doesn't.

ewoutkramer:
Fixed in e65355138488aeeaa128acbbdd73c8ec24aae916

PR #449, opened 2017-10-25
Feature/446 customize serialization

ewoutkramer:
This feature addition enables the user to customize both the serialization and deserialization process using the ParserSettings class. The ParserSettings now contains two properties with interfaces, which - when implemented - will be invoked at strategic times during the serialization and deserialization process to insert custom behaviour.

### DIFF
--- a/src/Hl7.Fhir.Core/Serialization/ElementNavFhirReader.cs
+++ b/src/Hl7.Fhir.Core/Serialization/ElementNavFhirReader.cs
@@ -21,10 +21,11 @@ namespace Hl7.Fhir.Serialization
     /// with the POCO-parsers.
     /// </summary>
 #pragma warning disable 612, 618
-    internal class ElementNavFhirReader : IFhirReader
+    internal class ElementNavFhirReader : IFhirReader, IElementNavigator
 #pragma warning restore 612,618
     {
         private IElementNavigator _current;
+
         public bool DisallowXsiAttributesOnRoot { get; set; }
 
         public ElementNavFhirReader(IElementNavigator root, bool disallowXsiAttributesOnRoot = false)
@@ -33,14 +34,11 @@ namespace Hl7.Fhir.Serialization
             _current = root;
         }
 
-        public object GetPrimitiveValue()
-        {
-            return _current.Value;
-        }
+        public object GetPrimitiveValue() => Value;
 
         public string GetResourceTypeName()
         {
-            if (_current.Type != null) return _current.Type;
+            if (Type != null) return Type;
 
             // No type name on this element....give the users some details about why?
             var xmlDetails = getXmlDetails(_current);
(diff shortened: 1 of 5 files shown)


ewoutkramer:
I pulled, fixes #446

src/Hl7.Fhir.Core/Serialization/ElementNavFhirReader.cs
```diff
@@ -123,8 +120,31 @@ private bool verifyXmlSpecificDetails(IElementNavigator child)
...
+        }
+
+        public bool MoveToFirstChild(string nameFilter = null)
+        {
+            return _current.MoveToFirstChild(nameFilter);
+        }
+
+        public IElementNavigator Clone()
+        {
+            return _current.Clone();
```
mharthoorn: The Clone should return a clone ElementNavFhirReader instead of the inner reader.

src/Hl7.Fhir.Core/Serialization/ElementNavFhirReader.cs
```diff
@@ -123,8 +120,31 @@ private bool verifyXmlSpecificDetails(IElementNavigator child)
                 throw Error.Format($"Xml node of type '{xmlDetails.NodeType}' is unexpected at this point", this);
         }
 
+        public bool MoveToNext(string nameFilter = null)
```
emanuelmd: Since the methods below are one-liners and you also have methods like LineNumber using the new C# syntax, you should also write those three below with the new syntax. Consistency is key :)

Eg:
`public bool MoveToNext(string nameFilter = null) => _current.MoveToNext(nameFilter);`

src/Hl7.Fhir.Core/Serialization/ElementNavFhirReader.cs
ewoutkramer: Yes, it should indeed.

src/Hl7.Fhir.Core/Serialization/ElementNavFhirReader.cs
ewoutkramer: will do

src/Hl7.Fhir.Core/Serialization/ElementNavFhirReader.cs
ewoutkramer: Turned ElementNavReader into a struct, create a clone of itself by creating a new instance + clone the inner IElementNavigator (_current)

src/Hl7.Fhir.Core/Serialization/ElementNavFhirReader.cs
ewoutkramer: Funny, all these methods were generated by a Visual Studio refactoring (implement interface by forwarding to a private member variable).  It will generate lambdas for properties, but obviously not for methods!

PR #450, opened 2017-10-26
feature: using pre-compiled expressions instead of reflection for POCO construction

ewoutkramer:
When serializing a POCO, we use reflection to get the values out of the POCO into the serialization format. Conversely, when parsing, we use reflection to initialize the POCO with the values from the serialized format.

Reflection is slow, so I replaced that code by two things
* During initial construction of the parser/serializer, I create an expression tree for all setters/getters and compile these to a lambda
* During parsing/serialization I use the generated lambda to get "compile speed" performance

This means setup is a bit more expensive, but parsing is 25% faster.

Worth doing?

I know one possible disadvantage is the fact that on iOS, codegeneration is not available. So, I disabled this feature for netstandard1.1.  Not that netstandard1.1 cannot do it, but at least iOS users have a package that does not use codegen to chose from....


wmrutten:
Sounds good!

src/Hl7.Fhir.Core/Introspection/PropertyInfoExtensions.cs
```diff
@@ -0,0 +1,70 @@
...
+        public static Action<object, object> GetValueSetter(this PropertyInfo propertyInfo)
+        {
+            var instance = Expression.Parameter(typeof(object), "i");
+            var convertIn = Expression.Convert(instance, propertyInfo.DeclaringType);
+            var argument = Expression.Parameter(typeof(object), "a");
+            var setterCall = Expression.Call(
+                convertIn,
+                propertyInfo.SetMethod,
+                Expression.Convert(argument, propertyInfo.PropertyType));
+            return (Action<object, object>)Expression.Lambda(setterCall, instance, argument).Compile();
```
brendankowitz: This looks mostly the same as the method above, why not try to combine these and just add the convert operation when needed?

ewoutkramer: I probably just use one of the two - so we can remove one of the "almost" copies.

ewoutkramer: Ok, I isolated the repeated code in a separate functions (buildSetter, buildGetter)

ewoutkramer:
Fixes #604

wmrutten:
Great!

Currently the PropertyInfoExtensions class is declared as `internal`:
https://github.com/ewoutkramer/fhir-net-api/blob/develop/src/Hl7.Fhir.Core/Introspection/PropertyInfoExtensions.cs

I would like to apply similar optimization in Forge. Is it possible to make this class public?

PR #455, opened 2017-10-26
Develop - Merge ArtifactSummary & DirectorySource from feature/indexed-source

wmrutten:
Not merged: BlobDb, FhirArchive


src/Hl7.Fhir.Specification/Specification/Source/Summary/ArtifactSummaryDetails.cs
```diff
@@ -0,0 +1,200 @@
...
+using System;
+using System.Linq;
+
+namespace Hl7.Fhir.Specification.Source.Summary
+{
+    // DSTU2 Specific!
+
+    /// <summary>For accessing common artifact summary details.</summary>
+    /// <remarks>Common details are initialized by the <see cref="ArtifactSummaryGenerator"/> itself.</remarks>
+    public static class ArtifactSummaryDetails
```
ewoutkramer: Doesn't this mean that all extension methods on IArtifactSummaryDetailsProvider are in scope all the time?  So, now there's no real notion of "these attributes are right for a ValueSet"?

ewoutkramer: Since this are all extension methods on static classes, why do we still have an ArtifactSummaryFactory? There are no specific classes of instances to create?

src/Hl7.Fhir.Specification/Specification/Source/DirectorySourceSettings.cs
```diff
@@ -0,0 +1,273 @@
...
+        /// interface to extract summary information from FHIR artifacts, independent of the underlying
+        /// resource serialization format. By default, the <see cref="DirectorySource"/> calls the
+        /// <see cref="DefaultNavigatorStreamFactory.Create(string)"/> method to create
+        /// a concrete navigator stream for the current file. The default factory supports
+        /// navigators for "*.xml" and "*.json" files.
+        /// The caller can override the default navigator streams by specifying a custom
+        /// <see cref="NavigatorStreamFactory"/> delegate. This allows clients to implement
+        /// support for alternative serialization formats.
+        /// </remarks>
+        public NavigatorStreamFactory StreamFactory { get; set; } // = DefaultNavigatorStreamFactory.Create;
```
ewoutkramer: To align with the rest of the API, I think I prefer an event-based approach here, instead of a factory.

Settings.OnNavigatorStreamNeeded, with the given path as a parameter to the event

src/Hl7.Fhir.Specification/Specification/Source/Summary/ArtifactSummaryDetailsCollection.cs
```diff
@@ -0,0 +1,54 @@
...
+using System.Collections.Generic;
+
+namespace Hl7.Fhir.Specification.Source.Summary
+{
+    /// <summary>Common interface for retrieving artifact summary details by key.</summary>
+    /// <remarks>
+    /// Implemented by <see cref="ArtifactSummaryDetailsCollection"/> as well as <see cref="ArtifactSummary"/>.
+    /// Allows common extension methods for retrieving specific summary details.
+    /// </remarks>
+    public interface IArtifactSummaryDetailsProvider
```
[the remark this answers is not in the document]
ewoutkramer: Can we not combine ArtifactSummary and ArtifactSummaryCollection?  Why would we keep both?

src/Hl7.Fhir.Specification/Specification/Source/Summary/ArtifactSummaryDetails.cs
```diff
@@ -0,0 +1,200 @@
...
+        /// <remarks>The <see cref="ArtifactSummaryGenerator"/> calls this method from a <see cref="ArtifactSummaryDetailsExtractor"/> delegate.</remarks>
+        internal static bool Extract(IElementNavigator nav, ArtifactSummaryDetailsCollection details)
+        {
+            // if (details.IsStructureDefinition())
+            if (details.GetResourceTypeName() == StructureDefinitionTypeName)
+            {
+                // Explicit extractor chaining
+                if (ConformanceSummaryDetails.Extract(nav, details))
+                {
+                    nav.TryExtractValue(details, StructureDefinitionKindKey, "kind");
```
ewoutkramer: The boolean results from TryExtractValue() are not processed into the return value of the extraction?

src/Hl7.Fhir.Specification/Specification/Source/Summary/ArtifactSummaryGenerator.cs
```diff
@@ -0,0 +1,203 @@
...
+
+                while (navStream.MoveNext())
+                {
+                    var current = navStream.Current;
+                    if (current != null)
+                    {
+                        var props = new ArtifactSummaryDetailsCollection();
+
+                        // Add default summary details
+                        props[ArtifactSummaryDetails.OriginKey] = origin;
```
ewoutkramer: Waarom zijn deze regels niet in een extractor gestopt?

src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
```diff
@@ -4580,32 +4580,27 @@ public void TestTypeProfileWithChildElementBinding()
             Base = ModelInfo.CanonicalUriForFhirCoreType(FHIRDefinedType.QuestionnaireResponse),
```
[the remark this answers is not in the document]
wmrutten: Nope, this shouldn't be part of this commit, not related to DirectorySource at all.
Actually, this unit test is commented out on DSTU2 branch.

src/Hl7.Fhir.Specification/Specification/Source/DirectorySourceSettings.cs
wmrutten: Sure! And also rewrite SummaryFactory delegate to event?

src/Hl7.Fhir.Specification/Specification/Source/IConformanceSource.cs
```diff
@@ -24,6 +22,9 @@ public interface IConformanceSource : IResourceResolver
         /// <returns></returns>
         IEnumerable<string> ListResourceUris(ResourceType? filter = null);
 
+        // [WMR 20171011] New
```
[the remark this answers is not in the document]
wmrutten: Yes, I think IConformanceSource should also include the following new method:
IEnumerable<ArtifactSummary> List()

src/Hl7.Fhir.Specification/Specification/Source/Summary/ArtifactSummaryDetails.cs
wmrutten: 1) Correct, I removed all concrete subclasses to simplify the default API design. As a consequence, we lose some type safety.
2) Client can override factor to create specific subclasses with additional type-safe properties. However by default, the API generates standard ArtifactSummary instances.

src/Hl7.Fhir.Specification/Specification/Source/Summary/ArtifactSummaryDetails.cs
wmrutten: Correct, usually the caller can (and will) safely ignore the value. In some circumstances, it may be useful for callers to detect if a match was found.

src/Hl7.Fhir.Specification/Specification/Source/DirectorySourceSettings.cs
ewoutkramer: I now understand you want to do this, not so much to support customization of creation of the navigator stream, but to have an extensible way to support more (future) formats of FHIR.  This is a good idea, but we should generalize that - i.e. how would I know which parser to create for a given format?  (not just the stream).  Let's think about that for a bit before deciding on this.

ewoutkramer: I'd like to know what @christiaan and @martijn think about this too.

src/Hl7.Fhir.Specification/Specification/Source/Summary/ArtifactSummaryDetailsCollection.cs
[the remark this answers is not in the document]
wmrutten: I decided to encapsulate the backing store. Client code just needs an indexer property this[key], so this is the interface expresses. Implementation is free to choose a specific back store, we could change/optimize this at a later time.
However IReadOnlyDictionary seems to fit this use case, so we could decide to use this and drop the custom interface. At the cost of a bit more work when we'd want to change the backing store (to implement the additional methods).

I want to prevent consumers from changing harvested properties. However I also don't want to repeat all this property logic. So I decided on a kind of builder pattern. The extractor fills a read/write collection, then hands it over to initialize a read-only summary instance, which saves and wraps the r/w collection. By implementing a common interface, we can re-use extension methods on both the r/w collection class and the final r/o summary instance.

src/Hl7.Fhir.Specification/Specification/Source/Summary/ArtifactSummaryGenerator.cs
```diff
@@ -0,0 +1,203 @@
...
+using System.Collections.Generic;
+
+namespace Hl7.Fhir.Specification.Source.Summary
+{
+    /// <summary>Represents a factory method that creates a new <see cref="ArtifactSummary"/> instance.</summary>
+    /// <param name="typeName">The type name of the artifact.</param>
+    /// <param name="details">A collection of summary details extracted from the artifact.</param>
+    /// <param name="error">A error that occured while extracting details from the artifact, or <c>null</c>.</param>
+    /// <returns>A new <see cref="ArtifactSummary"/> instance.</returns>
+    public delegate ArtifactSummary ArtifactSummaryFactory(string typeName, ArtifactSummaryDetailsCollection details, Exception error = null);
```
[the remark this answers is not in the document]
wmrutten: As explained above, the extractor(s) first initialize a r/w collection. Then the extractor calls factory method to create a r/o Summary instance, initialized from the r/w collection.

src/Hl7.Fhir.Specification/Specification/Source/Summary/ArtifactSummaryDetails.cs
ewoutkramer: Since we've chosen to have no subclasses, I think we should ditch the factory pattern for ArtifactSummaries then.  Your namespace will contain those extension methods, so for the API user this will only create confusion.

src/Hl7.Fhir.Specification/Specification/Source/Summary/ArtifactSummaryNavigationExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+    {
+        /// <summary>
+        /// Try to position the navigator on the current or sibling element with the specified name.
+        /// If the current element name matches, then maintain the current position.
+        /// Otherwise, navigate to the next matching sibling element (if it exists).
+        /// </summary>
+        /// <param name="nav">An <see cref="IElementNavigator"/> instance.</param>
+        /// <param name="element">An element name.</param>
+        /// <returns><c>true</c> if, upon return, the navigator is positioned on a matching element, or <c>false</c> otherwise.</returns>
+        public static bool Find(this IElementNavigator nav, string element)
```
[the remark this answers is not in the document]
wmrutten: Sure, but we have to make sure the semantics fit the generic case.
These extension methods were specifically designed to facilitate the extraction process.

src/Hl7.Fhir.Specification/Specification/Source/Summary/ArtifactSummaryNavigationExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+            return nav.Find(element) && nav.TryExtractValue(details, key);
+        }
+
+        /// <summary>Extract the value of a child element into an <see cref="ArtifactSummaryDetailsCollection"/> using the specified key.</summary>
+        /// <param name="nav">An <see cref="IElementNavigator"/> instance.</param>
+        /// <param name="details">An <see cref="ArtifactSummaryDetailsCollection"/> instance.</param>
+        /// <param name="key">A collection key.</param>
+        /// <param name="element">An element name.</param>
+        /// <param name="childElement">A child element name.</param>
+        public static bool TryExtractValue(this IElementNavigator nav, ArtifactSummaryDetailsCollection details, string key, string element, string childElement)
```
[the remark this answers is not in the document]
wmrutten: I don't think so....
All methods are expected to advance to the matching element, if found.
The child element value is extracted using TryExtractValue, which DOES clone the navigator, so the original navigator remains positioned on the matching element.
Behavior is tested and confirmed by unit tests (but we might need some more...).

src/Hl7.Fhir.Specification/Specification/Source/Summary/ArtifactSummaryNavigationExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        /// <summary>Add the value of the current element to a list, if not missing or empty.</summary>
+        /// <param name="nav">An <see cref="IElementNavigator"/> instance.</param>
+        /// <param name="values">A list of values.</param>
+        public static bool TryExtractValue(this IElementNavigator nav, List<string> values)
+        {
+            var value = nav.Value; // ?.ToString();
+            if (value != null)
+            {
+                // Assumption: navigator always returns string values (?)
```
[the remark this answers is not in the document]
wmrutten: Will do!

src/Hl7.Fhir.Specification/Specification/Source/Summary/ArtifactSummaryNavigationExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+            return false;
+        }
+
+        /// <summary>Extract an array of child element values into an <see cref="ArtifactSummaryDetailsCollection"/> using the specified key.</summary>
+        /// <param name="nav">An <see cref="IElementNavigator"/> instance.</param>
+        /// <param name="details">An <see cref="ArtifactSummaryDetailsCollection"/> instance.</param>
+        /// <param name="key">A collection key.</param>
+        /// <param name="element">An element name.</param>
+        /// <param name="childElement">A child element name.</param>
+        public static bool TryExtractValues(this IElementNavigator nav, ArtifactSummaryDetailsCollection details, string key, string element, string childElement)
```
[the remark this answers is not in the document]
wmrutten: Hmmm... thx ;p

src/Hl7.Fhir.Specification/Specification/Source/Summary/ArtifactSummaryGenerator.cs
ewoutkramer: Because all that information does not come from an input navigator ;-)

src/Hl7.Fhir.Specification/Specification/Source/Summary/ArtifactSummaryNavigationExtensions.cs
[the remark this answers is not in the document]
ewoutkramer: So, maybe the documentation should be clearer that these are specific "consumer" or "munch" methods, that take a very specific set of parameters

wmrutten:
Pushed an updated version for review.
Tried to simplify and improve the design, in response to code review.
Renamed some types for clarification.
Added unit test to demonstrate/test custom harvester implementation.

wmrutten:
Updated ArtifactSummary design in response to code review.
Merged updated logic from feature/indexed-source into develop (DSTU2).
Does not include new experimental NonReentrantResolver (moved to separate feature branch).

Issue #478, opened 2017-11-16
[FhirClient] Provide a way to manipulate the body in OnBeforeRequest/OnAfterResponse


ewoutkramer:
To be able to implement content encryption, we need to be able to encrypt (on the way out) en decrypt (on the way in) the data flowing through the FhirClient.

Currently, we do have events to support this (and manipulate the *headers*), but these won't allow you to manipulate the body (technically, you can manipute the byte array passed into the event, but that would not allow you to resize it)

Maybe we should introduce additional callbacks that would allow you to manipulate the body in addition to the current callbacks (which would not require us to break the current interface)

ewoutkramer:
In fact, this originated from a discussion of how to implement

https://tools.ietf.org/html/rfc8188

in FHIR

brianpos:
This would need to be on both sides, to encrypt on write, and also to decrypt on read

kennethmyhra:
@ewoutkramer @brianpos I have referenced a commit to this issue which should solve it, it's based on the code I retrieved from Brian at FHIR Dev Days.

I would though like to see this implemented it in a fashion that would give me more control over the request/response pipeline much in the same way as is done with MessageHandlers/DelegateHandlers in ASP.NET, it will then allow me to add the required handlers when setting up the FhirClient. For example when encrypting it would make sense to encrypt after you have compressed the content rather than before.

Maybe the request/response pipeline is something that could be considered when implementing issue #480 as it seems to me it would break the current implementation.

ewoutkramer:
Yes, connecting this issue with #480 makes sense, since we only need to break the interface once, and also it is likely whatever pipelining we make up will need to be redone/retested when we switch to the new http client...

kennethmyhra:
Ewout, saw your comment in #480 in regards to creating a new FhirWebClient class. Do you want me to issue a PR for c8dfc8e (referenced above) into the current FhirClient? If any changes are required before pulling let me know and I will fix them.

ewoutkramer:
Yes, that seems to make sense to me.

marcovisserFurore:
Solved by the introduction of the new FhirClient (version SDK 2.0)

PR #496, opened 2017-12-04
Support for manipulating request and response body

kennethmyhra:
* Added callbacks that allows manipulation of the request and response
  stream
* See issue #478 for more information

### DIFF
--- a/src/Hl7.Fhir.Core/Rest/FhirClient.cs
+++ b/src/Hl7.Fhir.Core/Rest/FhirClient.cs
@@ -13,6 +13,7 @@ using Hl7.Fhir.Support;
 using Hl7.Fhir.Utility;
 using System;
 using System.Collections.Generic;
+using System.IO;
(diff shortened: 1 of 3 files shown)


ewoutkramer:
Anyway, maybe we can piggyback on the new pipeline capabilities in the HttpClient. I am still getting updated on that new class, so I don't know yet!

kennethmyhra:
Agree, didn't look into it in detail earlier, but I see now that it's the same interface that's used on the server side and it should therefore work quite similar.

ewoutkramer:
@kennethmyhra -- oops, I never submitted my review on this PR....Is this PR still of interest to you? (the new client written in #535 will actually fix this is a more elegant way, because of the new .NET Core webclient used).

kennethmyhra:
@ewoutkramer I am happy with the solution that the .NET Core WebClient gives us in PR #507.

PR #498, opened 2017-12-04
fix: Validator no longer produces a warning about unknown references …

ewoutkramer:
…if ResolveReferences is disabled in settings

### DIFF
--- a/src/Hl7.Fhir.Specification.Tests/Validation/BasicValidationTests.cs
+++ b/src/Hl7.Fhir.Specification.Tests/Validation/BasicValidationTests.cs
@@ -318,12 +318,12 @@ namespace Hl7.Fhir.Specification.Tests
 
             Patient p = new Patient();
             p.Active = true;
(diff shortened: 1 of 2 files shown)


src/Hl7.Fhir.Specification.Tests/Validation/BasicValidationTests.cs
```diff
@@ -311,6 +311,27 @@ public void ValidatesMultiplePossibleTypeRefs()
...
 
 
+        [Fact]
+        public void DoNotFollowRefsSuppressesWarning()
+        {
+            var validator = new Validator(new ValidationSettings { ResourceResolver = _source, ResolveExteralReferences = true });
+
+            Patient p = new Patient();
+            p.Active = true;
+            p.ManagingOrganization = new ResourceReference("http://isure.cannot.be.found.nl/fhir/Patient/1");
```
marcovisserFurore: Text error: isure must be issue. Or better: reference could not be found?

ewoutkramer: yep, indeed.

PR #499, opened 2017-12-04
Feature/494 Directory Source - single threaded by default

wmrutten:
Default behavior is single threaded, safe default for all environments (especially servers).
Desktop clients can explicitly enable multi-threading if desired.


wmrutten:
IMPORTANT: first merge pull request #500 (pre-DevDays changes to ArtifactSummary)

PR #503, opened 2017-12-05
feature 371: will now check Discriminator.type before creating SlicingBucket

ewoutkramer:


src/Hl7.Fhir.Specification/Validation/Validator.cs
```diff
@@ -490,12 +493,16 @@ private OperationOutcome snapshotGenerationNeeded(StructureDefinition definition
             if (generator != null)
             {
                 generator.Update(definition);
+
 #if DUMP_SNAPSHOTS
 
-                string xml = FhirSerializer.SerializeResourceToXml(definition);
-                string name = definition.Id ?? definition.Name.Replace(" ", "");
+                string xml = (new FhirXmlSerializer()).SerializeToString(definition);
```
marcovisserFurore: Should SerializeToString not be static?

src/Hl7.Fhir.Specification/Validation/Validator.cs
```diff
@@ -490,12 +493,16 @@ private OperationOutcome snapshotGenerationNeeded(StructureDefinition definition
...
+                string name = definition.Id ?? definition.Name.Replace(" ", "").Replace("/", "");
+                var dir = @"c:\temp\validation";
```
marcovisserFurore: Hmmm, would this not be better?
`var dir = Path.Combine(Path.GetTempPath(), "validation");`

ewoutkramer: Yes, great idea, will do.  It's only run for debugging, but still, this is placed within production code, so it should be portable.

PR #507, opened 2017-12-05
Feature/480 FhirClient implementation with HttpClient

Devereux-Henley:
- Created new implementation of FhirClient using [System.Net.Http.HttpClient](https://msdn.microsoft.com/en-us/library/system.net.http.httpclient(v=vs.118).aspx). 
- New mirrored classes placed in `Hl7.Fhir.Rest.Http` namespace
- Common functionality extracted to a base class.
- Old implementation marked as `[Obsolete]`
- `FhirClient` now `IDisposable`
- Simply swapping classes will compile, throw exceptions on obsoleted functionality.


ewoutkramer:
Thanks! That's a major chunk of work. I will need some time to digest it ;-)

src/Hl7.Fhir.Core/Rest/Http/HttpToEntryExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+            result.Response.Location = response.Headers.Location?.AbsoluteUri ?? response.Content.Headers.ContentLocation?.AbsoluteUri;
+
+            result.Response.LastModified = response.Content.Headers.LastModified;
+            result.Response.Etag = response.Headers.ETag?.Tag;
+
+            if (body != null && body.Length != 0)
+            {
+                result.Response.SetBody(body);
+
+                if (Rest.HttpToEntryExtensions.IsBinaryResponse(result.Response.Location, contentType.ToString()))
```
ewoutkramer: I think this must be contentType.MediaType, since otherwise it might include the charset parameter, which was removed by the original code using `getContentType()`.

src/Hl7.Fhir.Core/Rest/Http/HttpToEntryExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+                {
+                    result.Resource = Rest.HttpToEntryExtensions.MakeBinaryResource(body, contentType.ToString());
```
ewoutkramer: This is actually better than the original, since it will include the charset parameter, which the original code did not.

src/Hl7.Fhir.Core/Rest/Http/HttpToEntryExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+                        result.Resource.Id = ri.Id;
+                        result.Resource.Meta = new Meta();
+                        result.Resource.Meta.VersionId = ri.VersionId;
+                        result.Resource.ResourceBase = ri.BaseUri;
+                    }
+                }
+                else
+                {
+                    var bodyText = Rest.HttpToEntryExtensions.DecodeBody(body, charEncoding);
+                    var resource = Rest.HttpToEntryExtensions.ParseResource(bodyText, contentType.ToString(), parserSettings, throwOnFormatException);
```
ewoutkramer: Also here, contentType.ToString() would include the charset parameter (if the client passed that in), which will not work well with `DecodeBody`, I think this should be `contentType.MediaType`

src/Hl7.Fhir.Core/Rest/Http/EntryToHttpExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            var location = new RestUrl(interaction.Url);
+
+            if (useFormatParameter)
+                location.AddParam(HttpUtil.RESTPARAM_FORMAT, Hl7.Fhir.Rest.ContentType.BuildFormatParam(format));
+
+            var request = new HttpRequestMessage(getMethod(interaction.Method), location.Uri);
+
+            if (!useFormatParameter)
+                request.Headers.Add("Accept", Hl7.Fhir.Rest.ContentType.BuildContentType(format, forBundle: false));
```
ewoutkramer: What is the reason you are not doing request.Headers.Accept.Add()?

src/Hl7.Fhir.Core/Rest/Http/EntryToHttpExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            if (interaction.IfMatch != null) request.Headers.TryAddWithoutValidation("If-Match", interaction.IfMatch);
```
ewoutkramer: Is the way we use If-Match in FHIR illegal, so we need to avoid validation?

src/Hl7.Fhir.Core/Rest/Http/EntryToHttpExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+
+                // This is done by the caller after the OnBeforeRequest is called so that other properties
+                // can be set before the content is committed
+                // request.WriteBody(CompressRequestBody, body);
+                contentType = Hl7.Fhir.Rest.ContentType.BuildContentType(format, forBundle: false);
+            }
+
+            request.Content = new ByteArrayContent(body);
+
+            // MediaTypeHeaderValue cannot accept a content type that contains charset at the end, so that value must be split out.
```
ewoutkramer: Yes, I noticed the same, weird. Maybe I'll take a look at the reference source to see why.

ewoutkramer: If you look at https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/Headers/MediaTypeHeaderValue.cs#L203, it seems there is support for parsing mediatypes with parameters...

src/Hl7.Fhir.Core.Tests/Rest/Mocks/MockHttpMessageHandler.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Hl7.Fhir.Core.Tests.Rest.Mocks
+{
+    public class MockHttpMessageHandler : HttpClientHandler
```
ewoutkramer: This MessageHandler could be used by people who want to still use the OnBefore/OnAfter logic, right?  The name "mock" did not suggest that to me (I would think it has something to do with testing).  Maybe call it "RequestResponseEventMessageHandler" or something like that?

ewoutkramer: Oh, this is actually there to support testing ;-)   But we could provide it more generally, as I mentioned, right?

Devereux-Henley: Yes, this could be provided more generally. I will move this to the new namespace and rename.

src/Hl7.Fhir.Core/Rest/Http/EntryToHttpExtensions.cs
Devereux-Henley: Will replace with `MediaTypeHeaderValue.Parse(contentType)`

src/Hl7.Fhir.Core/Rest/Http/HttpToEntryExtensions.cs
Devereux-Henley: Will change to `MediaType`.

src/Hl7.Fhir.Core/Rest/Http/HttpToEntryExtensions.cs
Devereux-Henley: Will replace with `MediaType`

src/Hl7.Fhir.Core/Rest/Http/HttpToEntryExtensions.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+            catch (ArgumentException e)
+            {
+                charEncoding = Encoding.UTF8;
+            }
+
+            result.Response.Location = response.Headers.Location?.AbsoluteUri ?? response.Content.Headers.ContentLocation?.AbsoluteUri;
+
+            result.Response.LastModified = response.Content.Headers.LastModified;
+            result.Response.Etag = response.Headers.ETag?.Tag;
```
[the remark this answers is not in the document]
Devereux-Henley: Will restore trim.

src/Hl7.Fhir.Core/Rest/Http/EntryToHttpExtensions.cs
Devereux-Henley: No. This was added while debugging and is no longer needed. Will replace with `Add`.

src/Hl7.Fhir.Core/Rest/Http/EntryToHttpExtensions.cs
Devereux-Henley: Did not see static `MediaTypeWithQualityHeaderValue.Parse`. Replaced this with a call to  `request.Headers.Accept.Add`.

src/Hl7.Fhir.Core/Rest/Http/EntryToHttpExtensions.cs
```diff
@@ -105,10 +105,7 @@ private static void setBodyAndContentType(HttpRequestMessage request, Resource d
 
             request.Content = new ByteArrayContent(body);
 
-            // MediaTypeHeaderValue cannot accept a content type that contains charset at the end, so that value must be split out.
-            var contentTypeList = contentType.Split(';');
-            request.Content.Headers.ContentType = new MediaTypeHeaderValue(contentTypeList.FirstOrDefault());
-            request.Content.Headers.ContentType.CharSet = System.Text.Encoding.UTF8.WebName;
+            request.Content.Headers.ContentType = MediaTypeHeaderValue.Parse(contentType);
```
ewoutkramer: Good find!  Didn't notice it - indeed constructor would not accept it.

mmsmits:
@Devereux-Henley Thanks so much for all your efforts on this. We have re-used a lot of your code, in a separate PR which now has (finally) been merged into develop-r4 and develop-stu3. If you like, please take a look to see if you like how we implemented it. We would appreciate your feedback very much. Hope you like it! Thanks again!

PR #519, opened 2017-12-14
Feature/461 validation with reg ex stu3

marcovisserFurore:


src/Hl7.Fhir.Specification/Validation/ElementDefinitionNavigatorExtensions.cs
```diff
@@ -37,9 +37,7 @@ public static bool IsPrimitiveValueConstraint(this ElementDefinition ed)
             //TODO: There is something smarter for this in STU3
```
ewoutkramer: We'll have to remove the "there is something smarter in STU3" if we don't put something smarter ;-)

PR #523, opened 2017-12-21
Fix null reference exception in ComplexTypeWriter.

jwdhaan:
Hi Ewout,
by adding this null check, the DispatchingWriter will throw an error in it's Serialize method:
"The FHIR serialization does not support arrays with empty (null) elements"


ewoutkramer:
Thanks, this bug was still present in the DSTU2 and STU3 versions too, so I cherry-picked your changes to the newer branches as well!

Issue #479, opened 2017-11-16
[FhirClient] Use search using POST


ewoutkramer:
Currently, search is only supported using url parameters.  For security reasons, you might want to put those in the body and POST the search (as specified in the spec here http://hl7.org/fhir/search.html#2.21.1.2).

kennethmyhra:
Referenced PR #526 is ready for review, will need some input on this. Added several SearchUsingPost methods to interface IFhirClient which are identical to the already existing Search methods, could probably be done in differently.

dougludlow:
We have a requirement for this on a current project. I was just looking into how to implement this in a fork, but it looks like it is currently under way. What is the current status of this? I see a couple of PRs. But no updates since early February.

dougludlow:
Oh, I see they were merged into develop and develop-stu3. I currently have v0.94 of Hl7.Fhir.Specification.STU3 and it doesn't appear to be there. Are there any plans for another release soon?

brianpos:
I'm coincidentally integrating this branch into our server today, not sure when the plan to release a new NuGet package is though

dougludlow:
@ewoutkramer any news on when the latest nuget package release will be?

dougludlow:
Oh sorry, @brianpos, I didn't realize you were also a maintainer. We're trying to decide whether we should fork the repo and build our own release or if we can wait for you guys to publish one. Do you have any insight into plans on creating a new release/nuget package?

ewoutkramer:
Hi @dougludlow, back from holidays & I am doing release planning today, so I'll let you know shortly!

ewoutkramer:
@dougludlow we are going to release the new 0.95 next week. Hope that's on time for you.

dougludlow:
@ewoutkramer yes, that will work for us, thank you!

ewoutkramer:
@dougludlow will be released today!

dougludlow:
@ewoutkramer awesome! Thank you!

dougludlow:
@ewoutkramer - Thanks so much for releasing this. I've got it installed and it appears to be working great. I had one question. I'm testing against https://vonk.fire.ly/ and the `SearchAsync<Patient>(params)` and `SearchUsingPostAsync<Patient>(params)` are pulling back entirely different results. Do you think it's a bug in `fhir-net-api` or just an issue with the test data on vonk?

ewoutkramer:
[...]

PR #526, opened 2018-01-09
WIP: Support search using POST

kennethmyhra:
PR for issue #479.
Added methods for executing search using POST
Can be tested by using end point at http://nde-fhir-ehelse.azurewebsites.net/fhir

### DIFF
--- a/src/Hl7.Fhir.Core/Rest/EntryToHttpExtensions.cs
+++ b/src/Hl7.Fhir.Core/Rest/EntryToHttpExtensions.cs
@@ -135,7 +135,7 @@ namespace Hl7.Fhir.Rest
                     if (!string.IsNullOrEmpty(bodyParameters)) bodyParameters += "&";
                     bodyParameters += $"{parameter.Name}={parameter.Value}";
                 }
-                body = Encoding.UTF8.GetBytes(bodyParameters);
+                body = Encoding.UTF8.GetBytes(Uri.EscapeDataString(bodyParameters));
                 request.ContentType = "application/x-www-form-urlencoded";
             }
             else
--- a/src/Hl7.Fhir.Core/Rest/EntryToHttpExtensions.cs
+++ b/src/Hl7.Fhir.Core/Rest/EntryToHttpExtensions.cs
@@ -57,13 +57,13 @@ namespace Hl7.Fhir.Rest
             
             if (entry.Resource != null)
             {
-                bool isSearch =
+                bool searchUsingPost =
                     interaction.Method == Bundle.HTTPVerb.POST
                     && (entry.HasAnnotation<TransactionBuilder.InteractionType>()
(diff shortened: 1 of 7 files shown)


src/Hl7.Fhir.Core/Rest/EntryToHttpExtensions.cs
```diff
@@ -118,6 +127,17 @@ private static void setBodyAndContentType(HttpWebRequest request, Resource data,
...
+            else if (isSearch)
+            {
+                string bodyParameters = null;
+                foreach(Parameters.ParameterComponent parameter in ((Parameters)data).Parameter)
+                {
+                    if (!string.IsNullOrEmpty(bodyParameters)) bodyParameters += "&";
+                    bodyParameters += $"{parameter.Name}={parameter.Value}";
+                }
+                body = Encoding.UTF8.GetBytes(bodyParameters);
+                request.ContentType = "application/x-www-form-urlencoded";
```
wmrutten: Doesn't the contenttype "x-www-form-urlencoded" imply that the parameters need to be url-encoded?

kennethmyhra: True, committed fix

wmrutten: Thanks, great job. Would be really awesome if you could also add a unit test for some special characters like <&>"'äëïo€ etcetera. Don't know if we do this for GET already, but I think we should. However that's another issue.

kennethmyhra:
Added test cases per request:
* FormUrlEncoding (Special characters)
* Method == POST
* ContentType == application/x-www-form-urlencoded
* Url Suffix == _search

 @wmrutten see commit 381a5d2

src/Hl7.Fhir.Core/Rest/EntryToHttpExtensions.cs
```diff
@@ -53,9 +54,17 @@ internal static class EntryToHttpExtensions
             {
                 request.Headers["Prefer"] = bodyPreference == Prefer.ReturnMinimal ? "return=minimal" : "return=representation";
             }
-
+            
             if (entry.Resource != null)
-                setBodyAndContentType(request, entry.Resource, format, CompressRequestBody, out body);
+            {
+                bool isSearch =
```
ewoutkramer: The name of this variable put me on the wrong foot.  Shouldn't this be called  `bools isSearchUsingPost`?

kennethmyhra: Agree, maybe lose the is-part altogether and call it searchUsingPost.

ewoutkramer:
Hi Kenneth....I applied your changes to STU3 and did some updates to make it compile....

Could you review my changes so we can have your code in STU3 too?

https://github.com/ewoutkramer/fhir-net-api/pull/530

kennethmyhra:
Will do

PR #532, opened 2018-01-29
Develop stu3

jahansha:
I would like to enhance the FHIRClient Search. 
Currently, Search  it is treated as HTTP Get with parameters passed in the Url as query string. 
I  have modified  my local version to use Custom HTTP headers.
This is because we like to hide all the parameters passed in query string.

This is OK for now, but I like to handle Search, using POST method


jahansha:
Thanks

PR #533, opened 2018-01-31
Develop stu3

jahansha:


ewoutkramer:
We have chosen to use this solution (#526) instead. Closing this PR.

PR #546, opened 2018-02-28
fix issue #544, DateTimeOffset cannot convert to DateTime directly, s…

### LINKED ISSUE #544: Date.ToDateTime failed to convert "1976-12-12"
Following code failed at ToDateTime():

private static void TestDateToDateTime()
		{
			var patient = new Patient() {BirthDate = "1976-12-12"};
			var dt = patient.BirthDateElement.ToDateTime();
			Console.Write(dt.Value);
		}

Hl7.Fhir.STU3 version 0.94.0

lbcsy:
This commit is to fix #544 . The bug was due to an invalid cast from DateTimeOffset to DateTime. It should use DateTimeOffset.DateTime to convert a DateTimeOffset to a DateTime.

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/Serialization/ConversionTest.cs
+++ b/src/Hl7.Fhir.Core.Tests/Serialization/ConversionTest.cs
@@ -108,7 +108,13 @@ namespace Hl7.Fhir.Tests.Serialization
             Assert.AreEqual(new DateTimeOffset(2011,3,4,16,45,33, TimeSpan.Zero),ins5.Value);
         }
 
-        [TestMethod]
+	    [TestMethod]
+	    public void TestStringToDateTime()
+	    {
+		    Assert.AreEqual(PrimitiveTypeConverter.ConvertTo<DateTime>("1976-12-12"), new DateTime(1976, 12, 12));
+	    }
+
+	    [TestMethod]
         public void TestStringToUri()
         {
             var result = PrimitiveTypeConverter.ConvertTo<Uri>("http://www.nu.nl/test");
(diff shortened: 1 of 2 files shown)


ewoutkramer:
I noticed there is also a XmlConvert.ToDateTime().  Which is better?

XmlConvert.ToDateTimeOffset().DateTime or XmlConvert.ToDateTime()?  Are we sure about that or do we need to explore a bit further?

ewoutkramer:
For discussion:

`XmlConvert.ToDateTimeOffset("2003-02-02T15:00:00+03:00").DateTime` returns `2-2-2003 15:00:00`

`XmlConvert.ToDateTime("2003-02-02T15:00:00+03:00", XmlDateTimeSerializationMode.Utc)` returns `2-2-2003 12:00:00`
`XmlConvert.ToDateTime("2003-02-02T15:00:00+03:00", XmlDateTimeSerializationMode.Unspecified)` returns `2-2-2003 13:00:00`

What is the desired behaviour

ewoutkramer:
Fortunately, all of these return the same value:

XmlConvert.ToDateTimeOffset("2003-02-02").DateTime.Dump();
XmlConvert.ToDateTime("2003-02-02", XmlDateTimeSerializationMode.Utc).Dump();
XmlConvert.ToDateTime("2003-02-02", XmlDateTimeSerializationMode.Unspecified).Dump();

which is midnight 2003-02-02

ewoutkramer:
Since DateTimeOffset and DateTime *do* have timezones, but the source string does not, one has to choose what to do with the timezone.  It seems safe to take Zulu, since anything else would maybe change the date  (If you'd convert 2019-01-01 just after midnight on a localtime +07:00, you'd be back in 2018 - which when you are talking about "just" 2019-01-01 probably isn't what you wanted).

Issue #423, opened 2017-09-13
[Validation] Internal references within contained resources are not validated properly


ArdonToonstra:
In the attached FHIR list are contained resources that reference to one another using "#id". The validator gives the error: Instance failed constraint ref-1 "SHALL have a contained resource if a local reference is provided".

The validator doesn't seem to resolve the contained resource if it is inside a contained resource.
[MedicationOverview.txt](https://github.com/ewoutkramer/fhir-net-api/files/1299175/MedicationOverview.txt)

ewoutkramer:
I think the problem is in my implementation of %resource , as used in this invariant (ref-1):
      reference.substring(1).trace('url') in %resource.contained.id.trace('ids')))

The way I have implemented it, contained resources can only refer to contained resources inside the contained resource, not to sibling contained resources - which is obviously wrong. 

I will double check the code, and create a new unit-test in the FHIRPath tests to make sure.

ewoutkramer:
In fact %resource would always be the root resource, or the resource in a Bundle you are currently validating (the spec says nothing about nested Bundles though). %resource never moves down to contained resources within contained resources (which is disallowed by the way). This I will need to fix.

marcovisserFurore:
After some research, this problem only arise at STU3, in DSTU2 the validator will not complain. In the attached file there is simplified example of this issue (a List with a contained Observation and a contained Patient).
[observation-list.txt](https://github.com/ewoutkramer/fhir-net-api/files/1606617/observation-list.txt)

brianpos:
Yep, that's a good example of the bug.

PR #549, opened 2018-03-02
Bugfix/423 internal references validate incorrectly r4

marcovisserFurore:
Solves issue #423 in R4

Not sure, if this fix should also be in this new r4 branch. But I think yes..


ewoutkramer:
auto-included in R4 by a merge from STU3->R4

PR #563, opened 2018-04-04
Added to ModelInfo AddToModelInfo method

marcocaminiti:
the method AddToModelInfo allow to manage custom fhir resource with BaseFhirParser

### DIFF
--- a/src/Hl7.Fhir.Core/Model/ModelInfo.cs
+++ b/src/Hl7.Fhir.Core/Model/ModelInfo.cs
@@ -441,7 +441,24 @@ namespace Hl7.Fhir.Model
         {
             return CanonicalUriForFhirCoreType(type.GetLiteral());
         }
+      
+        /// <summary>
+        /// Adds a custom FHIR class to Model Info.
+        /// </summary>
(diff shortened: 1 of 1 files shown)


ewoutkramer:
This will fix serialization and deserialization - but everything in the specification library (and probably more) like validation and snapshot generation will not work, even if you would create additional StructureDefinitions for the new types.  This is due to the fact that the FhirKnownTypes enumeration cannot be expanded without regenerating the source code - and quite some code depends statically on that enumeration.

I don't mind thinking about how to make this more flexible, but are reluctant to pull a partial solution.....

kfrancis:
@ewoutkramer Do you mean how SupportedResources is generated when that static T4 template is executed?

brianpos:
You can do this stuff locally in your own branch/build.
Issue for me is no one outside your lib can work with the resource, or know what to do when they see it.
I wouldn't like this in the core lib. It doesn't improve standard interop.

ewoutkramer:
I am working on a more general parsing framework for custom resources - probably extending that into the POCO parsers. For now, will not pull a partial solution.

Issue #545, opened 2018-02-28
Support XML/JSON serialization to in-memory DOM


wmrutten:
If a client needs to perform some further processing on the generated serialization output, then currently it is necessary to first serialize to XML/JSON string, then deserialize the output back to an in-memory XML/JSON object model.

For example, the Forge XML view parses the serialized XML output string back into an XML DOM in order to perform syntax highlighting. Currently, this is quite a CPU- and memory-intensive operation.

Would be more efficient if the API would also support serialization to an in-memory DOM. This would allow clients to skip the intermediate string serialization altogether.

Also see: https://stackoverflow.com/a/1347364

tstolker:
Hi Michel,

There is no need to first serialize; given that you can use client.LastBodyAsText.

Best regards,

Theo.

brianpos:
Or use the streaming serializers with no string or the inavigable interface that the xsl can work on Ewout created.

wmrutten:
I don't think that will help. In Forge, the edited profile is internally represented as a FHIR (StructureDefinition) PoCo. The application needs to display the model as formatted XML (for the preview tab). The rendering logic takes an XML DOM and generates a XAML DOM that is injected into the view. This operation needs to be fast and GC friendly. Therefore, Forge needs to efficiently serialize a FHIR PoCo to an XML DOM, i.e. from one object model to another. Preferably directly, without any intermediate steps. I'm proposing for the FHIR API to implement and expose such a serialization method. as it would improve efficiency in scenario's such as described. The article on SO explains how this can be implemented.
And I would also like to have serialization from FHIR PoCo to JSON DOM.

wmrutten:
Brian explained how I can leverage the Xml/Json navigators, so I'm going to try that. Thanks B!

Issue #543, opened 2018-02-27
remove ListSummaries() from IConformanceSource 


mmsmits:
It's little used and breaks existing code.

PR #564, opened 2018-04-09
ISummarySource & Serialize to DOM

wmrutten:
* #543 Remove ListSummaries() from IConformanceSource
previous build introduced new method IConformanceSource.ListSummaries(), however this breaks binary compatibility => move this method to a separate new interface ISummarySource.

* #545 Support XML/JSON serialization to in-memory DOM
Implement public methods to serialize to Xml/Json document
(FhirXmlSerializer, FhirJsonSerializer)


wmrutten:
Added some more commits, most importantly: Fix contentreference resolving

Issue #536, opened 2018-02-12
FhirClient.Transaction() method incorrectly POSTs to FHIR Base URl with trailing slash 


tstolker:
My server is the forCare MHD endpoint. When I want to post a document, it must be sent as a Transaction. However, the transactions fails due to the fact that the FhirClient append a trailing slash to the Base Url before sending the request.

According to the vendors interpretation of https://www.hl7.org/fhir/DSTU2/http.html#general, that is incorrect: 

2.1.0.1 Service Root URL 
The Service Root URL is the address where all of the resources defined by this interface are found. The Service Root URL takes the form of

http(s)://server{/path}
**The path portion is optional, and does not include a trailing slash.**

I had hoped I could fix this by removing the trailing slash from the url in the OnBeforeRequest handler, but unfortunately there the Url cannot be rewritten.

I am using the DSTU2 version of the API Hl7.Fhir.DSTU2.0.94.0.

tstolker:
I was able to fix this by changing a line of code in FhirClient.TransactionAsync:

var tx = new TransactionBuilder**(new Uri(Endpoint.OriginalString.TrimEnd('/'))**).Transaction(bundle).ToBundle();

ewoutkramer:
I feel this should be the responsibility of the TransactionBuilder, not the client.

I see there's some special handling in Get(), and then there's newRestUrl(). I think we should fix it there.

PR #566, opened 2018-04-10
fix: remove trailing slash in case of Transaction

marcovisserFurore:
This fix is part of issue #536


ewoutkramer:
Refusing the PR, since this change has been integrated in STU3 by a direct merge from DSTU2->STU3.

Issue #524, opened 2018-01-02
FhirPath does not resolve correctly for some elements 


khronosX:
Hi there,

This is related to issue #522. Since it was closed by the owner I couldn't reactivate the issue.

Some of the FHIR Path defined in ModelInfo.cs does not resolved correctly.

```new SearchParamDefinition() { Resource = "StructureDefinition", Name = "valueset", Description = @"A vocabulary binding reference", Type = SearchParamType.Reference, Path = new string[] { "StructureDefinition.snapshot.element.binding.valueSetUri", "StructureDefinition.snapshot.element.binding.valueSetReference", }, Target = new ResourceType[] { ResourceType.ValueSet, }, XPath = "f:StructureDefinition/f:snapshot/f:element/f:binding/f:valueSetUri | f:StructureDefinition/f:snapshot/f:element/f:binding/f:valueSetReference", Expression = "StructureDefinition.snapshot.element.binding.valueSet[x]" },```

```valueSetUri``` and ```valueSetReference``` are not valid property. It should be ```valueSet```.

This is in STU3 branch.

Thanks

brianpos:
See my note on #522

cknaap:
I searched the ModelInfo.SearchParameters on the develop-stu3 branch, but none of the Expression fields contain [x].
Has this already been corrected in the meantime?

ewoutkramer:
This is still the case in the DSTU2 branch. Will make a fix.

PR #576, opened 2018-04-12
fix: Fix for #524, where the Modelnfo search parameters could end in [x]

ewoutkramer:
Addresses #524


marcovisserFurore:
This solves issue #524

PR #530, opened 2018-01-19
Feature/479 post based search

ewoutkramer:
Small updates to Kenneth's PR to make it work under STU3.

Issue #213, opened 2016-06-06
Enhancement: search with POST


mbaltus:
The specs offer the possibility to search with a POST where the parameters are encoded in a application/x-www-form-urlencoded body. The API only searches with the standard GET method. Could you add this functionality?

ewoutkramer:
I am pretty sure we have this now, as we even fixed bugs on it (#597). Sorry for not flagging this issue as fixed earlier!

PR #587, opened 2018-04-23
Remove escaped characters from POST body on Search

ryankelley:
In regards to #213, #479 and #530 which was recently released. The current implementation encodes and escapes the body on the request. This is not an apparent HTTP standard and breaks the `x-www-form-urlencoded` processing on FHIR servers I have tested with several servers from the sample list, as well as local test servers.

I first noticed this when connecting to several test servers after converting our usages from GET to POST and having wildly different results.

If this isn't the correct target branch, please advise.


ewoutkramer:
Thanks for this!  @kennethmyhra - as you wrote this, can you take a look too?

I'd prefer to have this PR against the develop branch, if possible.

ryankelley:
@ewoutkramer Would you prefer this against `develop` or `develop-stu3`?

kennethmyhra:
@ewoutkramer I haven't tested this PR, but I agree, my implementation was escaping the whole body instead of escaping the key and value separately.

@ryankelley According to how I read the W3C specification we should escape both the key and value:
1. The control name and value should be escaped. (...)

https://www.w3.org/TR/html401/interact/forms.html#h-17.13.4.1

Something along the lines of:
`bodyParameters += $"{Uri.EscapeDataString(parameter.Name)}={Uri.EscapeDataString(parameter.Value.ToString())}";`

In addition the specification says that white spaces should be replaced with `+`, but Postman replaces white spaces with `%20` the same way `Uri.EscapeDataString` does so this might have changed in later years.

ewoutkramer:
When done, please against "develop", unless it used dstu2/stu3 specific code (which I expect it does not), we can merge it from DSTU2->STU3->R4 without problems

ryankelley:
I just created a new Pull Request #597 to replace this, it was much cleaner that way.

Issue #556, opened 2018-03-19
Fhir validation against profile when running multi threaded fails


ArneHB:
Running the validation of the Fhir profile causes errors when running multi-threaded. If you run it in 5 threads then usually between 1 to 3 threads successfully validates. This is not a problem for single threads. I think I managed to reproduce the error we get in our production system in the tests. Past in the code I've added below into the BasicValidationTests.cs file and try to run it as debug. I've been doing some debugging but I'm not able to figure out what's going on.

What I found is in the LocalTerminologyService.cs, the vs.FindInExpansion returns null which causes the "... does not exist in valueset '..'". The FindInExpansion looks through the values in "Contains" in ValueSet.cs. The "Contains" is populated in ValueSetExpenader.cs ...Contains.AddRange(..).

I think the problem is located somewhere within CachedResolver.cs Get() method (which gets the value for "Contains"). I have really no idea why this should be a problem at the first time so any help here would be greatly appriciated. (The test seems to fail more often when I run it as debug then normal).

       public async System.Threading.Tasks.Task<OperationOutcome> RunMultipleThreads()
        {
            var careplanXml = File.ReadAllText("TestData\\validation\\careplan-example-integrated.xml");
            var cpDoc = XDocument.Parse(careplanXml, LoadOptions.SetLineInfo);
            return _validator.Validate(cpDoc.CreateReader());
        }

        [Fact]
        public void RunXsdValidationMultiThread()
        {
            var tasks = new List<System.Threading.Tasks.Task<OperationOutcome>>
            {
                System.Threading.Tasks.Task.Run(() => RunMultipleThreads()),
                System.Threading.Tasks.Task.Run(() => RunMultipleThreads()),
                System.Threading.Tasks.Task.Run(() => RunMultipleThreads()),
                System.Threading.Tasks.Task.Run(() => RunMultipleThreads()),
                System.Threading.Tasks.Task.Run(() => RunMultipleThreads())
            };
            System.Threading.Tasks.Task.WaitAll(tasks.ToArray());

            int successes = 0;
            foreach (var task in tasks)
            {
                var report = task.Result;
                if (report.Success) successes++;
            }

            Assert.Equal(5, successes);
        }

ArneHB:
[...]

PR #588, opened 2018-04-26
test: fhirpath-parallel

cknaap:
FhirPath evaluation gets mixed up when used heavily in parallel.
I have created a unittest to show this.
The unittest runs for about 3 min, because it loads all DataElement resources from specification.zip.
Unittest contains comments on what it does.

[FhirPath-Parallel-TestResults-BothFailing_201804226_1000.txt](https://github.com/ewoutkramer/fhir-net-api/files/1950281/FhirPath-Parallel-TestResults-BothFailing_201804226_1000.txt)

### DIFF
--- a/src/Hl7.FhirPath.Tests/HL7.FhirPath.Tests.csproj
+++ b/src/Hl7.FhirPath.Tests/HL7.FhirPath.Tests.csproj
@@ -14,14 +14,17 @@
 
   <ItemGroup>
     <ProjectReference Include="..\Hl7.Fhir.Core\Hl7.Fhir.Core.csproj" />
+    <ProjectReference Include="..\Hl7.Fhir.Specification\Hl7.Fhir.Specification.csproj" />
     <ProjectReference Include="..\Hl7.FhirPath\Hl7.FhirPath.csproj" />
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="FluentAssertions" Version="5.3.0" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.6.1" />
     <PackageReference Include="MSTest.TestAdapter" Version="1.2.0" />
     <PackageReference Include="MSTest.TestFramework" Version="1.2.0" />
-    <PackageReference Include="System.ValueTuple" Version="4.3.0" />
+    <PackageReference Include="System.Threading.Tasks.Dataflow" Version="4.8.0" />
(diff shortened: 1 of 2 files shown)


brianpos:
I've seen this too.

ewoutkramer:
This is related to #556 - it's the same problem.

We've now identified three places in the API where we alter the data from a resource (StructureDefinition, ValueSet) after we have gotten a reference to it from a IConformanceSource.  In particular, if this is a reference returned from the CachedResolver (which is often the case), we're modifying the cached copy. This is desired in many cases, but can also lead to race conditions, since in many cases the code looks like this:

```csharp
if(cachedResource.someKindOfModificationHasBeenDone == false)
{
      cachedResource.doTheModification();
}
```

In all three cases we've found, the `doTheModification()` is a lengthy process, and since there's no locking going on here, multiple threads will simultaneously enter the body of the if....

These cases are:

* Generating a snapshot on a cached StructureDefinition
* Generating an expansion on a cached ValueSet
* Generating a compiled FhirPath expression on a cached StructureDefinition's elements.

We should go out and add a locking strategy to these cases.

ewoutkramer:
It was probably not the same problem after all.   We think the cause is in the `Invoke()` function below. This is called as part of the lambda that gets produced when you compile a statement.

Look at what happens to `lastResult`, this is a local variable, but it's captured in the returned function. And then just set and immediately returned. If you share the compiled expression (which you should) across threads, this local becomes shared across the threads, turning it into a global variable.  Ouch.

```csharp
 public static Invokee Invoke(string functionName, IEnumerable<Invokee> arguments, Invokee invokee)
        {
            Func<Closure, IEnumerable<IElementNavigator>> boundFunc = (ctx) => invokee(ctx, arguments);
            IEnumerable<IElementNavigator> lastResult = null;

            return (ctx, _) =>
            {
                //  if (lastResult != null) return lastResult;

                try
                {
                    lastResult = boundFunc(ctx);
                    return lastResult;
                }
                catch (Exception e)
                {
                    throw new InvalidOperationException("Invocation of '{0}' failed: {1}".FormatWith(functionName, e.Message));
                }
            };
        }
```

I will modifying a captured variable a deadly sin from now on. Mea culpa.

ewoutkramer:
By the way, this bit of code was left after a refactoring and is competely unnecessary, we can just remove `lastResult`.

PR #590, opened 2018-04-27
Support: ignoreCase option for ParseLiteral

michelemottini:
Useful when parsing token search parameters, that must be case-insensitive

Also: moved the EnumUtility tests into Support.Tests - matching where the class is

(Is this the correct branch to merge to get into the next release of Hl7.Fhir.Support ?...)

### DIFF
--- a/src/Hl7.Fhir.Support.Tests/Utils/EnumMappingTest.cs
+++ b/src/Hl7.Fhir.Support.Tests/Utils/EnumMappingTest.cs
@@ -11,7 +11,7 @@ using Hl7.Fhir.Utility;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Hl7.Fhir.Model;
 
-namespace Hl7.Fhir.Tests.Introspection
+namespace Hl7.Fhir.Support.Tests.Utils
 {
     [TestClass]
     public class EnumMappingTest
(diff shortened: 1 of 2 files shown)


brianpos:
Some valuesets are case sensitive and shouldn't be treated like this. Even though its bad, I've heard of some where the code is only different by case, which would break this.
I'm very cautious of this change.

michelemottini:
The [spec](http://hl7.org/fhir/search.html#token) explicitly say that token searches must be case-insensitive: 'For tokens, matches are literal (e.g. not based on subsumption or other code system features), but not case sensitive.' (something I never noticed up until now!), so  I think a case-insensitive version of `ParseLiteral()` is needed to handle that case ....of course users of the library are free to use it or not, the default remains case-sensitive.

ewoutkramer:
This is ambiguous in DSTU3 - we know some tokens are case sensitive and some are not - depending on the codesystem(!). In R4 this bit of text was added to the description of token search parameters (http://build.fhir.org/search.html#token):

> Note: There are many challenging issues around case senstivity and token searches. Some code systems are case sensitive (e.g. UCUM) while others are known not to be. For many code systems, it's ambiguous. Other kinds of values are also ambiguous. When in doubt, servers SHOULD treat tokens in a case-insensitive manner, on the grounds that including undesired data has less safety implications than excluding desired behavior. Clients SHOULD always use the correct case when possible, and allow for the server to perform case-insensitive matching.

Now, the ParseLiteral() functions are called for enums - which are generated from HL7 mandated valuesets and codesystems, which are case-sensitive. So, in practice, our current case-sensitive match is fine.  

However, if this brings challenges to developers who have manually crafted their own enums based on mandatory valuesets derived from non-case-sensitive codesystems, it would be a waste that they need to copy & paste the current code.  So, I have no problem providing an overloaded version with an argument (much like String.compare) to change the case-sensitivity behaviour.

ewoutkramer:
@michelemottini, just against the "develop" branch would be fine.

PR #597, opened 2018-05-02
Remove Escaped characters from POST body on search

ryankelley:
In regards to #213, #479 and #530 which was recently released. The current implementation encodes and escapes the body on the request. This is not an apparent HTTP standard and breaks the x-www-form-urlencoded processing on FHIR servers I have tested with several servers from the sample list, as well as local test servers.

I first noticed this when connecting to several test servers after converting our usages from GET to POST and having wildly different results.

This is a replacement Pull Request for #587

Squashed commit of the following:

commit 0e127ee67c2f5e86628d86a7ccb4185153a83d7d
Author: Ryan Kelley <ryan.kelley@royaljay.net>
Date:   Wed May 2 15:53:22 2018 -0600

    Fixing tests

commit 2ace43ac7403db2515a6448f87bfa0eda33ac1e2
Author: Ryan Kelley <ryan.kelley@royaljay.net>
Date:   Tue May 1 13:50:39 2018 -0600

    Add encoding to property names on form post

commit 4b88a38c64ac97739165b0e9b6e5fdca068f8c1c
Author: Ryan Kelley <ryan.kelley@royaljay.net>
Date:   Mon Apr 23 16:11:10 2018 -0600

    Updated to escape the values, but not the entire body

commit a59a2cfd64ba1972820e9a8caf11d70aa56fff30
Author: Ryan Kelley <ryan.kelley@royaljay.net>
Date:   Mon Apr 23 15:19:47 2018 -0600

    Remove escaped characters from POST body on search re: #479

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/Rest/TransactionBuilderTests.cs
+++ b/src/Hl7.Fhir.Core.Tests/Rest/TransactionBuilderTests.cs
@@ -14,6 +14,7 @@ using Hl7.Fhir.Model;
 using System.Collections.Generic;
 using System.Text;
 using System.Net;
+using System.Runtime.Remoting.Channels;
 
 namespace Hl7.Fhir.Test
 {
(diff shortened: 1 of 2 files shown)


brianpos:
The StringBuilder would be a better thing to be using here, as it doesn't do as much string manipulation, and also the Escape routine should be for the Uri not Data (subtle differences in what it replaces)
```
request.ContentType = "application/x-www-form-urlencoded";
                StringBuilder postData = new StringBuilder();
                foreach (Parameters.ParameterComponent parameter in ((Parameters)data).Parameter)
                {
                    if (postData.Length > 0)
                        postData.Append("&");
                    postData.Append(Uri.EscapeUriString(parameter.Name));
                    postData.Append("=");
                    postData.Append(Uri.EscapeUriString(parameter.Value?.ToString()));
                }
                if (postData.Length > 0)
                    body = Encoding.UTF8.GetBytes(postData.ToString());
                else
                    body = null;
```
I've tested this against my server and works fine.

brianpos:
A better solution would be to use the class System.Net.Http.FormUrlEncodedContent to do all the work
(which I used in my unit test to check things)

ewoutkramer:
Pulled as part of #614

Issue #611, opened 2018-06-04
Bug: Snapshot Generator fails for derived profiles with sparse constraints on _some_ existing named slices


wmrutten:
Originally reported by a Forge user, issue seems to be caused by a bug in the snapshot generator:

https://chat.fhir.org/#narrow/stream/4-implementers/subject/Forge.2018.2E6.20validation.20errors.20when.20deriving.20profiles

PR #613, opened 2018-06-05
#611 Snapshot Generator fails for derived profiles with sparse constraints on _some_ existing named slices

wmrutten:
Backported from develop-stu3


ewoutkramer:
Fixes #611

PR #639, opened 2018-07-16
Upgraded to .Net Framework 4.7.1, .NET Standard 2.0, latest NuGet

IvanFarkas:
Upgraded to .Net Framework 4.7.1, .NET Standard 2.0, latest NuGet packages.
Added global.json


wmrutten:
Hi Ivan, great work, thank you for contributing to the API!

Today, we had a discussion with our team about supporting netstandard 2.0. To reduce impact on existing customers, we'd prefer to add netstandard2.0 as a new and separate compilation target, in addition to the existing netstandard1.1 and netstandard1.3 targets. This allows us to safely evaluate the new target without affecting existing (external) consumers.

Do you think you could update your pull request accordingly?

IvanFarkas:
Done

wmrutten:
Great work Ivan, thank you!

I noticed that you also updated some external dependencies:

* OWIN 3.1 => 4.0
According to the release notes, OWIN 4.0 introduces a breaking change:
https://github.com/aspnet/AspNetKatana/wiki/Roadmap
This might be a good time to update the OWN library dependency. But could this introduce issues for consumers?

* NewtonSoft.Json 9.0 => 11.0
Again, may be a good time to update this dependency, but I don't think this is strictly necessary,  considering the API does not specifically depend on version 11.0. Also, this forces clients to update their own dependency.

Last, we should carefully think about the dependency to the .NET framework. Bumping this dependency forces all framework (desktop) clients to update to the same version. Also, even though version 4.7.1 introduces full compatibility with netstandard 2.0, it seems that clients (such as Forge) would actually need version 4.7.2 as this fixes some critical issues. We should carefully analyze this.

Alternatively, we could pull your changes into a separate branch, which I could then try and evaluate in Forge.

wmrutten:
@brianpos maybe you can also review and comment this pull request?

IvanFarkas:
Updated packages except for **OWIN** and **NewtonSoft.Json**

ewoutkramer:
Can we retain 4.5 as a target, or am I being too conservative?

IvanFarkas:
I think it’s very conservative.

ewoutkramer:
What would we gain from going to 4.7? Obviously, we don't need to upgrade because we need features from the newer platforms (which would be a good reason), or am I mistaken?

ewoutkramer:
Performance improvements might be a reason.  And compatibility with netstandard 2.0 is what I am reading - which means we might ditch the net472 target completely?

IvanFarkas:
In my humble opinion targeting .NET Core 2.1 and above would be the primary target.
We might have a .Net Framework Legacy branch or fork for the legacy folks.
.NET Core 2.1 is a prerequisite for a Linux docker container.
New features ONLY added to .NET Core 2.1 branch/fork. This will motivate the legacy folks to join the modern world!

wmrutten:
@IvanFarkas Please mind that Forge is a WPF application that requires .NET framework (for now...). Therefore the API should keep providing a netframework target and/or a netstandard 2.0 target (fully supported by netframework 4.7.2).

aagarwal10:
Any plans to merge and release this to public nuget for .NET Core 2x versions/ .NET Standard 2.0?

ewoutkramer:
@aagarwal10 - yes, we are planning to work on this this week.

In fact, what we are going to do is this - we're basically going to integrate the changes of this PR, but we've seen that it is quite easy to support the following platforms:

* Netstandard 2.0 - this will be the newest "to go" target that has all the features.
* Netstandard 1.1 - this makes sure our stuff still works on .NET Core 1.x and even older Mobile platforms. Might miss some features that are not available (e.g. XSD schema validation)
* dotnet45 - yes - still. It's very easy to support (since the API surface is almost the same as the standard 2.0 surface), and if we drop it some corporate users of the API - who cannot upgrade - will have to switch to Netstandard1.0 - and thus lose functionality.

We think can simplify the conditional compilations in the source code to have just two categories: Netstandard 1.1 and (Netstandard 2.0 or dotnet45).  So we can do some cleanup too.

Also, this PR upgrades some of our dependencies. It's indeed time to do so, so we'll integrate that change too.

IvanFarkas:
Fantastic!

ewoutkramer:
Pulled as part of #699.  Note: the WebApi bits were not pulled, since these are only valid for R4.

aagarwal10:
@ewoutkramer - First of all, this AWESOME !!! 
Also, Is this change also published via nuget package ?

ewoutkramer:
Not yet - the summer release (https://github.com/ewoutkramer/fhir-net-api/projects/3, which this change is part of) has a number of substantive changes that we like to test internally before we push a beta (call us over cautious if you like ;-)).  The workflow is like this:

* We have are currently doing internal alpha releases, testing the summer release (probably published as 1.0 later) against our products (Simplifier, Forge, Vonk).
* When no problems are encountered, we create a first public beta release (I hope in early october) on NuGet.
* After some betas we push the final version to NuGet.

Most of the time, we have a much shorter cycle, but because of the changes to serialization (documented here (http://docs.simplifier.net/fhirnetapi/parsing/intro-to-elementmodel.html) we are a bit more cautious than normal.

The reason we do this is that we expect the API to still have bugs that require us to be close at hand - so we're only bothering our direct colleagues with it - hence the private alphas.

If you feel brave, I can send you a zip with the alpha release ;-)

Issue #650, opened 2018-07-25
ArtifactSummarySource should also extract root element definition from StructDef


wmrutten:
On the HL7 FHIR website, each datatype and resource page starts with a brief introduction text that describes the meaning/intent of the model. The introduction text is taken from the definition property of the root element of the underlying StructureDefinition.

The `ArtifactSummarySource` should also harvest the root element definition, so clients can display the same familiar description as visible on the official FHIR spec website.

PR #651, opened 2018-07-25
#650 ArtifactSummarySource should also extract root element definition from StructDef

wmrutten:
Extend the ArtifactSummaryHarvesters.StructureDefinitionSummaryProperties class to also extract the definition text from the root element definition in the StructureDefinition, so clients can display the same resource/datatype descriptions as on the official HL7 FHIR spec website.


ewoutkramer:
Can you please create a DSTU2 PR for this as well?

wmrutten:
Yes, of course!
https://github.com/ewoutkramer/fhir-net-api/pull/658

wmrutten:
No prob, I fixed the merge conflicts.

Issue #552, opened 2018-03-07
Validator uses wrong regex for elements of type oid


mbaltus:
When validating a resource with a field of type oid, I get this error message:

Primitive value 'urn:oid:1.2.840.113674.514.212.200' does not match regex '(0|[1-9][0-9]\*)(\.(0|[1-9][0-9]\*))\*'

The regex for oid, according to the specs is: urn:oid:\[0-2\](\.[1-9]\d\*)+, but in specification.zip it has the regex the validator uses.

marcovisserFurore:
There is a difference indeed between the specification (https://www.hl7.org/fhir/datatypes.html#oid) and the specification.zip. We will create a gforge ticket for this. 
BTW, the oid in this issue (1.2.840.113674.514.212.200) is valid for both regexs.

ewoutkramer:
This is now [GF#16023](https://gforge.hl7.org/gf/project/fhir/tracker/?action=TrackerItemEdit&tracker_item_id=16023&start=0)

ewoutkramer:
Let's re-open and fix this expression locally, as suggested by Elliot Silver in the referenced gForge issue. Should be applied to both DSTU2 + STU3.

PR #654, opened 2018-07-30
fix: other regex for Oid. See also https://gforge.hl7.org/gf/project/…

marcovisserFurore:
This fix solves issue #552 (DTSU2)


ewoutkramer:
Fixes #552.

PR #661, opened 2018-08-08
rewrite: used Lazy<> to avoid the cost of building setter/getter DSTU2

ewoutkramer:
used Lazy<> to avoid the cost of building setter/getter expressions when not using poco's


wmrutten:
Can we expose the `PropertyInfoExtensions` class from the API by making it public? Forge also uses dynamic property accessors and could benefit from similar optimizations.

ewoutkramer:
@wmrutten - I have moved the class to the Hl7.Fhir.Utility namespace and made it public.

ewoutkramer:
Closing this PR, as #663 is a superior solution.

PR #663, opened 2018-08-13
Faster generation of property getters and setters with reflection emit

brendankowitz:
Hi @ewoutkramer, great optimization, we can also avoid the penalty of `Lambda.Compile()` by using a little bit of lightweight codegen. Given there is a known context around the models it seems relatively straight forward to do.

By calling Compile vs ILGen I was getting this type of perf to generate 100,000 getters:
```
00:00:00.8095830 (Generated using ILGen)
00:00:19.8382566 (Generated using Expression.compile)
```

### DIFF
--- a/src/Hl7.Fhir.Core/Hl7.Fhir.Core.csproj
+++ b/src/Hl7.Fhir.Core/Hl7.Fhir.Core.csproj
@@ -40,6 +40,8 @@
     <PackageReference Include="System.ComponentModel.Annotations" Version="4.3.0" />
     <PackageReference Include="System.Net.Http" Version="4.3.1" />
     <PackageReference Include="System.Net.Requests" Version="4.3.0" />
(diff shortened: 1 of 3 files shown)


wmrutten:
Hi @brendankowitz, thank you very much, this certainly looks like a great improvement!

ewoutkramer:
With this gain in generation speed I think #661 is not necessary anymore.

brianpos:
Is that in generating the getter setter, or calling them?
Checking if this is startup improvement, or operational improvement.

ewoutkramer:
By generating getters/setters we improved operational improvement - the actual getting and settings, but we introduced quite some startup delay - the generation of the getters/setters using Lambda.Compile took some time.  This new PR actually circumvents the lambda AST compiler and generates IL directly (since it's just a simple get/set) - hence it decreases the start-up cost that we introduced when moving to generated setters/getters.

Issue #662, opened 2018-08-13
DirectorySource - optimize locking strategy


wmrutten:
DirectorySource currently uses a standard lock to implement thread-safe access. The lock ensures that the collection will be initialized on a single thread; other threads will block and wait for completion. All reads also need to be synchronized using the lock.

Suggestion for improvement:
* Use internal Lazy<> to load artifact summaries on demand; will synchronize (re-)load
* Refresh method assigns a new internal Lazy instance (atomic)
* All read methods call Lazy<>.Value, which takes care of synchronization
* Remove locks (now redundant)

PR #666, opened 2018-08-15
fix: #662 - DirectorySource - optimize locking strategy

wmrutten:
Implement #662 DirectorySource - optimize locking strategy

Benefits:
- Read access is now lock-free
- Reduce complexity, facilitate maintenance

Note: this change required a minor breaking change, as the ArtifactSummaryGenerator class is no longer static. However this is unlikely to affect many consumers, and compilation errors are trivial to fix.


ewoutkramer:
Too big for me to pull at this stage, I'll pull after we have done the first alpha.

wmrutten:
Sure, no problem, this is just an optimization. Seems safe, but let's make sure this doesn't break anything. I can create a local API test build and evaluate the behavior in Forge.

PR #672, opened 2018-08-15
rewrite: #670 DifferentialTreeConstructor - expose static default instance

wmrutten:


src/Hl7.Fhir.Specification/Specification/Snapshot/DifferentialTreeConstructor.cs
```diff
@@ -64,6 +62,9 @@ namespace Hl7.Fhir.Specification.Snapshot
     /// </summary>
     public class DifferentialTreeConstructor
     {
+        /// <summary>Singleton. Global default instance.</summary>
+        public static DifferentialTreeConstructor Default { get; } = new DifferentialTreeConstructor();
```
brendankowitz: Maybe `MakeTree` could just be a static method instead? Or even an extension for `IEnumerable<ElementDefinition>`

wmrutten: Hi Brendan, great suggestion! I added a public extension on `StructureDefinition.DifferentialComponent`, as this method is expected to be called on differentials. The extension method simply calls the original (public) MakeTree() method, which expects a list of `ElementDefinitions`.

ewoutkramer:
One day you are going to regret turning an instance into a static....

ewoutkramer:
I am not sure these changes will merge cleanly to STU3, I'll wait to pull it...

Issue #682, opened 2018-09-14
Bug: ResourceResolverExtensions.FindAll<T> should call ResolveByUri


wmrutten:
[reported by @cknaap]
Method `ResourceResolverExtensions.FindAll<T>` first retrieves a list of resource uris, and then calls method `ResolveByCanonicalUri` to resolve individual resources. This is incorrect; instead, method `FindAll` should call method `ResolveByUri` instead.

PR #683, opened 2018-09-14
bugfix: #682 ResourceResolverExtensions.FindAll<T> should call ResolveByUri, not ResolveByCanonicalUrl

wmrutten:


wmrutten:
I tried to merge DSTU2 bugfix into develop-stu3 branch, but I think I need to cherry pick to other branch instead?

Issue #393, opened 2017-07-10
Add a call-back on DirectorySource (and possibly others) when duplicate canonicals are found


ewoutkramer:
This is so the caller can determine what to do with duplicates.  Default behaviour will be:

* Determine whether the filenames are the same (i.e. an xml and json version of the same thing) -> chose the xml version
* Same url in different files -> throw exception

The MultiSource may also benefit from this functionality

ewoutkramer:
I think some of this should be visible in an ArtifactSummary instead of throwing?

wmrutten:
Definitely!
The ArtifactSummaryGenerator does not care about duplicate canonical urls.

Currently, method `DirectorySource.ListSummaries()` also initializes creates a lookup Dictionary to map CanonicalUrls to FileNames. If the content directory contains duplicate canonical urls, then the initialization will fail and throw a `CanonicalUrlConflictException` and the summary list remains empty.

Proposal: do not throw when fetching summaries; allow duplicate canonical urls in list of summaries.
Instead, move error handling to method `ResolveByCanonicalUrl`. By default, if the content directory contains multiple resources that match the specified url, then throw a `CanonicalUrlConflictException`.

Once `DirectorySource` (and other `IResourceResolver` implementations) implement the new error event interface introduced by the validator, then we can allow clients to override the default behavior by hooking the event.

PR #688, opened 2018-09-14
feature: #393 DirectorySource - Improve handling of resolving conflicts

wmrutten:
Don't throw while indexing directory; instead, throw when resolving conflicting url


ewoutkramer:
Hi @wmrutten, I need a STU3 PR for this too, since the differences are more substantial than a trivial merge (caused by the split of ValueSet into ValueSet and CodeSystem in STU3).

PR #710, opened 2018-09-26
Feature/fix default use stu3

ewoutkramer:


ewoutkramer:
Oops, this PR was accidentally based on the develop (DSTU2) branche, should be STU3.

PR #714, opened 2018-09-26
replace remaining uses of element navigator

BradBarnich:
This replaces the remaining uses of `IElementNavigator` with `ITypedElement` and `ISourceNode`

In our own project, it was simpler to only have to use the new types.

Is this something that is useful for the main project?

In case you want to leave the `IElementNavigator` code as Obsolete, the removal code is a separate commit.


ewoutkramer:
Hi, Brad, thanks!  You're an early adopter for sure - I have only recently posted this code. But we had scheduled this work for the next week, so this is a very timely PR ;-)

Did you see the preliminary documentation I wrote about the new classes and interfaces?  http://docs.simplifier.net/fhirnetapi/parsing/intro-to-elementmodel.html

ewoutkramer:
We did a quick survey. Great work!  What we probably will need to change/add is retain the ability to use IElementNavigator - we go into great lengths to not break the build for people upgrading - so some of the changes to the FhirPath engine would need addition of overloads on methods to still accept IElementNavigator.

BradBarnich:
> Did you see the preliminary documentation 

I hadn't. Thanks, that is helpful

src/Hl7.FhirPath/FhirPath/IValueProviderFPExtensions.cs
```diff
@@ -50,61 +45,59 @@ private static CompiledExpression getCompiledExpression(string expression)
...
             return evaluator(input, ctx ?? EvaluationContext.CreateDefault());
         }
 
-        [Obsolete("Replace with the overload taking an EvaluationContext, initialized with the resource parameter")]
-        public static IEnumerable<IElementNavigator> Select(this IElementNavigator input, string expression, IElementNavigator resource)
+        [Obsolete("Use Select(this ITypedElement input) instead")]
+        public static IEnumerable<IElementNavigator> Select(this IElementNavigator input, string expression, EvaluationContext ctx = null)
         {
-            return input.Select(expression, new EvaluationContext(resource));
+            return Select(input.ToTypedElement(), expression, ctx).Select(t => t.ToElementNavigator());
```
BradBarnich: This is a bit odd, in converting the output back to an `IElementNavigator` the node type is lost. So I don't think it is possible to cast to the node type to access things like `ShortPath`
see: https://github.com/ewoutkramer/fhir-net-api/pull/714/files#diff-9a88142c6637196a80e58ef1714b96fbL115

marcovisserFurore:
@BradBarnich : we are still processing your pull request. At first glance it looks very good. Because we want to provide backwards compatibility, we will provide some classes with overload obsolete methods. So I will add some changes in your pull request this week to solve this.

PR #721, opened 2018-09-30
Feature/r4 validation fixes

brianpos:
The case sensitivity of the htmlChecks is incorrect.
Also include an opportunity to pass the fhirpath compiler to the validator (otherwise custom function tables will not be used)

### DIFF
--- a/src/Hl7.Fhir.Specification/Validation/ValidationSettings.cs
+++ b/src/Hl7.Fhir.Specification/Validation/ValidationSettings.cs
@@ -23,6 +23,12 @@ namespace Hl7.Fhir.Validation
         public ITerminologyService TerminologyService { get; set; }
 
         /// <summary>
+        /// An instance of the FhirPath compiler to use when evaluating constraints
+        /// (provide this if you have custom functions included in the symbol table)
+        /// </summary>
+        public Hl7.FhirPath.FhirPathCompiler FhirPathCompiler { get; set; }
+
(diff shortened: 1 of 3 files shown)


src/Hl7.Fhir.Specification/Validation/ValidationSettings.cs
```diff
@@ -22,6 +22,12 @@ public class ValidationSettings
 
         public ITerminologyService TerminologyService { get; set; }
 
+        /// <summary>
```
ewoutkramer: Make sure you update the CopyTo() function so it clones this setting.

src/Hl7.Fhir.Specification/Validation/Validator.cs
```diff
@@ -290,6 +293,9 @@ internal FhirPathCompiler FpCompiler
                     symbolTable.AddFhirExtensions();
 
                     _fpCompiler = new FhirPathCompiler(symbolTable);
+
```
ewoutkramer: No - settings are read-only. In fact, the Settings property has a copy of the original settings passed to it, so it's no use setting it, no one will see it.

ewoutkramer:
Hi @brianpos  - what's the status of this PR?  Is this separate from the stuff you showed us on our call last thursday?

brianpos:
I don't recall if this was an invariant or search parameter, but it was one of those.

PR #727, opened 2018-10-15
For discussion: Replatform FhirClient on HttpClient

BradBarnich:
I don't have much confidence in these changes yet. I can't get the Integration tests to pass with any of the listed DSTU2 endpoints, even before any changes.

The extent of the public API changes are:
1. the removal of the `Last...` properties from `IFhirClient`, I felt these were problematic because they make `FhirClient` not threadsafe. For test code, it was trivial to write a custom httpclienthandler that captures these.
1. the argument to the event handlers is now a `HttpRequest/ResponseMessage`

Most of the code was back ported from #507 that was merged into develop-r4.

Also addresses #478 

Would such API changes be acceptable? Or must an HttpClient implementation remain separate?


src/Hl7.Fhir.Core/Rest/IFhirClient.cs
```diff
@@ -13,12 +13,6 @@ public interface IFhirClient
         bool CompressRequestBody { get; set; }
 #endif
         Uri Endpoint { get; }
-        byte[] LastBody { get; }
-        Resource LastBodyAsResource { get; }
-        string LastBodyAsText { get; }
-        HttpWebRequest LastRequest { get; }
-        HttpWebResponse LastResponse { get; }
-        Bundle.ResponseComponent LastResult { get; }
```
BradBarnich: breaking change

src/Hl7.Fhir.Core/Rest/FhirClient.cs
```diff
@@ -1097,30 +1115,48 @@ private void verifyServerVersion()
...
+        public AfterResponseEventArgs(HttpResponseMessage response)
         {
-            this.RawResponse = webResponse;
-            this.Body = body;
+            Response = response;
         }
 
-        public HttpWebResponse RawResponse { get; internal set; }
-        public byte[] Body { get; internal set; }
+        public HttpResponseMessage Response { get; internal set; }
```
BradBarnich: breaking change

brianpos:
I haven't had a good look over these tests, but one if the things I'd like to see with the client is a pure streaming option that doesn't read the lot into a string first... but that has issues with the existing event handler approach.

BradBarnich:
This doesn't read the body for the event handlers, it could be a stream. But the first step of parsing reads it into a string because that is what the parsers accept, I didn't change that.

It should be possible to read the the content as a stream, create an JsonTextReader/XmlTextReader and pass that to the parser. I can give it a try

brianpos:
There is a streaming version if the parsers too, in fact I believe the string version uses a string stream...

brianpos:
The old code reads it so that the event handler doesn't mess with the stream as it can't rewind...

BradBarnich:
@brianpos I was playing around and I think it should be possible to serialize through a stream directly into the network request. That might reduce allocations a bit, I think one less copy of the xml/json bytes in memory. Did you have a particular use case in mind for a streaming API?

Issue #737, opened 2018-10-22
Replace FhirOperationException.GetObjectData override for compatibility with low-trust environments


wmrutten:
@mbaltus reported that the API packages are incompatible with https://dotnetfiddle.net/ (an online C# playground). DotNetFiddle refuses to load the Core package, reporting a security violation on override method `FhirOperationException.GetObjectData`.

Other libraries have reported a similar issue:
https://github.com/axuno/SmartFormat.NET/issues/104

Explanation:
Method `Exception.GetObjectData` is marked with the [SecurityCritical] attribute (in the dotNet runtime).
Hl7.Fhir.Core.FhirOperationException overrides this method and also specifies the [SecurityCritical]  attribute, as required. However in a low trust environment, all explicit security attribute declarations are ignored. Consequently, the override is no longer valid.

Fortunately, dotNet runtime provides an alternative to overriding GetObjectData that also works in low-trust environments:
https://stackoverflow.com/questions/14124874/how-do-i-implement-exception-getobjectdata-in-net-4-in-a-library-assembly-that

I propose that we update the implementation to the alternative approach.

wmrutten:
Turns out that the System.Exception.SerializeObjectState event is not supported in netcore...

https://github.com/dotnet/coreclr/blob/3b807944d8822b44eb5085d6b95b130b4a91808f
/src/System.Private.CoreLib/src/System/Exception.cs#L421

```
protected event EventHandler<SafeSerializationEventArgs> SerializeObjectState
{
	add { throw new PlatformNotSupportedException(SR.PlatformNotSupported_SecureBinarySerialization); }
	remove { throw new PlatformNotSupportedException(SR.PlatformNotSupported_SecureBinarySerialization); }
}
```

However, it also turns out that in dotnetcore, the Exception.GetObjectData method is NOT decorated with the `SecurityCrititical` attribute.

Suggestion:

* in dotnetframework, custom exceptions can hook the `SerializeObjectState` event
  => compatible with partial trust environment
* in dotnetcore, custom exceptions can override method `Exception.GetObjectData`
  No need to specify SecurityCritical attribute
  => compatible with partial trust environment

ewoutkramer:
Is this not a complete duplicate of #735?

wmrutten:
> 
> 
> Is this not a complete duplicate of #735?

Indeed... Github was experiencing some stability issues yesterday, resulting in duplicate post. I'll close #735.

[...]

Issue #735, opened 2018-10-22
Replace FhirOperationException.GetObjectData override for compatibility with low-trust environments


wmrutten:
@mbaltus reported that the API packages are incompatible with https://dotnetfiddle.net/ (an online C# playground). DotNetFiddle refuses to load the Core package, reporting a security violation on override method `FhirOperationException.GetObjectData`.

Other libraries have reported a similar issue:
https://github.com/axuno/SmartFormat.NET/issues/104

Explanation:
Method `Exception.GetObjectData` is marked with the [SecurityCritical] attribute (in the dotNet runtime).
Hl7.Fhir.Core.FhirOperationException overrides this method and also specifies the [SecurityCritical]  attribute, as required. However in a low trust environment, all explicit security attribute declarations are ignored. Consequently, the override is no longer valid.

Fortunately, dotNet runtime provides an alternative to overriding `GetObjectData` that also works in low-trust environments:
https://stackoverflow.com/questions/14124874/how-do-i-implement-exception-getobjectdata-in-net-4-in-a-library-assembly-that

I propose that we update the implementation to the alternative approach.

wmrutten:
Closed, duplicate of #737

PR #738, opened 2018-10-22
Replace FhirOperationException.GetObjectData override for compatibility with low-trust environments

wmrutten:
See #737

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/Hl7.Fhir.Core.Tests.csproj
+++ b/src/Hl7.Fhir.Core.Tests/Hl7.Fhir.Core.Tests.csproj
@@ -1,9 +1,17 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
+    <!--
+    <TargetFramework>net45</TargetFramework>
+    -->
+    <!-- [WMR 20181022]
+         * VS2017 Test Explorer will only run *first* specified target
+         * dotnet command line tool executes *all* specified targets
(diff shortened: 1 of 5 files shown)


ewoutkramer:
@wmrutten - as I suggested in #737 and #735 - can't we just remove the override of GetObjectData?  I had never intended to do this for all custom Exceptions - and when I see your PR I am not sure I really want all that code for just defining a new exception.

@mharthoorn - is Orleans running dotnetcore or still on net framework?

mharthoorn:
Orleans 2.x runs on .net core 2.y

wmrutten:
@ewoutkramer sure, we can take it out. This would solve the issue at hand (support low-trust environments), but also introduce a breaking change. I've been trying to assess the actual impact.

* Pro: get rid of complexity.
* Con: drop support for exception serialization

Considerations:

* Distributed environments such as Orleans depend on exception serialization  (via `GetObjectData`) for error routing and reporting
* (also inbetween AppDomains, which are obsolete in netCore)
* `FhirOperationException` is used by the `FhirClient` to report `OperationOutcome` and `HttpStatusCode`
* netCore dropped, then re-introduced some runtime support for serialization; this is still a 'hot topic'

wmrutten:
Removed support for serialization from `FhirOperationException` altogether.

ewoutkramer:
@marcovisserFurore  and I have just removed build variables for doing conditional compiles for different platforms - now this PR is about to re-introduce that.  @marcovisserFurore - what do you think?

ewoutkramer:
I have proposed the alternative #752 - which just removes the serialization. I really don't want to introduce multi-platform unit-testing and conditional compiles just for this at this moment - but I do want to help @mbaltus.

ewoutkramer:
Done using #752.

Issue #754, opened 2018-10-31
new Patient().ToTypedElement().ToPoco<Patient>() fails


cknaap:
Probably because TypedElementToSourceNodeAdapter does not add the IResourceTypeSupplier annotation.

PR #765, opened 2018-11-20
fix: POCO - ITypedElement - POCO roundtrip (issue #754)

cknaap:


ewoutkramer:
Had to revert it, since it is incorrect. The implementation should be like so:

`public string ResourceType => Current is Resource ? InstanceType : null;`

since ResourceType should return `null` if it is not a resource....

ewoutkramer:
Will re-issue this as a DSTU2 PR and then merge forwards.

Issue #726, opened 2018-10-10
Add type-less overload for Parse method


brynrhodes:
The now deprecated FhirParser.Parse method had a typeless overload, but the new methods all require a type, either through a generic, or by passing a type. To get a "typeless" parse, you pass `typeof(Resource)` for the type argument, which makes sense, but is not immediately obvious when poking around the methods in an editor. Having an overload with no type argument would have made that more clear.

wmrutten:
Thanks for your suggestion Bryn!

IMHO We should harmonize the same pattern uniformly across all generic methods with a type parameter T that is constrained to Resource.

ewoutkramer:
I was wondering why no one found this one before ;-)  I guess most people use `FhirXmlParser.Parse<T>()`.

PR #766, opened 2018-11-20
Add type-less overload for Parse method #726

gstoian:
Made dataType parameter optional in FhirXmlParser. Added Unit test for this change.

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/Serialization/SerializationTests.cs
+++ b/src/Hl7.Fhir.Core.Tests/Serialization/SerializationTests.cs
@@ -53,6 +53,20 @@ namespace Hl7.Fhir.Tests.Serialization
             Assert.AreEqual(metaXml, xml);
         }
 
+        [TestMethod]
+        public void ParsePatientXmlNullType()
+        {
+            string xmlPacientTest = TestDataHelper.ReadTestData("TestPatient.xml");
+            
+            var poco = new FhirXmlParser().Parse(xmlPacientTest);
+            var xml = new FhirXmlSerializer().SerializeToString(poco);
+            
+            Assert.AreEqual(((Patient)poco).Id, "pat1");
+            Assert.AreEqual(((Patient)poco).Contained.First().Id, "1");
+            Assert.AreEqual(((Patient)poco).Name.First().Family.First(), "Donald");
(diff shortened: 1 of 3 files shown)


src/Hl7.Fhir.Core.Tests/Serialization/SerializationTests.cs
```diff
@@ -53,6 +53,20 @@ public void ParseMetaXml()
             Assert.AreEqual(metaXml, xml);
         }
 
+        [TestMethod]
+        public void ParsePatientXmlNullType()
+        {
+            string xmlPacientTest = TestDataHelper.ReadTestData("TestPatient.xml");
+            
+            var poco = new FhirXmlParser().Parse(xmlPacientTest);
+            var xml = new FhirXmlSerializer().SerializeToString(poco);
```
ewoutkramer: I think we can remove the `var xml....` line, it does not do anything anymore?

src/Hl7.Fhir.Core.Tests/Serialization/SerializationTests.cs
```diff
@@ -631,18 +645,18 @@ public void SummarizeSerializingTest()
         public void DateTimeOffsetAccuracyTest()
```
ewoutkramer: I think these changes were not supposed to be part of this PR?

src/Hl7.Fhir.Core/Serialization/FhirXmlParser.cs
```diff
@@ -30,7 +30,7 @@ public FhirXmlParser(ParserSettings settings=null) : base(settings)
                     DisallowSchemaLocation = Settings.DisallowXsiAttributesOnRoot,
                 };
 
-        public Base Parse(string xml, Type dataType)
+        public Base Parse(string xml, Type dataType = null)
```
ewoutkramer: There is another overload on Parse() which takes an XmlReader, and is also missing the `= null`.  The same is true for these functions on the FhirJsonParser.cs

src/Hl7.Fhir.Core.Tests/Serialization/SerializationTests.cs
gstoian: Yes, they were not. My visual studio doesn't see them at all.

Issue #460, opened 2017-10-27
[serialization] Add support for _elements


ewoutkramer:

ewoutkramer:
The lower-lever serializers now support this, we should enable this in the FhirSerializer public API as well.

PR #778, opened 2018-11-22
Merge 460 to stu3

gstoian:
Added another parameter:elements in serialization method. Added necessary logic to configure the serialization around the new parameter. Unit Test method for the new scenarios.


ewoutkramer:
Fixes #460

PR #779, opened 2018-11-22
devops: defined a variable for running FhirPath tests in CI pipeline

marcovisserFurore:


ewoutkramer:
We agreed I should not pull this PR, but pull #780 instead.

ewoutkramer:
Merged it anyway, we are now explicit about the FhirPath tests in both branches.

Issue #793, opened 2018-11-28
InstanceType for Entry is different in getting vs setting


cknaap:
When getting the instancetype from a Bundle.Entry poco.ToTypedElement, it is 'BackboneElement' (as expected).
But when constructing an standalone Entry from SourceNode, you cannot provide 'BackboneElement' as the root type in ToTypedElement. It works if you provide 'Element'.

```csharp
        [Fact]
        public void TypeForEntry()
        {
            var bundleJson = "{\"resourceType\":\"Bundle\", \"entry\":[{\"fullUrl\":\"http://example.org/Patient/1\"}]}";
            var bundle = FhirJsonNode.Parse(bundleJson);
            var typedBundle = bundle.ToTypedElement(_sdsProvider, "Bundle");
            typedBundle.Select("$this.entry[0]").First().InstanceType.Should().Be("BackboneElement"); //Type of entry is BackboneElement, but you can't set that, see below.

            var entry = SourceNode.Node("entry", SourceNode.Valued("fullUrl", "http://example.org/Patient/1"));

            //What DOES work:
            var typedEntry = entry.ToTypedElement(_sdsProvider, "Element"); //But you can't use BackboneElement here, see below.
            typedEntry.InstanceType.Should().Be("Element");

            //But this leads to a System.ArgumentException: 
            //Type BackboneElement is not a mappable Fhir datatype or resource
            //Parameter name: type
            typedEntry = entry.ToTypedElement(_sdsProvider, "BackboneElement");
            typedEntry.InstanceType.Should().Be("BackboneElement", "Expected to be able to use BackboneElement as type for the entry SourceNode");
        }

```

Note: the unittest is written with XUnit and FluentAssertions.

ewoutkramer:
The bug was buried deep away in the older ClassMapper logic.  Initially, we would not allow you to use any type marked as "abstract" (e.g. Element, Base, DomainResource, BackboneElement, etc).  I had manually made exceptions for this in the generated code, by adding the [FhirType] attribute to it.  Had forgotten to do this for BackboneElement however.

ewoutkramer:
Fixed by #798

PR #798, opened 2018-12-04
fix: Added [FhirType] to BackboneElement to make sure it gets mapped. 

ewoutkramer:
BackboneElement wasn't marked with [FhirType], so the ClassMapper wouldn't map it, therefore the PocoSDProvider wasn't able to retrieve type information for BackboneElement.

Fixes 793


ewoutkramer:
Fixes #793.

Issue #180, opened 2016-03-09
Adding method to retrieve contained resource


mbaltus:
It would be nice to have a method to be able to retrieve a contained resource from its parent by its reference.

ewoutkramer:
For example, like so:

```
var cPat1 = new Patient() { Id = "pat1" };
            var cPat2 = new Patient() { Id = "pat2" };
            var parent = new Observation();

            parent.Contained = new List<Resource> { cPat1, cPat2 };

            var rref = new ResourceReference() { Reference = "#pat2" };

            Assert.IsNotNull(parent.FindContainedResource(rref));
            Assert.IsNotNull(parent.FindContainedResource(rref.Reference));
```

(in ModelTests.cs - FindContainedResource())

PR #800, opened 2018-12-04
feature: adding helper method FindContainedResource on DomainResource

marcovisserFurore:
Solves issue #180

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -161,6 +161,7 @@ src/Hl7.Fhir.Model/EwoutFhirNetApi.snk
 *.snk
 sonar-project.properties
 src/.vs/
+.vs/
 build/runbuild.txt
 
 specification.zip
--- a/src/Hl7.Fhir.Core/Model/DomainResource.cs
+++ b/src/Hl7.Fhir.Core/Model/DomainResource.cs
@@ -56,5 +56,32 @@ namespace Hl7.Fhir.Model
 
             return result;
         }
+
+        /// <summary>
+        /// Finds the contained resource defined by the <paramref name="reference"/>. A reference to a contained resource starts with the
+        /// character #.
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -161,7 +161,8 @@ src/Hl7.Fhir.Model/EwoutFhirNetApi.snk
 *.snk
```
ewoutkramer: Is this coming in from another fix?

src/Hl7.Fhir.Core/Model/DomainResource.cs
```diff
@@ -56,5 +56,32 @@ public override IEnumerable<ValidationResult> Validate(ValidationContext validat
...
+        /// <param name="reference">the reference to contained resource</param>
+        /// <returns>The resource referenced by <paramref name="reference"/>, null otherwise.</returns>
+        public Resource FindContainedResource(ResourceReference reference)
+        {
+            if (!reference.IsContainedReference) return null;
+
+            if (reference.Reference == "#") return this;
+
+            // search the contained resource by removing '#' 
+            return Contained.FirstOrDefault(c => c.Id == reference.Reference.Remove(0, 1));
```
ewoutkramer: This fails if reference.Reference is null.

.gitignore
marcovisserFurore: The newest version of Visual Studio places a map called ``.vs/`` in the root folder. This has to be ignored for our project

src/Hl7.Fhir.Core/Model/DomainResource.cs
marcovisserFurore: Oops, you are right of course. I created a new commit for that

ewoutkramer:
Fixes #180.

Issue #339, opened 2017-05-03
Wrong target list for 'patient' search parameters


michelemottini:
In the STU3 library ModelInfo.SearchParameters.Target contains `Patient` and `Group` for `patient` search parameters. It should be just `Patient`

brianpos:
This search parameter is shared between the 2 resources (and others) and what is output by the build.

michelemottini:
@brianpos : sorry, I don't understand what you mean. 

A lot of resources have a `patient` search parameters - bound to properties that can reference also other resources, but that search parameter itself is specific for patient references - that means you can do `&patient=<id>`  instead of having to specify the referenced resource type  `&subject=Patient/<id>` or `&subject:Patient=<id>`

ewoutkramer:
Michele, you are right, a "patient" search parameter could only have Targets "Patient".  Since this part of the code is generated from the STU3 source files, the error must be/have been there.  I am saying "have been", since we have regenerated the source now based on the technical correction 3.0.1 release and I cannot locate any errors anymore based on some randomly picked ones (too many to check them all)

Do you remember which "subject" search parameters were broken before?  If there were many, I am pretty confident is has been fixed....

michelemottini:
It seemed all of them - `AllergyIntolerance.patient` for example

ewoutkramer:
I found the cause. No, it's not all of them.  Some resources have their own defined parameter "patient", here the target is just "Patient".  However, there is a searchparameter in the spec called "clinical-patient" that is defined for a number of resources, and which has both Patient and Group. This is the cause.  It does not seem to make sense to me.  I'll continue this discussion in Zulip....

michelemottini:
Thanks Ewout

ewoutkramer:
This has been turned into [issue 13601 on gforge](http://gforge.hl7.org/gf/project/fhir/tracker/?action=TrackerItemEdit&tracker_item_id=13601).

When this issue is fixed, the bug will automatically disappear after re-generation of the object model.

ewoutkramer:
This issue has not been fixed, since there has not yet been a technical correction for STU3. We now have a few of these issues where we need the publication to change - maybe we should manually correct these instances and not wait for any technical updates from HL7.

michelemottini:
I corrected the target for those 'patient' reference search parameters in the code generation script on our multi-version branch.

ewoutkramer:
[...]

PR #801, opened 2018-12-06
fix: target of common SearchParameter clinical-patient is Patient (and not Group anymore)

marcovisserFurore:
Solves issue #339.

Target of SearchParameter clinical-patient is not a list of Patient and Group anymore, but only of Patient. This is an error in de source of the spec. Also the description of SP's for multiple resources has been updates. (Inspired by @michelemottini)


ewoutkramer:
Fixes #339.

PR #802, opened 2018-12-07
Feature/develop r4 brian

brianpos:
Updating to the latest R4 version (almost final)
and many updates supporting this release
(heaps to review - likely not approve as is)


src/Hl7.Fhir.Specification/Validation/FpConstraintValidationExtensions.cs
```diff
@@ -36,6 +36,9 @@ public static OperationOutcome ValidateFp(this Validator v, ElementDefinition de
 
             foreach (var constraintElement in definition.Constraint)
             {
+                // Skip any best practice constraints until that setting is available
```
brianpos: https://github.com/ewoutkramer/fhir-net-api/issues/796
This is relevant to the above issue

brianpos:
Code generation of enum values that start with a number should just prefix the value with an underscore rather than change the number to a word.

Issue #561, opened 2018-04-03
FhirPath cache without lock()


cknaap:
The lock() in IValueProviderFPExtensions._cache takes a lot of time, as my profiler revealed.
Consider to replace it with an implementation based on ConcurrentDictionary, available in .NetStandard 1.1 and up, and .NET 4.0 and up.

cknaap:
The same problem occurs in the CachedResolver.

wmrutten:
API locking strategy is quite crude yet, there is definitely room for improvement.
Currently, CachedResolver.Cache<T>.Get immediately takes a lock. Possibly we can rewrite this method to an optimistic locking strategy?

PR #803, opened 2018-12-10
feature: new and faster cache for compiled FhirPaths

marcovisserFurore:
Solves issue #561.

Time measurements:

_Before this improvement:_
- hitting the cache 20,000 times in 20 parallel tasks took an average time of 7,850ms (almost 8 sec).

This improvement:
- hitting the cache 20,000 times in 20 parallel tasks took an average time of 5,691ms.
- when we empty the cache by 10% of its size we get even a better performance. Then we have 4,725ms of time. This method is now implemented.


ewoutkramer:
Fixes #561

Issue #756, opened 2018-11-01
Validation issue with text.div?


alexzautke:
Currently, Vonk is throwing an internal server error when trying to serialise the attached Patient resource. The problem is caused by an invalid content in the text.div element. However, when calling $validate with this resource, HTTP 200 is returned.

Could you please check why the validator is accepting this resource?

[Patient.txt](https://github.com/ewoutkramer/fhir-net-api/files/2537967/Patient.txt)

brianpos:
The text element is not a valid xhtml value, it should be like the below and it should load ok
  "text": {
    "status": "generated",
    "div": "<div xmlns=\"http://www.w3.org/1999/xhtml\">testje</div>"
  },

brianpos:
Vonk may even return an OperationOutcome that has this location as where the issue is, and what the problem was.

alexzautke:
Yes that's completely true. But, shouldn't the validator also flag the content as invalid? We were running into the issue because validation is disabled by default on the production environment of Vonk, so it was written as is into the database. But even with validation enabled, it would have been accepted.

brianpos:
Then this is likely a vonknissue rather than fhir-net-api

cknaap:
We could check this in Vonk, but I would have expected to validator to return an error on a Narrative that is not valid xhtml.
This unittests shows it outside the context of Vonk. It fails currently.

```csharp
using Hl7.Fhir.Model;
using Hl7.Fhir.Serialization;
using Hl7.Fhir.Specification.Source;
using Hl7.Fhir.Validation;
using Microsoft.VisualStudio.TestTools.UnitTesting;

namespace Hl7.Fhir.Specification.Tests.Validation
{
    [TestClass]
    public class NarrativeValidationTests
    {

        [TestMethod]
        public void NarrativeMustBeValidXml()
        {
            var json = "{\"resourceType\": \"Patient\", \"text\": \"text without div\"}";
            var patient = new FhirJsonParser().Parse<Patient>(json);
            var validationSource = new CachedResolver(ZipSource.CreateValidationSource());
            var validator = new Validator(new ValidationSettings() { ResourceResolver = validationSource, EnableXsdValidation = true });
            var validationResult = validator.Validate(patient);
            Assert.IsFalse(validationResult.Success, "Expect failure on narrative without <div>");
        }
    }
}
```

ewoutkramer:
[...]

PR #810, opened 2018-12-13
Feature/756 json parser refactoring

gstoian:
https://github.com/ewoutkramer/fhir-net-api/issues/756

Changes were needed also in Unit tests and other related files.

### DIFF
--- a/src/Hl7.Fhir.Core/Serialization/RepeatingElementReader.cs
+++ b/src/Hl7.Fhir.Core/Serialization/RepeatingElementReader.cs
@@ -31,7 +31,7 @@ namespace Hl7.Fhir.Serialization
 
 #pragma warning restore 612, 618
 
-        public IList Deserialize(PropertyMapping prop, string memberName, string typeName, IList existing = null)
+        public IList Deserialize(PropertyMapping prop, string memberName, IList existing = null)
         {
             if (prop == null) throw Error.ArgumentNull(nameof(prop));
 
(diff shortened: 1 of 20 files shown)


src/Hl7.Fhir.Core/Serialization/RepeatingElementReader.cs
```diff
@@ -31,7 +31,7 @@ internal RepeatingElementReader(ISourceNode reader, ParserSettings settings)
 
 #pragma warning restore 612, 618
 
-        public IList Deserialize(PropertyMapping prop, string memberName, IList existing = null)
+        public IList Deserialize(PropertyMapping prop, string memberName, string typeName, IList existing = null)
```
ewoutkramer: This should not need a typeName, since _current.InstanceType already has a type....

src/Hl7.Fhir.Core/Serialization/ResourceReader.cs
```diff
@@ -41,11 +41,11 @@ public Resource Deserialize(Resource existing=null)
         {
             // If there's no a priori knowledge of the type of Resource we will encounter,
             // we'll have to determine from the data itself. 
-            var resourceTypeName = _reader.GetResourceTypeIndicator();
-            var mapping = _inspector.FindClassMappingForResource(resourceTypeName);
+
+            var mapping = _inspector.FindClassMappingForResource(_reader.InstanceType);
```
ewoutkramer: Just like in ComplexTypeReader, there's an assumption here that InstanceType != null. I am not so sure...

ewoutkramer:
Fixes #756

PR #816, opened 2019-01-03
.NET 4.0 compatibility

lstratman:
It's been a while since I last checked on the .NET 4.0 compatibility status, so I went ahead and branched `develop-r4` and added the necessary .NET 4.0 package references and a minor code change to get everything to compile.


ewoutkramer:
Hi Luke,

Thanks for the PR, we've discussed the pro's and cons of maintaining .NET 4.0 during our standup calls and the last .NET users meeting in San Antonio last week and the consensus seems to be that we don't want to support this in our main branch.  There is a non-zero cost for doing this (mainly having to make sure it keeps working when adding new functionality and dependencies), although this is admittedly a pretty small PR.

I'll keep this PR open for now, so we can see whether there are more takers - or you can join our next standup (https://github.com/ewoutkramer/fhir-net-api/wiki/20190204-Community-Meeting-Agenda) to argue in favor ;-)

lstratman:
Thanks for the info, Ewout, I’ll definitely join the next standup to plead my case :).


________________________________

ewoutkramer:
Had a standup meeting last monday - pitched the NET4 support - but could not find other fans for it.

marcovisserFurore:
This PR has been replaced by PR #1044. Thanks for your contribution @lstratman !

PR #818, opened 2019-01-03
Fixes Json serializer returning instead of continuing when encountering an empty object

### LINKED ISSUE #817: Setting an empty Meta object breaks Json Serialization
When setting the `Meta` property to an empty instance, the JSON serializer doesn't seem to serialize all properties.

In the test below, the value of `reserialized` is `{"resourceType":"Patient","id":"pat1"}` instead of containing all properties.

```
[TestMethod]
public void GivenAPatient_WhenModifingAndSerializing_ThenTheResultHasNoDataLoss()
{
    var fhirJsonParser = new FhirJsonParser();

    string json = TestDataHelper.ReadTestData("TestPatient.json");
    var poco = fhirJsonParser.Parse<Patient>(json);

    Assert.AreEqual(1, poco.Name.Count);

    // Setting an empty Meta instance causes JSON serialization to break
    poco.Meta = new Meta();

    var reserialized = poco.ToJson(); // Broken
    var xml = poco.ToXml(); // OK

    var newPoco = fhirJsonParser.Parse<Patient>(reserialized);

    // There are 0 names 
    Assert.AreEqual(1, newPoco.Name.Count);
}
```

brendankowitz:
Fixes  #817

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/Serialization/SerializationTests.cs
+++ b/src/Hl7.Fhir.Core.Tests/Serialization/SerializationTests.cs
@@ -932,5 +932,24 @@ namespace Hl7.Fhir.Tests.Serialization
             json2 = new FhirJsonSerializer().SerializeToString(patient);
             Assert.AreEqual(json, json2, "5");
         }
+
+        [TestMethod]
+        public void SerializerHandlesEmptyChildObjects()
+        {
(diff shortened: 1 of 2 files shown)


ewoutkramer:
Good catch ;-)

PR #829, opened 2019-01-10
Feature/develop r4 brian fixes

brianpos:
Additional Unit tests & bug fixes
* FhirPath PocoElement node value extraction throws cast exceptions when the item does not have a value (such as when only having an extension on the property)
* Validation of References to contained resources where there are multiple potential types, and the contained resource is not the first one targeted


ewoutkramer:
@marcovisserFurore will apply the changes to PocoElementNode.cs in DSTU2/STU3/R4 - he also created an alternative fix for the targetProfile issue, which he will apply to R4.  So, I will now close this PR, it will come back to life as separate PRs from @marcovisserFurore.  Thanks!

Issue #827, opened 2019-01-09
[Snapshot] Child constraints are not included in snapshot when constraining a .value element


ewoutkramer:
To reproduce:  take a look at buildOrganizationWithRegexConstraintOnType() in TestProfileArtifactSource.cs. This creates an Organization with a constraint on Organization.name.value (hence on string), adding a regex constraint (in the form of an extension) The relevant part of the differential looks like this:

```
 <element>
            <path value="Organization.name.value"/>
            <type>
                <extension url="http://hl7.org/fhir/StructureDefinition/regex">
                    <valueString value="[A-Z].*"/>
                </extension>
                <code value="string"/>
            </type>
  </element>
```

In the base (string.value), the `<type>` node looks like this:

```
<type>
     <extension url="http://hl7.org/fhir/StructureDefinition/regex">
              <valueString value="[^\s]+(\s[^\s]+)*" />
     </extension>
      <code>
                <extension url="http://hl7.org/fhir/StructureDefinition/structuredefinition-json-type">
                  <valueString value="string" />
                </extension>
                <extension url="http://hl7.org/fhir/StructureDefinition/structuredefinition-xml-type">
                  <valueString value="xsd:token" />
                </extension>
                <extension url="http://hl7.org/fhir/StructureDefinition/structuredefinition-rdf-type">
                  <valueString value="xsd:token" />
                </extension>
      </code>
</type>
```

The resulting snapshot after processing the differential shown above is:

```
            <type>
                <extension url="http://hl7.org/fhir/StructureDefinition/regex">
                    <valueString value="[A-Z].*"/>
                </extension>
                <code value="string"/>
            </type>
```

This results in loss of an important set of constraints (all encoded as extensions) in the snapshot.

The reason we have not found this before is because not many people actually do constraints on .value elements (the only usecase being adding an regex constraint).  And since the new regex constraint (which is normally more precise than the base one) IS present, all seems to work fine.

However, in R4, the validator is dependent on the additional extension on the `<type>/<code>` node and will fail to recognize constrained .value elements as primitives.  There is currently a work-around in place for this bug, which should be removed when this gets fixed (look for  `public static bool IsPrimitiveConstraint(this ElementDefinition ed) ` to find it).

wmrutten:
Confirmed, I can reproduce the behavior. 

This line is too blame:
[...]

PR #841, opened 2019-01-22
fix: #827 Snapshot should merge core type extensions from primitive value elements

wmrutten:
Note: this is a quick 'n dirty fix for the reported issue concerning `ElementDefinition.type`, however we actually need a generic solution for properly merging *all* the attributes of `ElementDefinition` (instead over replacing/overwriting) according to various rules, as outlined by Chris Grenz.


wmrutten:
Note: we discussed moving the custom methods for merging element type constraints to partial class `ElementDefinition.TypeRefComponent`. However I decided not to, as this would require adding custom methods to a number of other PoCo classes that are only relevant to Snapshot Generation. Seemed better to keep this logic together. We can refactor this in the future if/when necessary.

ewoutkramer:
Agree, can't remember discussing this, but still agree ;-)

PR #846, opened 2019-01-23
Enable cross-platform development

kennethmyhra:
This PR is probably in the nice to have section, it enables the tests to run on a Linux-based OS, as well as Windows. It should also work on Mac OS X, but I haven't tested it on OS X yet.

There are a few things here that I am not quite happy with.

1. In src/fhir-net-api.targets I have commented out the Otherwise section for VSTestLogger, since I was not able to get a proper listing of the tests in vscode. Not sure if this has any other implications like the build process or anything?
2. ~~In .gitattributes line endings are now `lf` for *.xml and *.json independently of platform, tried back and forth between `crlf` and `lf`, but `lf` was what ended up working on both Linux and Windows~~
3. Added a Thread.Sleep in ArtifactSourceTests::ZipCacherShouldCache at line 77.  Not sure what is going on, and it still fails sometimes on Linux, but if I do not wait before the statement File.SetLastWriteTimeUtc fa.IsActual always returns true and the test fails.

I added these settings in .vscode/settings.json to get the test runner in vscode run properly + only run the unit tests:
```
{
    "dotnet-test-explorer.testProjectPath": "**/*Tests.csproj",
    "dotnet-test-explorer.testArguments": "--filter TestCategory!=IntegrationTest"
}
```

Commit summary:
* Basically changed all paths that had a hard-coded separator to use
  Path.Combine instead
* Removed the Otherwise section in fhir-net.api.targets since
  /logger:Console does not give the appropriate output to i.e. list the
  tests in the .NET Test Explorer in vscode
* Added a Thread.Sleep in ArtifactSourceTests::ZipCacherShouldCache at
  line 77. See comment for details, not sure what is going on, but it
  seems that it hasn't solved the actual problem since it sometimes
  works, while other times will fail.
* ~~eol always lf for .xml and .json. If not the navigator compare
  examples where 'expected' input is xml and  'actual' input is json will
  fail. This is because strings in json cannot contain line breaks
  whereas in xml they can.~~

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/FhirPathExtensionsTest.cs
+++ b/src/Hl7.Fhir.Core.Tests/FhirPathExtensionsTest.cs
@@ -29,7 +29,7 @@ namespace Hl7.Fhir.Tests.Introspection
         public void SetupSource()
         {
             ElementNavFhirExtensions.PrepareFhirSymbolTableFunctions();
-            var bundleXml = File.ReadAllText("TestData\\bundle-contained-references.xml");
+            var bundleXml = File.ReadAllText(Path.Combine("TestData", "bundle-contained-references.xml"));
 
             _parsed = (new FhirXmlParser()).Parse<Bundle>(bundleXml);
             _bundleElement = new ScopedNode(_parsed.ToTypedElement());
--- a/src/Hl7.Fhir.Core.Tests/PocoNavigatorTests.cs
+++ b/src/Hl7.Fhir.Core.Tests/PocoNavigatorTests.cs
@@ -179,7 +179,7 @@ namespace Hl7.Fhir
         [TestMethod]
         public void IncorrectPathInTwoSuccessiveRepeatingMembers()
         {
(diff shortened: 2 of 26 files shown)


kennethmyhra:
Tests related to the crlf and lf issue seems to still be failing in the build process, Thought I had fixed these, but apparently not.

wmrutten:
The failing unit test for SetLastWriteTimeUtc on Linux might be related to this:
https://github.com/dotnet/corefx/issues/26024

kennethmyhra:
Thanks, Michel. I'll look into that

kennethmyhra:
Reverted the eol settings for xml/json I had previously set in .gitattributes.

Instead having a special case for json strings when on a unix-based platform:
```
// If on a Unix platform replace \\r\\n in json strings to \\n.
if(Environment.NewLine == "\n")
    tpJson = tpJson.Replace(@"\r\n", @"\n");
```

ewoutkramer:
Good improvements. We should remain vigilant that we don't add vanilla DOS paths anymore but use Path.Combine instead....

PR #850, opened 2019-01-30
Develop stu3

mthakershi:


ewoutkramer:
I am pretty sure this PR is created by accident.  It mentions no fixes, no issue numbers and just merges STU3 into DSTU2.  Note that we do make sure all functionality of DSTU2 is carried over to 3 and similarly STU3 changes are forwarded to R4, so we do not need any further merges across branches like this one.

PR #861, opened 2019-02-11
fix: change the technical license structure

marcovisserFurore:
Change the technical license structure in the solution to meet the nuget requirements. Also update the year in the copyright statement.


wmrutten:
Note: copyright year should always refer to date of *first* release and not be updated incrementally; i.e. interpret this as "copyright as of/since ...".

PR #751, opened 2018-10-29
#748, #749, #750 - Expose ParserSettings & ExceptionNotificationHandler from ResourceResolvers

wmrutten:
Add support for custom `ParserSettings` & `ExceptionNotificationHandler` to `DirectorySource`, `ZipSource` and `WebResolver`

Issue #750, opened 2018-10-29
WebResolver should support ParserSettings


wmrutten:
The `WebResolver` should allow clients to configure custom `ParserSettings` for de-serialization (cf. #748).

Issue #748, opened 2018-10-25
DirectorySource(Settings) should also expose ParserSettings


wmrutten:
The `DirectorySource` should allow clients to configure `ParserSettings` for de-serialization.

Especially useful for clients such as Forge that want to allow invalid input, so the user can correct errors.

PR #869, opened 2019-02-13
Feature/748 750 parser and web resolver settings

marcovisserFurore:
Solves #748 and #750. 

This PR is part of PR #751 thanks to @wmrutten

### DIFF
--- a/src/Hl7.Fhir.Serialization/JsonNavigatorStream.cs
+++ b/src/Hl7.Fhir.Serialization/JsonNavigatorStream.cs
@@ -72,7 +72,7 @@ namespace Hl7.Fhir.Serialization
             _stream = stream ?? throw Error.ArgumentNull(nameof(stream));
             _reader = SerializationUtil.JsonReaderFromStream(stream);
             _disposeStream = disposeStream;
-            ParserSettings = settings ?? FhirJsonParsingSettings.CreateDefault();
+            ParserSettings = settings?.Clone() ?? FhirJsonParsingSettings.CreateDefault();
 
             // [WMR 20181026] TODO: Don't reset stream by default!
             // cf. XmlNavigatorStream
--- a/src/Hl7.Fhir.Serialization/JsonNavigatorStream.cs
+++ b/src/Hl7.Fhir.Serialization/JsonNavigatorStream.cs
@@ -6,15 +6,13 @@
  * available at https://github.com/ewoutkramer/fhir-net-api/blob/master/LICENSE
  */
 
-using System.Collections.Generic;
-using System.IO;
-using System.Linq;
+using Hl7.Fhir.ElementModel;
 using Hl7.Fhir.Utility;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using System;
-using Hl7.Fhir.ElementModel;
 using System.Collections;
+using System.IO;
 
 #if !NETSTANDARD1_1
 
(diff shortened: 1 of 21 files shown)


src/Hl7.Fhir.Specification/Specification/Source/DirectorySource.cs
```diff
@@ -884,32 +891,47 @@ private List<ArtifactSummary> harvestSummaries(List<string> paths)
...
+            var settings = _settings;
+            var factory = _navigatorFactory;
+            settings.XmlParserSettings.CopyTo(factory.XmlParsingSettings);
+            settings.JsonParserSettings.CopyTo(factory.JsonParsingSettings);
+
+            // Also use the current PoCo parser settings
+            var pocoSettings = new PocoBuilderSettings();
+            if (settings.ParserSettings != null)
             {
+                pocoSettings.AllowUnrecognizedEnums = settings.ParserSettings.AllowUnrecognizedEnums;
```
ewoutkramer: Isn't this fragile?  What if we add new parser settings, these will then not get copied over?

src/Hl7.Fhir.Serialization/JsonNavigatorStream.cs
```diff
@@ -55,14 +61,25 @@ public static JsonNavigatorStream FromPath(string path)
...
+        /// <summary>Create a new <see cref="JsonNavigatorStream"/> instance for the specified json resource stream.</summary>
+        /// <param name="stream">A stream that returns a serialized json resource.</param>
+        /// <param name="disposeStream">Determines if the <see cref="Dispose()"/> method should also dispose the specified <paramref name="stream"/> instance.</param>
+        /// <param name="settings">Parser settings that control de-serialization behavior.</param>
+        public JsonNavigatorStream(Stream stream, bool disposeStream, FhirJsonParsingSettings settings)
         {
             _stream = stream ?? throw Error.ArgumentNull(nameof(stream));
             _reader = SerializationUtil.JsonReaderFromStream(stream);
             _disposeStream = disposeStream;
+            ParserSettings = settings ?? FhirJsonParsingSettings.CreateDefault();
```
ewoutkramer: Shouldn't this be `ParserSettings = settings.Clone()`?

src/Hl7.Fhir.Specification/Specification/Source/DirectorySource.cs
wmrutten: Marco suggests to implement members on settings classes themselves to copy internal state to another settings class. For example:

```
ParserSettings.CopyTo(PocoBuilderSettings settings) { /* ... */ }
PocoBuilderSettings.CopyTo(ParserSettings settings) { /* ... */ }
```

This is much better maintainable, esp. when adding new properties.

src/Hl7.Fhir.Serialization/JsonNavigatorStream.cs
wmrutten: Agreed, the API should always clone incoming settings.
And we should ensure that all the settings classes have similar behavior.

marcovisserFurore: I've changed this behaviour to 
`ParserSettings = settings?.Clone() ?? FhirJsonParsingSettings.CreateDefault();`

src/Hl7.Fhir.Specification/Specification/Source/DirectorySource.cs
marcovisserFurore: I've created a new method for this:
`public void ParserSettings.CopyTo(PocoBuilderSettings settings)`

wmrutten:
Great job. The new `CopyTo` methods on the settings classes are definitely an improvement. Way easier to maintain when we need to add new settings property.

PR #886, opened 2019-02-27
Allow now also Exponents in de primitive type decimal

marcovisserFurore:
Allow now also Exponents in de primitive type decimal, which actually only allowed from R4

So 1.2e2 is now also correct decimal.

### DIFF
--- a/src/Hl7.Fhir.Support/Serialization/PrimitiveTypeConverter.cs
+++ b/src/Hl7.Fhir.Support/Serialization/PrimitiveTypeConverter.cs
@@ -13,11 +13,14 @@ using Hl7.Fhir.Model.Primitives;
 using Hl7.Fhir.Support.Model;
 using System.Numerics;
 using System.Globalization;
+using System.Linq;
 
 namespace Hl7.Fhir.Serialization
 {
     public static class PrimitiveTypeConverter
     {
+        static readonly string[] FORBIDDEN_DECIMAL_PREFIXES = new[] { "+", ".", "00" };
+
         public static object FromSerializedValue(string value, string primitiveType)
         {
             var type = Primitives.GetNativeRepresentation(primitiveType);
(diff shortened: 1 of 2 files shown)


src/Hl7.Fhir.Support/Serialization/PrimitiveTypeConverter.cs
```diff
@@ -120,7 +121,14 @@ private static object convertXmlStringToPrimitive(Type to, string value)
             if (typeof(DateTime) == to)
                 return ConvertToDatetimeOffset(value).UtcDateTime;  // Obsolete: use DateTimeOffset instead!!
             if (typeof(Decimal) == to)
-                return XmlConvert.ToDecimal(value);
+            {
+                if (Array.Exists(new[] { "+", ".", "00" }, c => value.StartsWith(c)) || value.EndsWith("."))
```
ewoutkramer: This results in constructing this array for every single parse.  I think we could define:

```
    static readonly string[] FORBIDDEN_PREFIXES = new[] {...};
```

to avoid that.

ewoutkramer: And I notice that if the array has 'N' values, `value.Endswith(.)` is also executed 'N' times.

ewoutkramer: Alternatief: `if( Array.Exists(new[] { "+", ".", "00" }, c => value.StartsWith(c))) ||  value.EndsWith("."))`

ewoutkramer: And you could replace the Exists with the expression `FORBIDDEN_PREFIXES.Any(c => value.StartsWith(c))`

marcovisserFurore: > And I notice that if the array has 'N' values, `value.Endswith(.)` is also executed 'N' times.

That is not really true. The if-statement is constructed by two operands: 
- Array.Exists() 
- value.EndsWith() 
The operator is an OR between these operands.

So first the Array.Exists() will be evaluated. If that results in false, then the second operand will be evaluated.

marcovisserFurore: > And you could replace the Exists with the expression `FORBIDDEN_PREFIXES.Any(c => value.StartsWith(c))`

I will change this

Issue #892, opened 2019-03-05
ArtifactSummaryGenerator should return error summaries with file information


wmrutten:
The `ArtifactSummaryGenerator` class captures runtime exceptions and returns error summaries, allowing clients to list invalid files. However the returned error summaries do not initialize the `ArtifactSummary.Origin` property that actually identifies the offending file.

Proposal: the `ArtifactSummaryGenerator` should initialize error summaries with basic file properties (Origin, FileSize, LastModified).

PR #893, opened 2019-03-05
fix: #892 - ArtifactSummaryGenerator should return error summaries with file information

wmrutten:


ewoutkramer:
Hi @wmrutten - because of other PRs this PR now has a (simple) merge conflict - the new unit-test needs to be put at the end of the conflicting file.  Can you do that for me and send the PR again?

Issue #894, opened 2019-03-05
DirectorySource should expand specified contentDirectory to full path


wmrutten:
The `FilePatternFilter` implementation logic assumes that the base directory is a full path specification, otherwise the specified filters will not match any files.

Proposal: the `DirectorySource` constructor should always expand the specified `contentDirectory` folder path to a full (absolute) path specification.

ewoutkramer:
I guess that is true ;-)

ewoutkramer:
Fixed by #895.

PR #895, opened 2019-03-05
fix: #894 DirectorySource should expand specified contentDirectory to full path

wmrutten:


marcovisserFurore:
This PR solves #894

PR #897, opened 2019-03-05
better async for operations

scottfavre:
Uses first-class async primitives to handle timeouts in GetResponseAsync instead of a blocking .Wait call

This solves the case where `FhirClient` is used for large number of parallel calls (such as `TranslateConceptAsync`) and the `.Wait` call times out because active task slots are taken up by the waiting tasks.

The extra error handling for timeouts was not necessary, as the task returned by `Task.Factory.FromAsync` failed with an appropriate exception when it timed out.

Note that .NET Standard 1.1 still uses the old code because it doesn't have a timeout on the web request
removes several unused helper methods

### DIFF
--- a/src/Hl7.Fhir.Core/Rest/WebRequestExtensions.cs
+++ b/src/Hl7.Fhir.Core/Rest/WebRequestExtensions.cs
@@ -95,6 +95,7 @@ namespace Hl7.Fhir.Rest
 
         internal static Task<WebResponse> GetResponseAsync(this WebRequest request, TimeSpan timeout)
         {
+#if NETSTANDARD1_1
             return Task.Factory.StartNew<WebResponse>(() =>
             {
                 var t = Task.Factory.FromAsync<WebResponse>(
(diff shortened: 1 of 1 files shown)


ewoutkramer:
Thanks @scottfavre! I initially created this async code about 5 years ago under the .NET4 framework, I guess times have changed. I do remember having to add the `WaitOne()` calls etc. to have stable behaviour.

I must certainly hope this has been fixed over the past years, and we can now use your improved code. @lstratman - would this be portable to NET4?

lstratman:
With some minor modifications, yes, I believe so.  I'm going to add it to my list of branches/PRs to look at.

lstratman:
Can confirm that this will work just fine on .NET 4.0.  No modifications needed.

marcovisserFurore:
Thanks @lstratman for your work

marcovisserFurore:
@scottfavre: we will merge your PR in the coming period.

scottfavre:
Great! Thanks.

Issue #833, opened 2019-01-14
FhirClient.Read<Resource>(string) - null reference when server not responding


mharthoorn:
When requesting a resource with the FhirClient.Read<Resource>(string) command, a null reference error is thrown when the server does not respond. 

System.NullReferenceException
  Message=Object reference not set to an instance of an object.
  Source=Hl7.Fhir.STU3.Core
  StackTrace:
   at Hl7.Fhir.Rest.TaskExtensions.WaitResult[T](Task`1 task)

(version 1.0.0-alpha3)

marcovisserFurore:
In the latest version of the API we get now a different exception when we perform the following Read:
```
    FhirClient client = new FhirClient("https://aidbox.io");
    var patient = client.Read<Patient>("Patient/example");
```
The exception is:
```
System.ArgumentException: Cannot determine the type of the root element at 'Resource', please supply a type argument.
Parameter name: type
```
So still not a nice response and we could return a more meaningful message.

PR #906, opened 2019-03-13
833 FhirClient.Read null reference exception

gstoian:
The thrown exception is now FhirOperationException and the message is clear. Also fixed the case when user is trying to read a certain type from a different type path. Created a Unit test for this.

### DIFF
--- a/src/Hl7.Fhir.Core/Rest/FhirClient.cs
+++ b/src/Hl7.Fhir.Core/Rest/FhirClient.cs
@@ -1058,9 +1058,11 @@ namespace Hl7.Fhir.Rest
                 if (result is OperationOutcome)
                     return null;
 
-                var message = String.Format("Operation {0} on {1} expected a body of type {2} but a {3} was returned", response.Request.Method,
-                    response.Request.Url, typeof(TResource).Name, result.GetType().Name);
-                throw new FhirOperationException(message, _requester.LastResponse.StatusCode);
+                var message = String.Format("Operation {0} on {1} expected a body of type {2} but a {3} was returned", request.Request.Method,
+                    request.Request.Url, typeof(TResource).Name, result.GetType().Name);
+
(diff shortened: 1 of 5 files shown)


marcovisserFurore:
This PR solves #833

src/Hl7.Fhir.Core/Rest/FhirClient.cs
```diff
@@ -1058,9 +1058,11 @@ protected virtual void AfterResponse(HttpWebResponse webResponse, byte[] body)
                 if (result is OperationOutcome)
                     return null;
 
-                var message = String.Format("Operation {0} on {1} expected a body of type {2} but a {3} was returned", response.Request.Method,
-                    response.Request.Url, typeof(TResource).Name, result.GetType().Name);
-                throw new FhirOperationException(message, _requester.LastResponse.StatusCode);
+                var message = String.Format("Operation {0} on {1} expected a body of type {2} but a {3} was returned", request.Request.Method,
+                    request.Request.Url, typeof(TResource).Name, result.GetType().Name);
+
+                Enum.TryParse(response.Response.Status, out HttpStatusCode code);
```
ewoutkramer: Why do you switch to use response.Response.Status instead of _requester.LastResponse.StatusCode?

src/Hl7.Fhir.Core/Rest/HttpToEntryExtensions.cs
```diff
@@ -151,6 +151,10 @@ private static Resource parseResource(string bodyText, string contentType, Parse
 
                 return null;
             }
+            catch (ArgumentException ae) when (!throwOnArgumentException)
```
ewoutkramer: We should not do this - instead, the error raised (currently ArgumentException) should be turned into a FormatException so it is caught by the original catch handler.

PR #915, opened 2019-03-25
Do not serialize empty values

michelemottini:
...because they are invalid when parsing


ewoutkramer:
Hi Michele, I have added this unit-test to the project, but it runs OK against the current version (1.2), so it seems this has recently been fixed (indeed, we have made changes since 1.0). I'll keep the test in.

PR #935, opened 2019-04-09
Add memory-based implementation of ITypedElement

### LINKED ISSUE #790: Please add ITypedElement.NewChild(ISourceNode source)
This allows users to add new children to an ITypedElement, reusing the type information that is already known to the parent. 

Analysis: probably by splitting up the for loop inside TypedElementNode.enumerateElements()

ewoutkramer:
Adds a class ElementNode (much like SourceNode) that can be used to create a "DOM-like" model that implements ITypedElement.

Fixes #790.


src/Hl7.Fhir.ElementModel.Tests/SourceNodeTests.cs
```diff
@@ -0,0 +1,180 @@
+﻿/* 
+ * Copyright (c) 2015, Firely (info@fire.ly) and contributors
```
marcovisserFurore: 2015? Should we not use the year when the file has been created? So 2019?

src/Hl7.Fhir.ElementModel/ElementNode.cs
```diff
@@ -0,0 +1,220 @@
...
+using Hl7.Fhir.Utility;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+
+namespace Hl7.Fhir.ElementModel
+{
+    public class DomNode<T> : IAnnotatable where T:DomNode<T>
```
marcovisserFurore: Do you want to have this class public?

src/Hl7.Fhir.ElementModel/ElementNode.cs
```diff
@@ -0,0 +1,220 @@
...
+    {
+        public string Name { get; set; }
+
+        protected List<T> ChildrenInternal = new List<T>();
+
+        internal IEnumerable<T> ChildrenByName(string name = null) =>
```
marcovisserFurore: This name is a bit confusing: When you want all children, then you leave name = null. Why not Children()?

src/Hl7.Fhir.ElementModel/ElementNode.cs
```diff
@@ -0,0 +1,220 @@
...
+                }
+                else
+                    return Name;
+            }
+        }
+
+    }
+
+
+    public class DomNodeList<T> : IEnumerable<T> where T:DomNode<T>
```
marcovisserFurore: Do you want this class public?

src/Hl7.Fhir.ElementModel/SourceNode.cs
```diff
@@ -15,149 +15,106 @@
...
-
-        public void RemoveAnnotations(Type type)
-        {
-            annotations.RemoveOfType(type);
+                : AnnotationsInternal.OfType(type);
         }
-    }
 
-
-    public class ChildNodes : IEnumerable<ISourceNode>
```
marcovisserFurore: ChildNodes will be removed? Is that not breaking the interface?

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
```diff
@@ -3,7 +3,7 @@
  * See the file CONTRIBUTORS for details.
  * 
  * This file is licensed under the BSD 3-Clause license
- * available at https://github.com/FirelyTeam/fhir-net-api/blob/master/LICENSE
+ * available at https://github.com/ewoutkramer/fhir-net-api/blob/master/LICENSE
```
marcovisserFurore: Oops: can you change this back to FirelyTeam?

src/Hl7.Fhir.ElementModel/ElementNode.cs
ewoutkramer: Unfortunately, you cannot make subclasses internal if a public class derives from them, so no, I don't want them public, but I have to.

src/Hl7.Fhir.ElementModel.Tests/SourceNodeTests.cs
ewoutkramer: The code is older, but certainly not from 2015 ;-)

src/Hl7.Fhir.ElementModel/ElementNode.cs
ewoutkramer: It's an internal method that will be called by the actual public Children() method that is part of subclasses of DomNode implementing ITypedElement.  Renaming this to Children() would make the actual ITypedElement.Children() shadow this function.

I did however find a mistake in doing this (SourceNode's Children() should just call base.ChildrenByName()), which I will need to fix.

src/Hl7.Fhir.ElementModel/ElementNode.cs
ewoutkramer: We need to, it's used as the return type of the indexer operator and so is returned to the caller.

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
ewoutkramer: That must be coming from a merge, yes.

src/Hl7.Fhir.Support/Specification/ElementDefinitionSummary.cs
```diff
@@ -6,74 +6,37 @@
  * available at https://github.com/FirelyTeam/fhir-net-api/blob/master/LICENSE
  */
 
-using Hl7.Fhir.Utility;
 using System;
 
 namespace Hl7.Fhir.Specification
 {
-    public class ElementDefinitionSummary : IElementDefinitionSummary
```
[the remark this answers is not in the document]
ewoutkramer: Yes, I had this on the whiteboard behind me, you mean the rename to TypeRootSummary, yes, I need to revert this.

src/Hl7.Fhir.ElementModel/ElementNode.cs
ewoutkramer: I renamed the method to ChildrenInternal(), and renamed the list to ChildList.

src/Hl7.Fhir.Support/Specification/ElementDefinitionSummary.cs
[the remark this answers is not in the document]
ewoutkramer: I re-introduced this class. The API doesn't use it itself, but it's useful anyway.

src/Hl7.Fhir.ElementModel/ElementNode.cs
marcovisserFurore: ah of course!

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
```diff
@@ -182,15 +206,15 @@ private IEnumerable<TypedElementNode> enumerateElements(ElementDefinitionSummary
...
+                string instanceType = info == null ? null :
                     deriveInstanceType(scan, info);
 
-                // If we found a type, but we don't know about the specific child, complain
-                if (dis != ElementDefinitionSummaryCache.Empty && !match.IsTracking)
+                // If we have definitions for the children, but we didn't find definitions for this 
+                // child in the instance, complain
+                if (dis.Any() && info == null)
                 {
-                    raiseTypeError($"Encountered unknown element '{scan.Name}' at location '{scan.Location}' while parsing", this,
```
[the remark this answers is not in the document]
ewoutkramer: Why would a merge overwrite this change? I have no idea....

src/Hl7.Fhir.ElementModel/SourceNode.cs
ewoutkramer: I guess I have. Unfortunately, I have no way to keep this class.  Well, technically we could paste it back in without using it ourselves and then marking it [Obsolete] I guess?

src/Hl7.Fhir.ElementModel/TypedElementOnSourceNode.cs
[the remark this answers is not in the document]
ewoutkramer: So, this file was renamed, this must be at the root of the incorrect merge.  George's changes for #906 got lost too.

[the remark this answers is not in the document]
ewoutkramer: Re-instated the changes for both...

PR #938, opened 2019-04-10
Adding minimal implementation of ITypedElement for constant values

ewoutkramer:
The FhirPath assembly had internal functions to create ITypedElement implementations for primitive constant values.

This PR introduces this functionality in the public surface of the API, via static factory methods on ElementNode


ewoutkramer:
Has conflicts now, will re-issue.

PR #940, opened 2019-04-11
Develop dstu2

brianpos:
* Unit test cleanups
* ModelInspector visibility updates so that users that want to use the ClassMappings don't need to have their own inspector class and reprocess the reflection.
* Revert the property name change in ClassMapping back to ElementType from ImplementingType
* Test Package reference updates
* Add additional stats to the roundtrip unit tests
* set the visibility to FhirPathValueList and Constant value to public so that implementers can add their own custom functions added to the fhirpath symbol table (returning a constant value or a list)


ewoutkramer:
Hi Brian, since the publication of 1.3, we no longer maintain DSTU2.  This branch is closed.  Are you sure you wanted this against DSTU2?

ewoutkramer:
I'm ok with most changes - but not the changes of the visibility. As promised, we're working on that for 1.3.  In fact, there's already a PR for it (#938).  Instead of making this class public, we will have factory methods on ElementNode:

* EmptyList
* CreateConstantList
* CreateConstant

Much like the constructors of ConstantValue and FhirValueList.   Since creating values implementing ITypedElement is useful beyond FhirPath, we've chosen to move this to the ElementModel library!

ewoutkramer:
Replaced by #1001, which has the same content.

Issue #943, opened 2019-04-15
Update ModelInfo.cs to R4


wmrutten:
* `IsConformanceResource()` - also support `GraphDefinition` and `TerminologyCapabilities`
* `IsProfiledQuantity()` - also support `MoneyQuantity`
* `IsInstanceTypeFor()` - also support Uri subtypes

ewoutkramer:
I think the IsInstanceTypeFor() is wrong - there are no Uri subtypes, see here: http://hl7.org/fhir/datatypes.html#primitive

"Although the url and canonical are specializations of uri, they are never substituted for each other"

ewoutkramer:
Unfortunately, this change has been applied as part of #951 - we might have to undo it when the gForge issue is resolved.

ewoutkramer:
The response to our gForge issue is in: https://gforge.hl7.org/gf/project/fhir/tracker/?action=TrackerItemEdit&tracker_item_id=20853&start=0

This makes clear there is *no* substitution between datatypes, will have to adapt IsInstanceTypeFor() accordingly.

PR #946, opened 2019-04-15
Bugfix/943 model info r4

wmrutten:
Fix for issue #943 - update ModelInfo.cs members to support new R4 types


wmrutten:
Woops, invalid branching - going to create a new isolated pull request.

Issue #944, opened 2019-04-15
FhirUrl should implement IStringValue


wmrutten:
[sic]

ewoutkramer:
Fixed in #949.

PR #951, opened 2019-04-15
#944: Update ModelInfo.cs for R4

wmrutten:


wmrutten:
Yay!

ewoutkramer:
See #943 - while pulling this PR we also incorporated  changes to ModelInfo that I think are incorrect and should be reverted depending on the outcome of https://gforge.hl7.org/gf/project/fhir/tracker/?action=TrackerItemEdit&tracker_item_id=20853&start=0

Issue #965, opened 2019-04-29
SnapshotGenerator - type.(target)profile should replace base, not merge


wmrutten:
ElementDefinition.type.(target)profile elements need custom merging behavior:
- Use the differential profiles if the list is not empty
- Otherwise inherit the snapshot profile list
- Do NOT merge differential & snapshot collections; this would prevent derived profiles from constraining the list of (target) profiles.

ewoutkramer:
PR #966 is a fix for this.

PR #966, opened 2019-04-29
fix: #965 SnapshotGenerator - type.(target)profile should replace base, not merge

wmrutten:


wmrutten:
Note: this fix is specific for the R4 branch

PR #978, opened 2019-05-20
Mandatory complex type serialization fix

### LINKED ISSUE #930: Mandatory complex types are not added for _elements
When you ask for _elements and set the MaskingNodeSettings.IncludeMandatory to true, only primitive mandatory elements are included in the serialization. Complex types like Reference and CodeableConcept are not present in the end result.

gstoian:
#930 
Bug fixed. If a child is mandatory to the parent, then all its children will be included. Added a new unit test for this case.


src/Hl7.Fhir.ElementModel/MaskingNode.cs
```diff
@@ -65,10 +65,12 @@ public MaskingNode(ITypedElement source, MaskingNodeSettings settings = null)
                 ies.ExceptionHandler = (o, a) => ExceptionHandler.NotifyOrThrow(o, a);
         }
 
-        private MaskingNode(MaskingNode parent, ITypedElement source)
+        private MaskingNode(MaskingNode parent, ITypedElement source, bool? includeAll = null)
         {
             Source = source;
-            _settings = parent._settings;
+            _settings = parent._settings.Clone();
```
ewoutkramer: No, we cannot do a clone - for performance reasons.  Every single child in every single instance represented by a MaskingNode will now get a copy of the settings, which will cause unnecessary GC pressure.  So, we should store this override in a local variable instead.

src/Hl7.Fhir.ElementModel/MaskingNode.cs
```diff
@@ -65,10 +65,12 @@ public MaskingNode(ITypedElement source, MaskingNodeSettings settings = null)
                 ies.ExceptionHandler = (o, a) => ExceptionHandler.NotifyOrThrow(o, a);
         }
 
-        private MaskingNode(MaskingNode parent, ITypedElement source)
+        private MaskingNode(MaskingNode parent, ITypedElement source, bool? includeAll = null)
```
ewoutkramer: why not make the new parameter mandatory?  That will ensure every call site needs to be explicit about this, which is fine (it's a private method anyway).

src/Hl7.Fhir.ElementModel/MaskingNode.cs
```diff
@@ -152,6 +154,6 @@ public IEnumerable<object> Annotations(Type type)
         }
 
         public IEnumerable<ITypedElement> Children(string name = null) =>
-            Source.Children(name).Where(c => included(c)).Select(c => new MaskingNode(this, c));
+            Source.Children(name).Where(c => included(c)).Select(c => new MaskingNode(this, c, _settings.IncludeMandatory && !Source.Location.StartsWith("Bundle") ? (bool?)true : (bool?)null));
```
ewoutkramer: This seems to suggest ANY (nested) child of Bundle will be exempt, is that correct?  Or did you mean the direct children of Bundle (=1 layer deep)?  The StartsWith() should probably include "Bundle.", to make sure you don't match "BundledMedication" (or such new resource names).

src/Hl7.Fhir.Core.Tests/TestData/summary/bundle-summary-text.json
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+{"resourceType":"Bundle","id":"my-bundle","type":"searchset","total":1803,"entry":[{"fullUrl":"http://base/Patient/patient-one","resource":{"resourceType":"Patient","id":"patient-one","meta":{"versionId":"eric-clapton","tag":[{"system":"http://hl7.org/fhir/v3/ObservationValue","code":"SUBSETTED"}]},"text":{"div":"<div xmlns='http://www.w3.org/1999/xhtml'>A great blues player</div>"},"active":true,"name":[{"use":"official","family":"Clapton"}],"gender":"male","birthDate":"2015-07-09"},"search":{"mode":"match"}},{"fullUrl":"http://base/Patient/patient-two","resource":{"resourceType":"Patient","id":"patient-two","meta":{"versionId":"bb-king","tag":[{"system":"http://hl7.org/fhir/v3/ObservationValue","code":"SUBSETTED"}]},"text":{"div":"<div xmlns='http://www.w3.org/1999/xhtml'>Another great blues player</div>"},"active":true,"name":[{"use":"nickname","family":"King"}]},"search":{"mode":"match"}}]}
```
ewoutkramer: I feel there should not be any change to how summaries are done.  Especially because it seems elements like "active" and "name" are now included - but they are not mandatory?

ewoutkramer: This is not correct - I have pushed an update that fixes this.

PR #979, opened 2019-05-20
Location and ShortPath consistency fix

### LINKED ISSUE #936: TypedElementOnSourceNode does not rewrite Location
Implementers of ISourceNode - which has no type information - will implement the Location property by suffixing all element names with an array indexer (which will be `[0]` for non-repeating items) .

ITypedElement implementers should improve on this (since they have type information) by cleaning up the path, and removing unnecessary indexers for non-repeating elements. This is the case for e.g. ElementNode and PocoElementNode.

TypedElementOnSourceNode, however, simply returns the Location value of its underlying ISourceNode, which is incorrect.

This should be fixed.

gstoian:
#936 
Changed the logic for Location in ElementNode for consistency reasons updated the unit tests and created a new one to test the new property ShortPath.


src/Hl7.Fhir.ElementModel/ElementNode.cs
```diff
@@ -176,6 +176,24 @@ public string Location
...
+                    var myIndex = Parent.ChildList.Where(c => c.Name == Name).ToList().IndexOf(this);
+                    return $"{basePath}.{Name}[{myIndex}]";
+
+                }
+                else
+                    return Name;
+            }
+        }
+
+        public string ShortPath
```
ewoutkramer: Nice, did you also add:

```
  if (type == typeof(ElementNode) || type == typeof(ITypedElement) || type == typeof(IShortPathGenerator))
                return new[] { this };
```

to the `Annotations(Type type)` function?  Otherwise, there is no way to reach out to this ElementNode!

Issue #989, opened 2019-05-29
DirectorySource.Refresh method does not update summary for changed files


wmrutten:
The `DirectorySource.Refresh()` method updates cached summaries for new and removed resources. However when a resource file has changed, the method fails to update the associated summary.

PR #990, opened 2019-05-29
fix: #989 DirectorySource.Refresh() update summaries of changed files

wmrutten:
Fix for the DirectorySource.Refresh() method to also update summaries for existing resource files that have changed.


wmrutten:
This fix can be merged with STU3 and R4 branches.

Issue #929, opened 2019-04-09
Support for Async processing request


brianpos:
In R4 FHIR has included an option for Async requesting content from the FHIR server, which is primarily intended for use with the Bulk Data processing.
We should include support for this.
http://hl7.org/fhir/r4/async.html

brianpos:
This will also need to update the prefer header handling, and also the acceptable status codes to include "202 Accepted"

ewoutkramer:
Well, yes, but this spec is very drafty.  E.g. what would the response be?  For now, the only defined response is for Bulk Data, which returns a non-FHIR json file.  This will require quite some refactoring, since we are not prepared to accept anything that is not a FHIR resource yet.  So - yes, but also: this is not trivial I think.

brianpos:
I'm working on this a bit, and wanted an issue to track it with.

brianpos:
Also, there is no actual response to this call, just a location for the status tracking afterwards in the header. So the real change here was to support the other acceptable status outcomes.

ewoutkramer:
Fixed by #996.

PR #996, opened 2019-05-29
Support the Async prefer header

brianpos:
#929 Support the async prefer header
This simply sets the inclusion of the new prefer header when required, and also supports that the response that comes back will not have content, or an operationoutcome, but a different HTTP status code returned, and a new location to follow up to retrieve the processing status requests.
The following of this link is outside the operation of the FHIR client itself.


brianpos:
As it stands you would have to intercept the result using the OnAfterResult callback to read the location from the header

PR #1003, opened 2019-06-06
_sort did not allow the use of underscore search parameter such as '_lastUpdated'

angusmillar:
When using the '_sort' search result parameter to order the returned results there was a bug in the API in that it checked that values of the _sort must begin with a character that is either a letter or '-' for decreasing order. However, this rule neglected the case where the search parameter being used to sort was one of the 'Parameters for all resources' which all begin with an underscore character '_'. The prime example of this would be '[base]/Patient?_sort=_lastUpdated'

This commit adds support for underscore and a new test case.


ewoutkramer:
Thanks @angusmillar , will do a PR now so it will become part of the upcoming 1.3 release.

ewoutkramer:
I also created a PR (#1003) with a cherry pick for this to STU3.

Issue #465, opened 2017-10-27
[Validator] Detect circular references from instance to instance


ewoutkramer:

marcovisserFurore:
This issue has been marked as stale because it has not had recent activity. It will be closed if no further activity occurs. Thank you  for your contributions.

cknaap:
@marcovisserFurore :I'm afraid VONK-3044 was waiting for this issue to be solved.

PR #1012, opened 2019-06-24
Detect circular references from instance to instance

gstoian:
#465 
The check for circular references takes in consideration the profile and the reference as a pair. Had to adapt the IBucket to pass the list of validated references through the validation flow.
Simple unit test included.


src/Hl7.Fhir.ElementModel/ScopedNode.cs
```diff
@@ -40,155 +26,31 @@ public ScopedNode(ITypedElement wrapped)
...
         }
 
-        public IEnumerable<ITypedElement> Children(string name = null) =>
-            Current.Children(name).Select(c => new ScopedNode(this, this.ParentResource, c));
+        public override IEnumerable<ITypedElement> Children(string name = null) =>
+            Current.Children(name).Select(c => new ScopedNode(this as ScopedNode, this.ParentResource, c));
+    }
+
+
+    public interface IBundledResource
```
ewoutkramer: Did we have this interface before? Why is this introduced?

src/Hl7.Fhir.Specification/Validation/TypeRefValidationExtensions.cs
```diff
@@ -156,16 +160,22 @@ internal static OperationOutcome ValidateResourceReference(this Validator valida
...
                 // references to external entities will operate within a new instance of a validator (and hence a new tracking context).
                 // In both cases, the outcome is included in the result.
-                OperationOutcome childResult;
-
+                OperationOutcome childResult = new OperationOutcome();
+                
                 if (encounteredKind != ElementDefinition.AggregationMode.Referenced)
                 {
-                    childResult = validator.Validate(referencedResource, typeRef.TargetProfile, statedProfiles: null, statedCanonicals: null);
+                    var validatedProfiles = referencedResource is ScopedNodeWrapper sn ? sn.ValidatedProfiles : null;
```
ewoutkramer: I do not understand this bit, maybe you can explain this on a call?

src/Hl7.Fhir.ElementModel/ScopedNode.cs
gstoian: ScopedNode had a public nested class BundledResource, which was moved now into BaseScopedNode. I think when I tried first to use an interface instead of the abstract class I needed the IBundledResource. now that we are using the BaseScopedNode abstract class, the IBundledResourcecan interface can be removed for the moment.

src/Hl7.Fhir.Specification/Validation/TypeRefValidationExtensions.cs
gstoian: referencedResource could in theory be also ScopedNode (actually anything that inherits form BaseScopedNode). I don't think it will ever be anything other than ScopedNodeWrapper, but I still wanted to be sure. 

Even so, the `: null` is definitely wrong given the next line of code. I'll change it.

gstoian:
This needs to be included in v1.5, so I created a new PR with a branch created from 1.x.

https://github.com/FirelyTeam/fhir-net-api/pull/1171

PR closed

Issue #631, opened 2018-07-02
Give more details when encountering unparseable LastModified header


ewoutkramer:
FhirClient throws a pretty generic "String was not recognized as a valid DateTime", we could include a reference to the header being parsed.

See:
https://chat.fhir.org/#narrow/stream/15-dotnet/subject/String.20not.20recognized.20as.20valid.20DateTime

ewoutkramer:
Since chat.fhir.org got updated, this is the new link: https://chat.fhir.org/#narrow/stream/179171-dotnet/topic/String.20not.20recognized.20as.20valid.20DateTime

PR #1014, opened 2019-06-25
Explicit exception when converting header string to datetime 

gstoian:
#631 
Fixed: made the exception more clear for NETSTANDARD1_1

### DIFF
--- a/src/Hl7.Fhir.Core/Rest/HttpToEntryExtensions.cs
+++ b/src/Hl7.Fhir.Core/Rest/HttpToEntryExtensions.cs
@@ -46,7 +46,13 @@ namespace Hl7.Fhir.Rest
 
 #if NETSTANDARD1_1
             if (!String.IsNullOrEmpty(response.Headers[HttpUtil.LASTMODIFIED]))
-                    result.Response.LastModified = DateTimeOffset.Parse(response.Headers[HttpUtil.LASTMODIFIED]);
+            {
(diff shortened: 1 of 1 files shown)


src/Hl7.Fhir.Core/Rest/HttpToEntryExtensions.cs
```diff
@@ -46,7 +46,13 @@ internal static Bundle.EntryComponent ToBundleEntry(this HttpWebResponse respons
 
 #if NETSTANDARD1_1
             if (!String.IsNullOrEmpty(response.Headers[HttpUtil.LASTMODIFIED]))
-                    result.Response.LastModified = DateTimeOffset.Parse(response.Headers[HttpUtil.LASTMODIFIED]);
+            {
+                DateTimeOffset dateTimeOffset = new DateTimeOffset();
+                bool success = DateTimeOffset.TryParse(response.Headers[HttpUtil.LASTMODIFIED], out dateTimeOffset);
+                if (!success)
+                    throw new FormatException($"String \"{response.Headers[HttpUtil.LASTMODIFIED]}\" was not recognized as a valid DateTime");
```
ewoutkramer: The error message should mention that this error was found while parsing the LastModified header, otherwise we're still not providing more context ;-)

ewoutkramer: Such a trivial issue, I fixed it for you.

Issue #865, opened 2019-02-13
Validator accepts Instant type without timezone


marcovisserFurore:
From the FHIR standard:

> instant : An instant in time - known at least to the second and **always includes a time zone**.

Unfortunately the validator accepts resources without a time zone. For example the following resource is accepted by the validator:

```
{
  "resourceType": "DocumentReference",
  "id": "example",
  "status": "current",
  "type": {
    "coding": [
      {
        "system": "http://loinc.org",
        "code": "34108-1",
        "display": "Outpatient Note"
      }
    ]
  },
  "indexed": "2005-12-24T09:43:41",
  "content": [
    {
      "attachment": {
        "contentType": "application/hl7-v3+xml",
        "language": "en-US",
        "url": "http://example.org/xds/mhd/Binary/07a6483f-732b-461e-86b6-edb665c45510",
        "size": 3654,
        "hash": "2jmj7l5rSw0yVb/vlWAYkK/YBwk="
      },
      "format": {
        "system": "urn:oid:1.3.6.1.4.1.19376.1.2.3",
        "code": "urn:ihe:pcc:handp:2008",
        "display": "History and Physical Specification"
      }
    }
  ]
}
```

ewoutkramer:
@gstoian has found that when parsing into a POCO, the POCO will add the system's timezone - so if you validate it after parsing into a POCO, this will go unnoticed.

We will now check whether the same applies when you directly parse into an ITypedElement.

gstoian:
The Pattern to check the string of an Instant type is missing in STU3. The documentation doesn't specify the pattern, but the description states that the time zone is mandatory. I asked the committers if anyone knows if there was a reason why this regex was not included in STU3. In case no good reason comes up, I'll create the PR with the fix.

ArdonToonstra:
Just tested the issue in Simplifier based on version 1.6.
![image](https://user-images.githubusercontent.com/20518398/78690637-00458500-78f8-11ea-9b5b-6bfc6cc66405.png)

Issue can be closed @ewoutkramer

ewoutkramer:
Yes, I have made the POCO parsers more strict - never realized it would fix this issue too ;-)   A welcome regression for a change!

PR #1028, opened 2019-07-03
Fix: Instant type without timezone 

gstoian:
#865 
Added the regex for Instant, updated the profiles files, created a unit test. 

Some files changed (small differences that VS didn't show on commit, I guess it has to do with how the encoding is done).


ewoutkramer:
We should report this using gForge so this correction becomes part of the technical corrections done by HL7.

PR #1036, opened 2019-07-09
fix: function ToString(0 was in R4, but not in STU3 yet

marcovisserFurore:


ewoutkramer:
The check failed because of Azure malfunction.  Marco reran this build here: https://firely.visualstudio.com/fhir-net-api/_build/results?buildId=4119

ewoutkramer:
Should be OK to pull.

Issue #900, opened 2019-03-06
[FhirPath] Add support for defining your own %variables


ewoutkramer:

kennethmyhra:
The thought is that this can better the readability and the process of writing longer FhirPath expressions.

Example with support for defining your own variables:
```
%relatedPersons	- List of RelatedPerson
%patient	- Patient

iif(%relatedPersons.where(relationship.where(coding.where(code = 'represents' and system = 'http://terminology.ehelse.no/CodeSystem/RoleCode'))).exists(), 
	// true part
	%relatedPersons.where(relationship.where(coding.where(code = 'represents' and system = 'http://terminology.ehelse.no/CodeSystem/RoleCode'))).identifier.where(system = 'http://terminology.fhir.org/CodeSystem/MRN').value
,
	// false part
	%patient.identifier.where(system = 'http://terminology.fhir.org/CodeSystem/MRN').value
)
```

Example without support for defining your own variables, reading from a Bundle:
```
iif(entry.where(resource.relationship.where(coding.where(code = 'represents' and system = 'http://terminology.ehelse.no/CodeSystem/Relationship'))).exists(),
	// true part
	entry.where(resource.relationship.where(coding.where(code = 'represents' and system = 'http://terminology.ehelse.no/CodeSystem/Relationship'))).resource.identifier.where(system = 'http://terminology.fhir.org/CodeSystem/MRN').value
,
	// false part
	entry.where(resource is Patient).resource.identifier.where(system = 'http://terminology.fhir.org/CodeSystem/MRN').value
)
```

ewoutkramer:
Yes, and we  need to find a way to provide the value of these things for every call to the evaluator.  I know we discussed doing this using the FpInit (which initialises the symbol table also used for functions) but I realized that's actually wrong, since you do that only once.  So we indeed need a separate way to pass environment variables into the FP evaluator, separate from the predefined ones in the spec.

brianpos:
The fhirpath engine has full support for this already - I've been using it in the SDC processing for some time.
Create a new SymbolTable and pass the base symbolTable in the constructor, then call addVar() to include your variables.
If you need to add in arrays of values, this works too.
``` c#
public static void AddVariable(Hl7.FhirPath.Expressions.SymbolTable table, string name, IEnumerable<ITypedElement> value)
{
    table.Add(name, () => { return value; });
}
```

kennethmyhra:
[...]

PR #1037, opened 2019-07-10
Support variables either internally or externally

brianpos:
#900 Support adding your own variables
There are still a few gotcha's with this, e.g. you shouldn't be doing this with the default symbol table, unless you're wanting to, such as with %ucum
Both modes of use work with CompiledExpression, and only resolve the value at execution time, not plan creation time.
The internal variables are provided through a dictionary in the symbol table, that works through the parent chains, or through external variables via a callback func on the lowest symbol table to resolve the names, and a callback func on the EvaluationContext to retrieve the actual value.


brianpos:
The caching would need to be associated with the SymbolTable rather than global, then caching that would be fine too.

ewoutkramer:
Per our (offline) discussion, will defer this PR until we have discussed the design of the unified engine for mapping + fhirpath, at which time we may re-open.

Issue #1067, opened 2019-08-06
SnapshotGenerator: fix incorrect root element base


wmrutten:
In some circumstances, the SnapshotGenerator raises the `PrepareElement` event with an incorrect base element for the root. Specifically, when expanding a custom Extension definition, the root element is sometimes mapped to `Element`, instead of `Extension`.

wmrutten:
In some circumstances (e.g. recursive profile references), the `SnapshotGenerator` class could leak internal annotations to the caller. This was caused by `DeepCopy` clone operations, which also perform a shallow (!) copy of the annotations on the source element.
Specifically, the `DifferentialTreeConstructor.MakeTree()` method would always clone all the element definitions in the specified differential. The copied annotations could potentially corrupt subsequent snapshots.
Method `MakeTree` has been updated to return the original `ElementDefinition` instances (cloning is not necessary, as the SnapshotGenerator never modifies diff elements). Also, the `SnapshotGenerator` now properly cleans up temporary internal annotations before returning to the caller.

PR #1068, opened 2019-08-06
fix: #1067 Incorrect root element base

wmrutten:
Fix for #1067 - Never expose/leak temporary internal annotations


ewoutkramer:
Included in #1089

PR #1088, opened 2019-08-29
Fix #1051 SnapshotGenerator R4 type slicing

### LINKED ISSUE #1051: R4: Update snapshot generator for type slicing
Update the snapshot generator implementation according to changes discussed with Grahame at DevDays US in Redmond.

Specifically, type slice in snapshot should *always* include slicing component. Slicing component may be omitted from the differential (implicit/redundant for type slice); in that case, the generator should emit a default type slicing component to the snapshot.

WARNING! R4 API packages did NOT process type slices correctly. Once this has been fixed, customers will have to re-generate some existing snapshots.

wmrutten:


wmrutten:
Woops, way too many commits... Going to retry merging this fix into develop-1.x/r4 using cherry picking.

Issue #1101, opened 2019-09-10
SnapshotGenerator - handle element without path


wmrutten:
If the differential component contains an `ElementDefinition` without a `path`, then Forge currently throws an internal ArgumentNullException originating from failed path parsing.
Instead, Forge should throw a specific error message with a detailed error message describing the issue.

PR #1102, opened 2019-09-10
fix: #1101 Handle element constraint without a path [R4]

wmrutten:
Throw a specific exception with a helpful error message when detecting an element constraint without a path in the differential.


marcovisserFurore:
This PR solves #1101 for r4

PR #1115, opened 2019-09-18
Clarify that the data should be converted to DateTimeOffset instead

vadi2:
Clarify that the data should be converted to DateTimeOffset instead subtly via a link to documentation on how to do it. 

It's not immediately obvious what the depreciation message is suggesting:

image


marcovisserFurore:
PRs are never from master, but always from develop*. I changed the base branch from `master-stu3` to `develop-1.x/stu3`

vadi2:
Thanks!

PR #1120, opened 2019-09-23
Feature/fhir client refactoring

gstoian:
Moved away from Core the Requester.cs and other stuff the was not Poco specific. 
Refactored a bit the current FhirClient (created a BaseFhirClient also) and adapted the use of the new Requester implementation. 
Added some unit tests (more to be done).


mmsmits:
replaced with; https://github.com/FirelyTeam/fhir-net-api/pull/1364

Issue #1093, opened 2019-09-05
Client does not close connection on time-out


michelemottini:
See https://github.com/CareEvolution/fhir-net-api/commit/6f0e42a3dce1b7c74c37d70ce7315c3fb50523e0

ewoutkramer:
That seems like a simple fix to apply.... @gstoian? @marcovisserFurore ?

PR #1136, opened 2019-10-07
fix: Call WebRequest.Abort on timeout

kennethmyhra:
* Call WebRequest.Abort on timeout
* Fixes issue #1093


kennethmyhra:
While adding this I noticed that the `#else`-part is missing the timeout-handling, not sure if this is by design or if we should add the same timeout-handling there?

https://github.com/FirelyTeam/fhir-net-api/blob/8f19c5e7792a58d7e0ffb98e0c3577624f001e05/src/Hl7.Fhir.Core/Rest/WebRequestExtensions.cs#L148

Issue #1065, opened 2019-08-05
Constraint Validations having issues with threading


brianpos:
When a validator shares its StructureDefinitions through a resource resolver (such as a CachedResourceResolver) the validator adds in an annotation of the compiled fhirpath expression.
This manipulation of the annotation is not thread safe, so instead we can put a lock on the outside while putting the annotation in.
https://github.com/FirelyTeam/fhir-net-api/blob/8deb078eee04e3ed8c0e34074bad4144fa4b5ee8/src/Hl7.Fhir.Specification/Validation/FpConstraintValidationExtensions.cs#L106
Also, this code should probably be SetAnnotation, not AddAnnotation as don't really want to put in multiple values for this annotation.

brianpos:
I've made this change in my fork to handle this in my server.
https://github.com/brianpos/fhir-net-api/commit/2c0a34c7cd0db328883ff1d64d39319985c79cf3

ewoutkramer:
Yes, we should add something along those lines!

cknaap:
Vonk recently failed a unittest and the cause is probably similar: concurrent access to Annotations.

Exception: `System.IndexOutOfRangeException : Index was outside the bounds of the array.`

Partial Stacktrace:
```
   at System.Collections.Generic.LargeArrayBuilder`1.Add(T item)
   at System.Linq.Enumerable.WhereListIterator`1.ToArray()
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Hl7.Fhir.Utility.ObjectListExtensions.RemoveOfType(IList`1 me, Type t) in d:\a\1\s\common\src\Hl7.Fhir.Support\Utility\ObjectList.cs:line 27
   at Hl7.Fhir.Utility.AnnotatableExtensions.SetAnnotation[A](IAnnotatable annotatable, A annotation) in d:\a\1\s\common\src\Hl7.Fhir.Support\Utility\IAnnotatable.cs:line 30
   at Hl7.Fhir.Specification.Snapshot.SnapshotGenerator.getSnapshotRootElement(StructureDefinition sd, String profileUri, String location) in d:\a\1\s\src\Hl7.Fhir.Specification\Specification\Snapshot\SnapshotGenerator.cs:line 1981
[...]

PR #1148, opened 2019-10-14
Fix: 1065 Constraint Validations having issues with threading

gstoian:
#1065 
Added a lock on constraint element. Instead of adding another annotation, the existing one is set.

This is more of a performance issue, hence I couldn't write a unit test for it.


wmrutten:
Microsoft recommends to always lock on a *private* object, to prevent deadlocks:
https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/lock-statement

brianpos:
I agree, this is a workaround, if you have a suggestion for how to resolve this would be good.
The issue is that the annotation being added isn't thread safe. So thread sharing the StructureDefintion between validators through a caching resolver is where this is impacted.
And since it breaks the instance inside the cache, its really broken from that point on in the server. Surprised Vonk hasn't experienced it yet.

PR #1171, opened 2019-11-12
465 Detect circular references

gstoian:
#465 
The check for circular references takes in consideration the profile and the reference as a pair. Had to adapt the IBucket to pass the list of validated references through the validation flow.
Simple unit test included.


marcovisserFurore:
This should be solved in the new validator

Issue #1174, opened 2019-11-17
TypedElementToSourceNodeAdapter should return IResourceTypeSupplier annotation


alexzautke:
TypedElementToSourceNodeAdapter already implements IResourceTypeSupplier but does not return the annotation. In cases where the underlying SourceNode does not implement IResourceTypeSupplier (like ElementNode) consumers of TypedElementToSourceNodeAdapter may run into issues.

PR #1175, opened 2019-11-18
Adjust TypedElementToSourceNodeAdapterTests

alexzautke:
TypedElementToSourceNodeAdapter should return IResourceTypeSupplier directly and not the underlying SourceNode


marcovisserFurore:
See #1174

Issue #1169, opened 2019-11-06
Setting Code properties is slow


michelemottini:
Around 100 times slower than setting properties of other types.

The issue is that setting a Code property calls ReflectionHelper.GetAttributeOnEnum() (via EnumUtility.GetLiteral()) - that does various reflection operations that are slow compared to a straight property set.

Probably a better way would be to cache the result of GetAttributeOnEnum() or of GetLiteral()

marcovisserFurore:
See also [this](https://chat.fhir.org/#narrow/stream/179171-dotnet/topic/Setting.20Code.20properties.20is.20slow) discussion on Zulip.

@michelemottini also timed this:

```
   var patient = new Hl7.Fhir.Model.DSTU2.Patient();
   var watch = Stopwatch.StartNew();
   const int count = 10_000;
   for ( var i = 0; i < count; i++ )
   {
      patient.Gender = Hl7.Fhir.Model.AdministrativeGender.Female;
   }
   watch.Stop();
   Console.WriteLine( "Set code X {0:N0}: {1:N1} msec", count, watch.ElapsedMilliseconds );
``` 

Output: Set code X 10,000: 62.0 msec
(for non-code properties the time is < 1 msec)

brianpos:
@michelemottini, any suggestions?

michelemottini:
Cache the enum -> string maps

marcovisserFurore:
This PR has been pulled for `develop-1.x` and `develop`

PR #1179, opened 2019-11-21
Improvement: 1169 Setting Code property performance

gstoian:
#1169 
Added a unit test for the performance of Code properties that use Enum literals. Debug only.

Actual fix: https://github.com/FirelyTeam/fhir-net-common/pull/15


marcovisserFurore:
This has been solved by https://github.com/FirelyTeam/fhir-net-common/pull/16.

PR #1182, opened 2019-12-02
Fix: now trims whitespaces of value attributes when serializing to xml

mmsmits:
https://github.com/FirelyTeam/fhir-net-api/issues/1176

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/Serialization/SerializationTests.cs
+++ b/src/Hl7.Fhir.Core.Tests/Serialization/SerializationTests.cs
@@ -254,11 +254,16 @@ namespace Hl7.Fhir.Tests.Serialization
                 }
             };
 
-            var xml = FhirXmlSerializer.SerializeToString(patient);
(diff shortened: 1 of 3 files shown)


brianpos:
I don't really like this change as it will slow down the serializers for something that isn't its fault.
It should be the responsibility of the consumer of the class to do this. Or if you're going to, make an opt out or in setting. 
Dont want to slow things down.

mmsmits:
> 
> 
> I don't really like this change as it will slow down the serializers for something that isn't its fault.
> It should be the responsibility of the consumer of the class to do this. Or if you're going to, make an opt out or in setting.
> Dont want to slow things down.

Good point, this is now a setting, which is turned on by default.

marcovisserFurore:
While the specification says 
> Implementers SHOULD trim leading and trailing whitespace before writing, ...

This means that the FhirXmlSerializer is also in charge of trimming whitespaces. 

@brianpos can you also review this issue?

PR #1204, opened 2020-01-02
Bugfix/1190 logical model resource resolver extensions

### LINKED ISSUE #1190: FindStructureDefinitionForCoreType does not account for logical model data types
FindStructureDefinitionForCoreType does currently append the official FHIR canonical prefix ("http://hl7.org/fhir/StructureDefinition/") to all datatypes. This behaviour does not account for logical models where the type may already be represented by an absolute URL (See StructureDefinition.type: "Absolute URLs are only allowed in logical models").

alexzautke:
This PR fixes #1190

### DIFF
--- a/src/Hl7.Fhir.Specification/Specification/Source/ResourceResolverExtensions.cs
+++ b/src/Hl7.Fhir.Specification/Specification/Source/ResourceResolverExtensions.cs
@@ -47,10 +47,10 @@ namespace Hl7.Fhir.Specification.Source
 
         public static StructureDefinition FindStructureDefinitionForCoreType(this IResourceResolver resolver, string typename)
         {
-            Canonical url;
+            string url;
             if (Uri.IsWellFormedUriString(typename, UriKind.Absolute))
             {
-                url = new Canonical(typename);
+                url = typename;
             }
             else
             {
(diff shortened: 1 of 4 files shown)


marcovisserFurore:
Why do you alter `FindStructureDefinitionForCoreType(` for this? You could also use `FindStructureDefinition()` . So like this

`var sd = resolver.FindStructureDefinition("http://hl7.org/fhir/cda/StructureDefinition/ANY")`

alexzautke:
This issue only occurs for complex non-FHIR data types. In this particular case ```FindStructureDefinitionForCoreType``` being called by the Snapshot generator.

marcovisserFurore:
The problem I have is that the method `FindStructureDefinitionForCoreType` does not cover the name anymore. 

We can also alter the method `FindStructureDefinition()` and check if the parameter `uri` is indeed an uri. If not, then call the method `FindStructureDefinitionForCoreType`. This involves also a change in the class `SnapshotGenerator.getStructureForContentReference` then

src/Hl7.Fhir.Specification/Specification/Source/ResourceResolverExtensions.cs
```diff
@@ -47,7 +47,16 @@ public static StructureDefinition FindStructureDefinition(this IResourceResolver
 
         public static StructureDefinition FindStructureDefinitionForCoreType(this IResourceResolver resolver, string typename)
         {
-            var url = ModelInfo.CanonicalUriForFhirCoreType(typename);
+            string url;
```
marcovisserFurore: I would prefer the following style:
`var url = Uri.IsWellFormedUriString(typename, UriKind.Absolute) ? typename : ModelInfo.CanonicalUriForFhirCoreType(typename)`

But technically it is exact the same, so you can decide what to do with that..

alexzautke: Changed as requested

Issue #1193, opened 2019-12-15
Make TypedElementOnSourceNode support other type models besides FHIR


alexzautke:
tryMapFhirPrimitiveTypeToSystemType in TypedElementOnSourceNode is currently hard-coded to FHIR. When needing to support different type systems (like CCDA), TypedElementOnSourceNode needs to be made more flexible in order to allow custom type -> TypeSpecifier mappings.

brianpos:
I've added stuff to this for my own use in some cases. So I believe it's possible (not just in my fork)

PR #1207, opened 2020-01-08
Bugfix/1193 map primitive type

alexzautke:
PR to prove correct behaviour after change for #1193

### DIFF
--- a/src/Hl7.Fhir.ElementModel.Tests/TypedElementOnSourceNodeTests.cs
+++ b/src/Hl7.Fhir.ElementModel.Tests/TypedElementOnSourceNodeTests.cs
@@ -11,6 +11,7 @@ namespace Hl7.Fhir.ElementModel.Tests
     {
         private readonly IStructureDefinitionSummaryProvider provider = new PocoStructureDefinitionSummaryProvider();
 
+        [ExpectedException(typeof(StructuralTypeException))]
         [TestMethod]
         public void TestExceptionComplexTypeValue()
         {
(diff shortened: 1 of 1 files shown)


src/Hl7.Fhir.ElementModel.Tests/TypedElementOnSourceNodeTests.cs
```diff
@@ -0,0 +1,31 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Hl7.Fhir.ElementModel.Tests
+{
+    [TestClass]
+    public class TypedElementOnSourceNodeTests
+    {
+        private readonly IStructureDefinitionSummaryProvider provider = new PocoStructureDefinitionSummaryProvider();
+
+        [TestMethod]
```
marcovisserFurore: Use here the the following attribute: `[ExpectedException(typeof(StructuralTypeException))]`
Then you do not have to use the try..except

alexzautke: Changed as requested

src/Hl7.Fhir.ElementModel.Tests/TypedElementOnSourceNodeTests.cs
```diff
@@ -0,0 +1,26 @@
...
+        [ExpectedException(typeof(StructuralTypeException))]
+        [TestMethod]
+        public void TestExceptionComplexTypeValue()
+        {
+            var bundleJson = "{\"resourceType\":\"Bundle\", \"entry\":\"Invalid\"}";
+            var bundle = FhirJsonNode.Parse(bundleJson);
+            var typedBundle = bundle.ToTypedElement(provider, "Bundle");
+
+            var value = typedBundle.Children("entry").First().Value;
+            Assert.Fail("Should have thrown on .Value as complex types can't have a value");
```
marcovisserFurore: This Assert.Fail is not needed. The message can be moved to the Attribute:
```
[ExpectedException(typeof(StructuralTypeException), "Should have thrown on .Value as complex types can't have a value")]
```

Issue #1188, opened 2019-12-09
[SnapshotGenerator] render canonicalUrl in elementdefinition.constraint.source instead of FhirType name


mmsmits:
```
<constraint>
        <key value="ele-1" />
        <severity value="error" />
        <human value="All FHIR elements must have a @value or children" />
        <expression value="hasValue() or (children().count() &gt; id.count())" />
        <xpath value="@value|f:*|h:div" />
        <source value="Element" />
</constraint>
```

Should be 

```
<constraint>
        <key value="ele-1"/>
        <severity value="error"/>
        <human value="All FHIR elements must have a @value or children"/>
        <expression value="hasValue() or (children().count() &gt; id.count())"/>
        <xpath value="@value|f:*|h:div"/>
        <source value="http://hl7.org/fhir/StructureDefinition/Element"/>
</constraint>
```

mmsmits:
Pull request: https://github.com/FirelyTeam/fhir-net-api/pull/1213

PR #1213, opened 2020-01-13
fix: #1188 [Snapshot] Render canonicalUrl in elementdefinition.constr…

mmsmits:
#1188 [Snapshot] Render canonicalUrl in elementdefinition.constraint.source instead of FhirType name

### DIFF
--- a/src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
+++ b/src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
@@ -240,7 +240,7 @@ namespace Hl7.Fhir.Specification.Snapshot
         public ElementDefinition MergeElementDefinition(ElementDefinition snap, ElementDefinition diff, bool mergeElementId)
         {
             var result = (ElementDefinition)snap.DeepCopy();
-            ElementDefnMerger.Merge(this, result, diff, mergeElementId);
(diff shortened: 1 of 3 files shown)


src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
```diff
@@ -771,6 +775,18 @@ void createNewElement(ElementDefinitionNavigator snap, ElementDefinitionNavigato
             mergeElement(snap, diff);
         }
 
+        private static void AddConstraintSource(ElementDefinition targetElement, string url)
+        {
+
+            if (targetElement != null && targetElement?.Constraint.Any(c => string.IsNullOrEmpty(c.Source)) == true)
+            {
+                foreach (var constraint in targetElement.Constraint.Where(c => string.IsNullOrEmpty(c.Source)))
```
marcovisserFurore:
```suggestion
                foreach (var constraint in targetElement?.Constraint.Where(c => string.IsNullOrEmpty(c.Source)) ?? Enumeration.Empty())
```

marcovisserFurore: The `if` statement is not needed here.

marcovisserFurore:
@mmsmits : is this only meant for r4?

mmsmits:
> 
> 
> @mmsmits : is this only meant for r4?

Yes! In Stu3 only the type name was rendered

PR #1225, opened 2020-01-27
Fixed some R5 unit tests

mmsmits:
fixed some unit tests


src/Hl7.Fhir.Specification.Tests/Hl7.Fhir.Specification.Tests.csproj
```diff
@@ -63,6 +63,18 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
+
```
marcovisserFurore: This is not needed, because the ItemGroup about it, will handle this. See 
```
<ItemGroup>
    <Content Include="TestData\**\*.xml;TestData\**\*.json">
      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
    </Content>
  </ItemGroup>
```

PR #1236, opened 2020-02-03
fix: correct the dataelements.xml to normal size again

marcovisserFurore:
reducing dataelements.xml to normal size. In the near future this xml file will not be included anymore in the distribution


brianpos:
Yeah, sorry.
Its really not deeded anyway. 
I am planning to implement using the package.tgz file from the build anyway, so those files don't exist in hit anyway.

PR #1259, opened 2020-02-11
CCDA support - Stu3

marcovisserFurore:


src/Hl7.Fhir.Specification/Specification/StructureDefinitionSummaryProvider.cs
```diff
@@ -231,10 +239,11 @@ public XmlRepresentation Representation
         private static bool isResource(ElementDefinition defn) => defn.Type.Count == 1 && 
             (defn.Type[0].Code == "Resource" || defn.Type[0].Code == "DomainResource");
 
+        public string DefaultTypeName => _definition.GetExtensionValue<FhirString>("http://hl7.org/fhir/StructureDefinition/elementdefinition-defaulttype")?.Value;
```
mmsmits: public string DefaultTypeName => _definition.GetStringExtension("http://hl7.org/fhir/StructureDefinition/elementdefinition-defaulttype");

mmsmits: Heel klein dingetje, maar dit is iets netter denk ik..

mmsmits: public string DefaultTypeName => _definition.GetStringExtension("http://hl7.org/fhir/StructureDefinition/elementdefinition-defaulttype");

src/Hl7.Fhir.Core/Specification/PocoStructureDefinitionSummaryProvider.cs
```diff
@@ -158,6 +158,8 @@ string getFhirTypeName(Type ft)
 
         public int Order => _pm.Order;
 
+        public string DefaultTypeName => null;
```
ewoutkramer: Is this member HL7v3/CDA specific?  If so, I think we should include that in the name.  If we can find an "equivalent" in FHIR, then we don't need to do that, but then it should not return null but some sensible value when we are dealing with FHIR data.

src/Hl7.Fhir.Specification/Specification/StructureDefinitionSummaryProvider.cs
```diff
@@ -231,10 +231,11 @@ public XmlRepresentation Representation
...
```
ewoutkramer: Is this member HL7v3/CDA specific? If so, I think we should include that in the name. If we can find an "equivalent" in FHIR, then we don't need to do that, but then it should not return null but some sensible value when we are dealing with FHIR data.

src/Hl7.Fhir.Specification.Tests/TestData/summary-test/TestXmlRepresentation.StructureDefinition.xml
```diff
@@ -0,0 +1,49 @@
...
+  <type value="TestXmlRepresentation" />
+  <baseDefinition value="http://hl7.org/fhir/StructureDefinition/DomainResource" />
+  <derivation value="specialization" />
+  <snapshot>
+    <element id="TestXmlRepresentation">
+      <path value="TestXmlRepresentation" />
+      <min value="0" />
+      <max value="*" />
+    </element>
+    <element id="TestXmlRepresentation.XmlAttr">
```
ewoutkramer: Paths with capital letters? Meaning element names starting with a capital letter?  This will be a problem, since the FhirXmlParser will assume this is a nested resource (in FHIR, only contained resources will have capital inital letters, so there is code depending on this)

src/Hl7.Fhir.ElementModel.Tests/TypedElementOnSourceNodeTests.cs
```diff
@@ -0,0 +1,25 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Hl7.Fhir.ElementModel.Tests
+{
+    [TestClass]
+    public class TypedElementOnSourceNodeTests
+    {
+        private readonly IStructureDefinitionSummaryProvider provider = new PocoStructureDefinitionSummaryProvider();
+
+        [ExpectedException(typeof(StructuralTypeException), "Should have thrown on .Value as complex types can't have a value")]
```
ewoutkramer: No! I never introduced this as an exception - it's not an exceptional circumstance, and I never coded any such constraints into the element model. There might be actually code checking .Value, irrespective of whether it is complex or not. It's not wrong to do that.

alexzautke: There is a call to tryMapFhirPrimitiveTypeToSystemType in the Value Getter. If it's not a primitive raiseTypeError is called. This test was added to make sure that we don't break this part of the code.

src/Hl7.Fhir.Specification.Tests/TestData/summary-test/TestXmlRepresentation.StructureDefinition.xml
ewoutkramer: Turns out this is just a test.

PR #1271, opened 2020-02-27
Remove duplicate validation errors

mmsmits:


brianpos:
Does this handle cases where validating a contained resource referenced by the resource?
I've had a case where a mandatory field was missing in the contained resource, then it reported the issue twice, once with a location contained.status and another through the reference.

src/Hl7.Fhir.Specification/Validation/OperationOutcomeExtensions.cs
```diff
@@ -32,6 +34,35 @@ public static void Flatten(this OperationOutcome outcome)
...
+
+    internal class IssueComparer : IEqualityComparer<OperationOutcome.IssueComponent>
+    {        
+        public bool Equals(OperationOutcome.IssueComponent x, OperationOutcome.IssueComponent y)
+        {
+            if (x == null && y == null)
+                return true;
+            else if (x == null || y == null)
+                return false;
+            else if (x.Location.FirstOrDefault() == x.Location.FirstOrDefault() && x.Details.Text == y.Details.Text)
```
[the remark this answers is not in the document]
mmsmits: Done

mmsmits:
> Does this handle cases where validating a contained resource referenced by the resource?
> I've had a case where a mandatory field was missing in the contained resource, then it reported the issue twice, once with a location contained.status and another through the reference.

Probably not. This solves the issue that as of R4, snapshot generators duplicate all invariants of underlying types into the elementdefinition. Meaning all elements have an explicit ele-1 constraint and all extensions have ext-1 etc. Instead of just referering to their type.
This means that validator now triggers on all bot the explicit ele-1 and on the ele-1 in the type definition after following the type reference. 

This should be solved more elegant by keeping state in the validator etc. But that's for 2.0, for now we just remove duplicate errors in the end. Checking on location and message.

So to answer your question: probably not for errors in referenced resources yet. But we can solve that in 2.0 by keeping state.

brianpos:
Thanks, understand.
I might further investigate my specific issue then.

src/Hl7.Fhir.Specification/Validation/OperationOutcomeExtensions.cs
```diff
@@ -32,6 +34,42 @@ public static void Flatten(this OperationOutcome outcome)
...
+                return false;
+            else if (x.Location?.FirstOrDefault() == x.Location?.FirstOrDefault() && x.Details?.Text == y.Details?.Text)
+                return true;
+            else
+                return false;
+        }
+
+        public int GetHashCode(OperationOutcome.IssueComponent issue)
+        {
+            var hash = issue?.Location?.FirstOrDefault()?.GetHashCode() ^ issue?.Details?.Text?.GetHashCode();
```
marcovisserFurore: There has been a lot of discussion about getting the right Hashcode on the internet. This is one implementation. Using the bitwise xor operation (^) can cause that the value is overflowing (does not fit into an int anymore). For that matter you can use the unchecked statement to prevent that so:

```suggestion
            var hash = unchecked(issue?.Location?.FirstOrDefault()?.GetHashCode() ^ issue?.Details?.Text?.GetHashCode());
```

src/Hl7.Fhir.Specification/Validation/OperationOutcomeExtensions.cs
```diff
@@ -32,6 +34,42 @@ public static void Flatten(this OperationOutcome outcome)
...
+            if(hash is null)
```
marcovisserFurore: This can be shorter: 
```
return (hash is null ? 0 : hash.value);
```

PR #1277, opened 2020-03-03
Embed link to and details about submodule cloning

jmandel:
https://chat.fhir.org/#narrow/stream/179171-dotnet/topic/Unable.20to.20build.20.2ENET.20.20FHIR.20API.20solution/near/189593713 shows confusion about how to clone this repo correctly. The docs exist but aren't readily discoverable; the README should include links + details directly.


marcovisserFurore:
I agree Josh. We are about to release a new version of fhir-net-api, and I can include this PR directly. So, the fastest accepted PR ever that will be directly released

PR #1278, opened 2020-03-03
Impl. of known TerminologyService operations

kennethmyhra:
* ITerminologyService - defined all methods from the [terminology service module](http://hl7.org/fhir/terminology-service.html)
* ExternalTerminologyService - implementation of the new methods
* FallbackTerminologyService - currently just NotImplementedException in the new methods
* Obsoleted the ValidateCode method that already was there - discussion material
* Added new Parameter class for each operation
* Every Parameter class has a ToParameters method which returns a Parameters resource


marcovisserFurore:
Ewout: And: what version would we like to integrate this in? 1.10? 2.0? 2.1? 3.0? It's not breaking the current interface I think, so it could be 1.10. But if we release 2.0 this week, maybe it's worth considering 2.1, or a later 2.0 beta? @marcovisserFurore @kennethmyhra ?

kennethmyhra:
>What are you thinking about, @kennethmyhra ?

@ewoutkramer, there's a fallback for the original `ValidateCode` method which does a local `ValidateCode`. So what I actually was wondering, should there be local fallbacks for these methods, and would it even be possible without implementing a kind of local terminology service. I can easily implement the fallback-sequence equal to `ValidateCode`, but throw a `NotImplementedException` in the `LocalTerminologyService`.

@marcovisserFurore I am guessing it kind of breaks the current interface since it is adding new methods, although that would be easily fixable by anyone that uses `ITerminologyService`. Because of this it would be nice to get it into 2.0, but depends on how ready you guys think this is.

ewoutkramer:
@kennethmyhra  - yes, you are right - this breaks the interface, it should go into 2.0!

We're allright throwing NotImplemented there I think for the fallback.

kennethmyhra:
@ewoutkramer I have added factory methods, let me know if they are according to what you had in mind.

kennethmyhra:
@marcovisserFurore Looked through your changes and looks pretty much good to me
* Makes sense to separate `ValidateCode` into `ValueSetValidateCode` and `CodeSystemValidateCode` instead of using the `typename` parameter as before
* Thumbs up for the refactoring of the `[OperationName]Parameters` classes into builder classes makes it much more readable
* The only thing that might be missing is having an easier way to access the result of the operations that now returns a `Parameters` resource, much the same way that is now done with the **in** parameters `[OperationName]Parameters` classes. Let me know your thoughts on this?

Also, did some minor changes:
* Changed the signature of the `Subsumes` operation from `Resource Subsumes` to `Parameters Subsumes` so it conforms with the other operations that return a `Parameters` resource
* Added test case for ´Closure´ operation

src/Hl7.Fhir.Specification/Specification/Terminology/LocalTerminologyService.cs
```diff
@@ -33,89 +33,148 @@ public LocalTerminologyService(IResourceResolver resolver, ValueSetExpanderSetti
 
         internal ValueSet FindValueset(string canonical)
         {
-// Don't want to redo ITerminologyService yet
+            // Don't want to redo ITerminologyService yet
```
ewoutkramer, line 36: This was about making the ITerminologyService _async_.  Now that we are adding so many new methods to it....shouldn't anything new we add be async?

src/Hl7.Fhir.Specification/Specification/Terminology/ITerminologyService.cs
```diff
@@ -7,11 +7,75 @@
  */
 
 using Hl7.Fhir.Model;
+using System;
 
 namespace Hl7.Fhir.Specification.Terminology
 {
     public interface ITerminologyService
```
ewoutkramer, line 14: I know I am a bit late in realizing this, but shouldn't all these new methods be async?

ewoutkramer, line 14: All these new methods accept the "untyped" Parameters as a parameter, while we also have the nicely "typed" equivalents (like `ExpandParameters`).  Can we also provide extension methods on top of this interface that accept the correct "typed" parameter, and just call these methods  (calling the correct `Build()` to turn them into an instance of `Parameters`)?

ewoutkramer (commented):
Nice work!

src/Hl7.Fhir.Specification/Specification/Terminology/ITerminologyService.cs
marcovisserFurore: Like this?

```c#
        public static Parameters Subsumes(this ITerminologyService srv, SubsumesParameters parameters)
        {
            return srv.Subsumes(parameters.Build());
        }
```

marcovisserFurore: Sigh, you are right. But a lot of work. I'll try to do this today

ewoutkramer: sorry 😢

marcovisserFurore:
> @marcovisserFurore Looked through your changes and looks pretty much good to me
> 
> * Makes sense to separate `ValidateCode` into `ValueSetValidateCode` and `CodeSystemValidateCode` instead of using the `typename` parameter as before
> * Thumbs up for the refactoring of the `[OperationName]Parameters` classes into builder classes makes it much more readable
> * The only thing that might be missing is having an easier way to access the result of the operations that now returns a `Parameters` resource, much the same way that is now done with the **in** parameters `[OperationName]Parameters` classes. Let me know your thoughts on this?
> 
> Also, did some minor changes:
> 
> * Changed the signature of the `Subsumes` operation from `Resource Subsumes` to `Parameters Subsumes` so it conforms with the other operations that return a `Parameters` resource
> * Added test case for ´Closure´ operation

Thanks @kennethmyhra  for reviewing it and the positive feedback. Also Ewout reviewed it and he suggested to make it async, which I did last Friday.

I agree about the returned Parameters. But it is (again) a lot of work. We can leave the interface as it is and change the extension methods later, so it does not break the interface. Perhaps in 2.1 or so?

src/Hl7.Fhir.Core.Tests/Rest/SearchAsyncTests.cs
```diff
@@ -66,20 +67,37 @@ public async Task SearchUsingPost_UsingSearchParams_SearchReturned()
...
             {
                 client.Settings.PreferredFormat = ResourceFormat.Json;
                 client.Settings.PreferredReturn = Prefer.ReturnRepresentation;
                 await searchUsingPost(client);
-            }           
+            }
+        }
+
+        [TestMethod]
+        public async Task SearchUsingPost_Issue1591()
```
kennethmyhra, line 87: This file probably should not be part of this PR?

kennethmyhra (commented):
Looks good!

I also realized handling the output parameters could be a great deal of work. So I agree that we should probably postpone it to a later release.

src/Hl7.Fhir.Core.Tests/Rest/SearchAsyncTests.cs
marcovisserFurore: Oops, you are right. I will fix it right away. Good catch!

marcovisserFurore (approved):

Issue #1312, opened 2020-04-06
DirectorySource and possible others should be async


ewoutkramer:
Our current implementations of IResourceResolver are synchronous in nature while most of them will involve I/O. It would be much better to use async methods here to improve server throughput.

ewoutkramer:
I add infrastructure to create asynchronous resolvers, and also updated the snapshot generator to use them.

Note: API still has many places where it uses the now-obsolete synchronous calls.

PR #1323, opened 2020-04-14
#1312 Make resolvers & snapgen work asynchronously [R3]

ewoutkramer:
Add infrastructure to create asynchronous resolvers, and update the snapshot generator to use them.

Note: API still has many places where it uses the now-obsolete synchronous calls.


src/Hl7.Fhir.Specification/Specification/StructureDefinitionSummaryProvider.cs
```diff
@@ -49,11 +52,16 @@ public IStructureDefinitionSummary Provide(string canonical)
                 if (!mapSuccess) return null;
             }
 
-            var sd = _resolver.FindStructureDefinition(mappedCanonical, requireSnapshot: true);
-            if (sd == null) return null;
-
-            return new StructureDefinitionComplexTypeSerializationInfo(ElementDefinitionNavigator.ForSnapshot(sd));
+            var sd = await _resolver.FindStructureDefinitionAsync(mappedCanonical);
+            
+            return sd == null ?
```
marcovisserFurore, line 57:
```suggestion
            return sd is null ? 
```

PR #1326, opened 2020-04-16
Removed resourcetype.episodeofcare as target for diagnosticreport.encounter parameter

### LINKED ISSUE #1298: Model info: wrong target for STU3 DiagnosticReport.encounter parameter
It should be just `Encounter` - see 
https://github.com/FirelyTeam/fhir-net-api/blob/develop-1.x/stu3/src/Hl7.Fhir.Core/Model/Generated/Template-ModelInfo.cs#L965

mmsmits:
fixes #1298


brianpos:
Isn't this code generated?
Does this mean that the core spec was incorrect?
(was it resolved in the core spec, and this could occur again if we have a tech correction done - unlikely)

Issue #1325, opened 2020-04-16
SnapshotGenerator: Expand copies referenced data type extensions to elementDef


abelenthoven:
When a ElementDefinition is expanded, referenced StructureDefinitions are loaded for data types. In 4.0.1, these data types contain extensions (on StructureDefinition level). Somehow these extensions are copied to the element that references the data type when using SnapshotGenerator.Expand

abelenthoven:
Has been resolved in the SnapshotGenerator (in 1.x). As it turned out, extensions on the root element in a referenced StructureDefinition were copied to the element in the referencing definition, resulting in any ElementDefinition getting the extensions defined in the root ElementDefinition of 'Element'

PR #1346, opened 2020-04-28
#1325 - Removed root element extension inheritance in profiles

abelenthoven:
In this PR the inheritance of a StructureDefinition's root element extensions is removed. This fix has to be ported to R5 and STU3 as well, although in the latter the inheritance does not lead to problems


marcovisserFurore:
Fixes #1325

PR #1347, opened 2020-04-28
#1325 - Removed extensions inheritance of root elements in a StructureDefinition

abelenthoven:
Removed extensions inheritance of root elements in StructureDefinition snapshotting


marcovisserFurore:
Fixes #1325

Issue #1303, opened 2020-03-24
Use of profiled datatypes within a SD results in invalid snapshot


MatthijsvdWielen:
Generating a snapshot using Simplifier or Vonk (3.3 / 3.4) of one of the following profile will create invalid results:
https://simplifier.net/packages/de.medizininformatikinitiative.kerndatensatz.laborbefund/0.9.0/files/114053/~json
https://simplifier.net/nictizstu3-zib2017/nl-core-patient

Some elements will be assigned a "max" of "*" whereas max is greater than base.max of "1". This should not be the case as it is not allowed.
The element DiagnosticReport.identifier:befund.assigner is generated incorrect in the german profile.
The elements Patient.name and Patient.address are generated incorrect in the Nictiz profile. 

The error seems to be in the snapshot generator. Downloading the snapshot from Simplifier shows the error in cardinality. While the diff is correct.

-  It seems to be caused by use of profiled datatype within a StructureDefinition.
-  The StructreDefinition XML seems ok. Doens’t violate the base. Seems to be an error in the snapshot.
-  Forge renders the 0..* as well. Instead of the base 0..*
-  The Simplifier validator seems to be ok with sending more than the base cardinality should allow. 
-    Another thing that is remarkable and maybe the cause of this, is that the root elements have cardinalities of 0..* as well. Perhaps by adding a profiled datatype, the root element is used for the cardinality.

ewoutkramer:
Related to #1108 ?

ewoutkramer:
Yes - this is all consistent and all about the same issue. There's simply no computable rule/invariant in the FHIR spec that would trigger validation warnings about this. But the core of the issue is that there are several inconsistent cardinalities in play: the one coming from the base and the one in the profile. For cardinality, it seems to make sense to take the "lowest" value of the two, but for other constraints, it is not so easy - or impossible.  I'll take a look at solving this at least for the cardinality.

PR #1351, opened 2020-04-30
#1303 - don't allow max to go from lower to higher

ewoutkramer:
When merging ElementDefs, the Max (cardinality) should never be merged in such a way that it is higher than the lowest of the base and diff.

### DIFF
--- a/src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefnMerger.cs
+++ b/src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefnMerger.cs
@@ -24,7 +24,7 @@ namespace Hl7.Fhir.Specification.Snapshot
         /// Private static helper for <see cref="SnapshotGenerator"/>.
         /// Merge two <see cref="ElementDefinition"/> instances and all their properties.
         /// </summary>
-        struct ElementDefnMerger
+        internal struct ElementDefnMerger
         {
             /// <summary>Merge two <see cref="ElementDefinition"/> instances. Existing diff properties override associated snap properties.</summary>
             public static void Merge(SnapshotGenerator generator, ElementDefinition snap, ElementDefinition diff, bool mergeElementId)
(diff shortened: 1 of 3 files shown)


src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefnMerger.cs
```diff
@@ -244,6 +245,58 @@ void onConstraint(Element snap)
...
+                if (!diff.IsNullOrEmpty() && !snap.IsNullOrEmpty() && !diff.IsExactly(snap))
+                {
+                    if (diff.Value == "*")
+                        return snap;
+
+                    // Now, diff has a numeric limit
+                    // So, if snap has no limit, take the diff
+                    if(snap.Value == "*")
+                    {
+                        var result = (FhirString)diff.DeepCopy();
```
marcovisserFurore, line 268: This part is coming back 3 times, perhaps create an inline function for this:

```
FhirString deepCopy(FhirString elt)
{
   var result = (FhirString)elt.DeepCopy();
   onConstraint(result);
   return result;
}
```

ewoutkramer: @marcovisserFurore , I tried that, but unfortunately, ElementDefnMerger is a struct - and local functions cannot create a closure over instance variables like `onConstraint` within a struct. I guess Michel made this a struct since it's being created quite often.

Anyway, since we both don't like copy & paste - I turned it into a full blown member function ;-)

ewoutkramer:
> Good catch!
> 
> In R4 the function `mergePrimitiveAttribute` does not exists anymore. So we have to find a solution for that. Maybe the function `mergePrimitiveElement` does exactly the same.
> 
> Also a minor enhancement requested in `ElementDefnMerger.cs`

It seem the function has been renamed to `mergePrimitiveElement` indeed, I see a few differences, but it's the same function.

PR #1363, opened 2020-05-11
Feature/update20r4 with 1xr4

marcovisserFurore:
Update 2.0 with changes from 1.7 for R4


marcovisserFurore:
This is forward merge from 1.7 to 2.0 (R4) and a forward merge from 2.0 STU3 to 2.0 R4

PR #1365, opened 2020-05-12
Name the backbones according to the `explicit-type-name` extension (for R4)

ewoutkramer:
Since R3 quite a few backbones carry an explicit name, given in the "http://hl7.org/fhir/StructureDefinition/structuredefinition-explicit-type-name" extension. Use this name for these types, including in the generated TypeName property.


brianpos:
@ewoutkramer   Make sure this one gets into the release notes, as there could be folks like me that could be effected by this one...
(I use it in some interim storage in some places - and I thought this was already used by the code generator)

ewoutkramer:
It was already used by the code generator, but the name "Component" was added for the C# type name (fine), but also for the real "FHIR" typename (not what I wanted now).  And to make them unique, I have added the resource name before them.

To conclude: the names for the C# types do not change, but the name of the TypeName changes to say  "Patient#Contact"  (where it was ContactComponent before).

The whole reason for this is that these type names now surface due to the FhirPath normative `type()` function, you can call this on `Patient.contact` too, and then what will appear as the name?

I discussed this with Bryn, and we are working towards the situation where our API and the CQL spec align.

brianpos:
Was that for the R5 stuff?
Each release I had to go in and tweak through that, I thought this was the name that James does in the HAPI libs too (using what we had, and likely gets this issue too)

ewoutkramer:
No, this is for the FhirPath normative stuff. And we really need this to be unique, hence the prefix.

ewoutkramer:
This will be part of the https://github.com/FirelyTeam/fhir-net-api/wiki/Breaking-changes-in-2.0 page!

brianpos:
Then yes, definitely include in the release notes, Is this also effecting the classname generated?
(if not then note quite so bad as won't impact compile time - they should be unique anyway)

ewoutkramer:
Should we include James in this discussion too?

brianpos:
HAPI will have the same issues (or maybe Grahame knows about it too - don't recall if he wrote that part of the JAVA stack or not)

ewoutkramer:
> Then yes, definitely include in the release notes, Is this also effecting the classname generated?
> (if not then note quite so bad as won't impact compile time - they should be unique anyway)

No, the .NET classnames remain the same.

src/Hl7.Fhir.Core/Model/Source/search-parameters.xml
```diff
@@ -13159,7 +13159,8 @@
         <xpath value="f:Composition/f:encounter | f:DeviceRequest/f:encounter | f:DiagnosticReport/f:encounter | f:DocumentReference/f:context/f:encounter | f:Flag/f:encounter | f:List/f:encounter | f:NutritionOrder/f:encounter | f:Observation/f:encounter | f:Procedure/f:encounter | f:RiskAssessment/f:encounter | f:ServiceRequest/f:encounter | f:VisionPrescription/f:encounter" />
         <xpathUsage value="normal" />
         <target value="Encounter" />
-        <target value="EpisodeOfCare" />
+        <!-- removed, see: https://github.com/FirelyTeam/fhir-net-api/issues/1298 -->
```
marcovisserFurore, line 13162: I agree that we should change the source to solve this issue. But should we not also change this file in `Hl7.Fhir.Specification.data`?

ewoutkramer: Turns out this specific file only exists in Source (unlike profiles-types and the others).

mmsmits: I just removed it

marcovisserFurore: Thanks @mmsmits. So `Hl7.Fhir.Specification.data.search-parameter.xml` did exist. But has now been synchronized with `Hl7.Fhir.Core/Model/Source/search-parameters.xml`

PR #1366, opened 2020-05-14
Removed EpisodeOfCare as target from multiple encounter searchparameters in STU3

### LINKED ISSUE #1298: Model info: wrong target for STU3 DiagnosticReport.encounter parameter

mmsmits:
fixes #1298


marcovisserFurore:
@mmsmits : is there a PR coming for develop-1.4/r4 as well?

Issue #1374, opened 2020-05-24
Add support for FHIR Patch operations


valeneiko:
FHIR defines patch operations: https://www.hl7.org/fhir/fhirpatch.html
But [fhir-net-api](/FirelyTeam/fhir-net-api) does not support them.

There was a similar issue (#813) open previosuly, requesting this, but was closed, with no support for patch operations added.

mmsmits:
Patch was added in a previous release

PR #1375, opened 2020-05-24
Implemented Fhir Patch

anyname2:
Added support for FHIR Patch operations. Only FHIRPath patch operation is implemented.
Addresses: #1374


marcovisserFurore:
@anyname2 , a big thank you for your contribution. We are investigating your PR at the moment and discussing it to add it to our fhir-net-api.

src/Hl7.Fhir.Patch.Tests/PatchDocumentTests.cs
```diff
@@ -0,0 +1,347 @@
...
+using Hl7.Fhir.Model;
+using Hl7.Fhir.Patch.Exceptions;
+using Hl7.Fhir.Serialization;
+using Hl7.Fhir.Specification;
+using Hl7.FhirPath;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Hl7.Fhir.Patch.Tests
+{
+    [TestClass]
```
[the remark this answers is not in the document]
cknaap: Also the logErrorAction is not tested anywhere.

src/Hl7.Fhir.Core/ElementModel/PocoBuilder.cs
```diff
@@ -51,18 +52,20 @@ public Base BuildFrom(ISourceNode source, Type dataType = null)
                     return null;
                 }
             }
-            return BuildFrom(source, typeFound);
+
```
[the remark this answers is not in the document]
anyname2: When making these changes, my intention was to fix the broken path (throwing `InvalidCastException` for custom POCO's) while not chaning the behaviour of any other case.

src/Hl7.Fhir.Core/Model/FhirPatch.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Collections.Generic;
+using System.ComponentModel.DataAnnotations;
+using System.Runtime.Serialization;
+
+namespace Hl7.Fhir.Model
+{
+    [FhirType("Parameters")]
+    [InvokeIValidatableObject]
+    [DataContract]
+    public class FhirPatch : Parameters
```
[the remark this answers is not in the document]
anyname2: I had 2 concerns when deciding to implement the POCO:

- Parameters object might be different between different versions of FHIR, and would require different Reader implementation. If this is not the case I can try implementing reading `PatchDocument` directly from `ITypedElement` within the `Fhir.Patch` project in `common`.

- Possibility to have FhirPatch as a `[FromBody]` input parameter in an ASP.NET Controller method
    ```csharp
    [HttpPatch]
    [Route(KnownRoutes.ResourceTypeById)]
    public async Task<IActionResult> Patch(string typeParameter, string idParameter, [FromBody]FhirPatch resource)
    ```
    Would it still work it without `FhirPatch` POCO?

anyname2:
I have addressed the comments here and in FirelyTeam/fhir-net-common#50.
Please review again and let me know if there are any further changes necessary.

src/Hl7.Fhir.Core/Model/Generated/Template-ModelInfo.cs
```diff
@@ -632,6 +632,7 @@ public static string Version
                 { typeof(Hl7.Fhir.Model.ValueSet), "ValueSet" },
                 { typeof(Hl7.Fhir.Model.VerificationResult), "VerificationResult" },
                 { typeof(Hl7.Fhir.Model.VisionPrescription), "VisionPrescription" },
+				 { typeof(Hl7.Fhir.Model.FhirPatchParameters), "Parameters" },
```
marcovisserFurore, line 635: This is not allowed. This file `Template-ModelInfo` will be generated, so all changes here will be overriden by a new generate. Also `FhirPathParameters` is not a standard Fhir resource (https://www.hl7.org/fhir/resourcelist.html)

src/Hl7.Fhir.Core/Model/FhirPatchParameters.cs
```diff
@@ -0,0 +1,61 @@
+﻿/*
```
marcovisserFurore, line 1: I don't see the added value of this class. I rather don't want to add it to the core library.

src/Hl7.Fhir.Core/Model/Generated/Template-ModelInfo.cs
anyname2: This was regenerated. There is a corresponding change to the template: `src/Hl7.Fhir.Core/Model/Template-Model.tt`.
Though I am not sure how to achieve what you are suggesting.

src/Hl7.Fhir.Core/Model/FhirPatchParameters.cs
anyname2: My intent with adding this class was to support `PatchDocument` as a `[FromBody]` input parameter in an ASP.NET Controller method. But current parser implementation only deserializes types derived from `Base`.

```csharp
[HttpPatch]
[Route(KnownRoutes.ResourceTypeById)]
public async Task<IActionResult> Patch(string typeParameter, string idParameter, [FromBody]FhirPatchParameters resource)
```

The benefit from having it as input parameter, rather than constructing it in the method body is implicit validation performed on input paremeters. 

I guess it is ok not addding it to the core library, since the class can be implemented by consumers of the library. Though I am not sure how to do that, since implementing this class required change to `ModelInfo` template ([see](https://github.com/FirelyTeam/fhir-net-api/pull/1375#discussion_r450845167))

marcovisserFurore:
We've decided to implement the `Patch` operation in another way. Therefor I close this PR.

PR #1381, opened 2020-05-27
Added test to verify enums can now be used as primitive values (for R3)

ewoutkramer:
Enums will be treated as System.Code primitives

### DIFF
--- a/src/Hl7.Fhir.ElementModel.Tests/ElementNodeTests.cs
+++ b/src/Hl7.Fhir.ElementModel.Tests/ElementNodeTests.cs
@@ -25,34 +25,34 @@ namespace Hl7.FhirPath.Tests
     [TestClass]
     public class ElementNodeTests
     {
-        readonly IStructureDefinitionSummaryProvider provider = new PocoStructureDefinitionSummaryProvider();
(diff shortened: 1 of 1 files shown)


ewoutkramer:
Note that I will need to create the R4 version of this PR, since it will require changes to the FhirPath engine.

src/Hl7.Fhir.ElementModel.Tests/ElementNodeTests.cs
```diff
@@ -285,13 +285,34 @@ public void CannotUseAbstractType()
...
+                    entry.ToTypedElement(_provider, "Element");
                 Microsoft.VisualStudio.TestTools.UnitTesting.Assert.Fail("Should have thrown on invalid Div format");
             }
             catch (ArgumentException)
             {
             }
         }
 
+        [TestMethod]
+        public void CreateFromPrimitive()
```
marcovisserFurore, line 297: Should this test not be in the fhir-net-common project?

ewoutkramer: Ik denk het van wel.

PR #1413, opened 2020-07-09
fix: removed LastReponse and LastRequest from FhirClient and LegacyFhirClient

### LINKED ISSUE #1132: FhirClient.LastResponse is disposed
Because it is set within a `using` clause: https://github.com/FirelyTeam/fhir-net-api/blob/develop-stu3/src/Hl7.Fhir.Core/Rest/Requester.cs#L103 . . . and so it cannot really be used for anything (eg see https://referencesource.microsoft.com/#system/net/system/Net/HttpWebResponse.cs,302)

mmsmits:
fixes #1132

Removed LastResponse and LastRequest from the FhirClient(s). They were already disposed, so no point having them around.

### DIFF
--- a/src/Hl7.Fhir.Core/Rest/FhirClient.cs
+++ b/src/Hl7.Fhir.Core/Rest/FhirClient.cs
@@ -19,6 +19,9 @@ namespace Hl7.Fhir.Rest
 {
     public partial class FhirClient : BaseFhirClient
     {
+//disables warning that OnBeforeRequest and OnAfterResponse are never used.
+#pragma warning disable CS0067
+        
         /// <summary>
         /// Creates a new client using a default endpoint
         /// If the endpoint does not end with a slash (/), it will be added.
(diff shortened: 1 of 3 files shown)


src/Hl7.Fhir.Core/Rest/Legacy/LegacyFhirClient.cs
```diff
@@ -174,18 +173,21 @@ public ParserSettings ParserSettings
...
             set => Settings.ParserSettings = value;
         }
-        
+
         /// <summary>
         /// Returns the HttpWebRequest as it was last constructed to execute a call on the FhirClient
         /// </summary>
-        public HttpWebRequest LastRequest { get { return LastClientRequest as HttpWebRequest; } }
+        /// 
+        [Obsolete("LastRequest was already disposed, so no point in having them around", true)]
```
ewoutkramer, line 181: Or should we just remove these?  At least they should throw instead of returning null.

src/Hl7.Fhir.Core/Rest/FhirClient.cs
```diff
@@ -182,25 +173,12 @@ public ParserSettings ParserSettings
...
-        /// Returns the HttpWebResponse as it was last received during a call on the FhirClient
-        /// </summary>
-        /// <remarks>Note that the FhirClient will have read the body data from the HttpWebResponse, so this is
-        /// no longer available. Use LastBody, LastBodyAsText and LastBodyAsResource to get access to the received body (if any)</remarks>
-        public HttpWebResponse LastResponse { get { return LastClientResponse as HttpWebResponse; } }
-
         #endregion
 
-        [Obsolete ("OnBeforeRequest is deprecated, please add a HttpClientEventHandler to the constructor to use this functionality", true)]
+        [Obsolete ("OnBeforeRequest is deprecated, please add a HttpClientEventHandler or another HttpMessageHandler to the constructor to use this functionality", true)]
```
ewoutkramer, line 178: We should touching these events throw an exception - or remove them completely.

ewoutkramer:
And: what version would we like to integrate this in?  1.10? 2.0?  2.1? 3.0?

src/Hl7.Fhir.Core/Rest/Legacy/LegacyFhirClient.cs
marcovisserFurore: The true at the end of Obsolete will cause a compiler warning, so that's good enough.

src/Hl7.Fhir.Core/Rest/FhirClient.cs
marcovisserFurore: The true at the end of Obsolete will cause a compiler warning, so that's good enough.

PR #1419, opened 2020-07-13
fix: snapshot generator: ignore invalid sliceName error on SimpleQuantity root

### LINKED ISSUE #618: Fix SimpleQuantity datatype definition, remove invalid sliceName on root element
In STU3, the SimpleQuantity datatype definition in profiles-types.xml specifies an invalid non-empty sliceName on the root element

```
          <element id="Quantity:simplequantity">
            <path value="Quantity"/>
            <sliceName value="SimpleQuantity"/>
```
This is illegal and causes the snapshot generator to always emit an error for each profile that references SimpleQuantity datatype, e.g. when expanding an Observation profile. Although technically, the error message is correct and indicates a bug in the spec, this is not very helpful and confusing to end users.

Propose to simply hack the included profiles-types.xml in specification.zip and remove the illegal sliceName property value from the SimpleQuantity datatype profile.

mmsmits:
Fixes #618

Ignores invalid slicename error on SImpleQuantity root.

### DIFF
--- a/src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
+++ b/src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
@@ -3194,12 +3194,9 @@ namespace Hl7.Fhir.Specification.Tests
             Assert.AreEqual(nav.Current.Type.FirstOrDefault().Code, FHIRAllTypes.Integer.GetLiteral());
 
             Assert.IsNotNull(outcome);
-            // [WMR 20170810] Fixed, now also expecting issue about invalid slice name on SimpleQuantity root element
-            //Assert.AreEqual(1, outcome.Issue.Count);
-            // assertIssue(outcome.Issue[0], SnapshotGenerator.PROFILE_ELEMENTDEF_INVALID_CHOICE_CONSTRAINT);
-            Assert.AreEqual(2, outcome.Issue.Count);
-            assertIssue(outcome.Issue[0], SnapshotGenerator.PROFILE_ELEMENTDEF_INVALID_SLICENAME_ON_ROOT);
-            assertIssue(outcome.Issue[1], SnapshotGenerator.PROFILE_ELEMENTDEF_INVALID_CHOICE_CONSTRAINT);
+
+            Assert.AreEqual(1, outcome.Issue.Count);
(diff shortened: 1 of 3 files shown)


src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
```diff
@@ -436,7 +436,10 @@ void fixInvalidSliceNameOnRootElement(ElementDefinition elem, StructureDefinitio
                 Debug.Assert(elem.IsRootElement());
                 if (!string.IsNullOrEmpty(elem.SliceName))
                 {
-                    addIssueInvalidSliceNameOnRootElement(elem, sd);
+                    if(sd.Url != ModelInfo.FhirCoreProfileBaseUri + ModelInfo.FhirTypeToFhirTypeName(FHIRAllTypes.SimpleQuantity))
```
marcovisserFurore, line 439: This would be better:
```
if (sd.Url != ModelInfo.CanonicalUriForFhirCoreType(FHIRAllTypes.SimpleQuantity))
```

src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
```diff
@@ -4478,13 +4473,25 @@ public void TestSimpleQuantityProfile()
...
-        // Complex extension:   structure.cdstools-typedstage
-        // Referencing Profile: structure.cdstools-basecancer
-        // Profile defines constraints on child elements of the complex extension
-        // Snapshot generator adds slicing component to Condition.extension.extension.extension:type - WRONG!
-        [TestMethod]   // test data needs to be converted from dstu2 -> stu3
+        //Ignore invalid slice name error on the root of SimpleQuantity.
+        [TestMethod]
+        public void TestSimpleQuantity()
+        {
+            var resource = _testResolver.FindStructureDefinition("http://hl7.org/fhir/StructureDefinition/SimpleQuantity");
```
marcovisserFurore, line 4480: Maybe user `ModelInfo.CanonicalUriForFhirCoreType(FHIRAllTypes.SimpleQuantity)`

mmsmits:
It not only suppresses the error. It also fixes the snapshot (the generator already did that before). But I can check.

marcovisserFurore:
> It not only suppresses the error. It also fixes the snapshot, by removing the sliceName from the root (the generator already did that before). But I can check.

Ah I see now:
It was already removing the SliceName from the root. But now we suppress the error about it. Then I think the validation will be okay

Issue #1418, opened 2020-07-13
Add expansion parameters to ValueSet $expand


MatthijsvdWielen:
There is currently limited support for $expand on Vonk: http://docs.simplifier.net/vonk/features/terminology.html#valueset-expand
Because expansion parameters are not supported in the api at the moment, these can not be implemented on Vonk. 

Feature request: Would it be possible to add funtionality to the api so that includeDesignations are shown in expansion? 

Expansion parameters: https://www.hl7.org/fhir/valueset.html#parameters

marcovisserFurore:
Work: 
- 8 hours + review and unittests
- release in 1.9

PR #1423, opened 2020-07-16
feature: add includeDesignations parameter in ValueSet expander

mmsmits:
fixes: #1418

Adds includeDesignations optionality in VS expander


src/Hl7.Fhir.Specification/Specification/Terminology/ValueSetExpander.cs
```diff
@@ -264,9 +289,27 @@ internal static ValueSet.ContainsComponent ToContainsComponent(this CodeSystem.C
 
             if (source.Concept.Any())
                 newContains.Contains.AddRange(
-                    source.Concept.Select(c => c.ToContainsComponent(system)));
+                    source.Concept.Select(c => c.ToContainsComponent(system, settings)));
 
             return newContains;
         }
+
+        private static List<ValueSet.DesignationComponent> ToValueSetDesignations(this List<CodeSystem.DesignationComponent> csDesignations)
```
marcovisserFurore, line 297: Throughout the api we normally don't prefix the parameters, variables, etc with their abbreviated types. It is not wrong, but it is more a certain style

Issue #1412, opened 2020-07-08
The search parameter _include and _revinclude  does not support iterate


whyfate:
http://build.fhir.org/search.html#include
<code>:iterate</code> is not supported.

hryh:
To support `:iterate`, just a `IList<string>` is not enough,I think.

ewoutkramer:
this will be fixed in 2.0

PR #1433, opened 2020-07-20
fix: Add modifier support for include and  revinclude params

### LINKED ISSUE #222: _include with modifier is not recognized as an include by Add
Wondering how to best to handle _include:recurse= . The SearchParams.add put them as a regular parameter instead of part of _include collection. Wondering if i should add a IncludeRecurse collection.

mmsmits:
fixes #222 and #1412


src/Hl7.Fhir.Core/Rest/FhirClientSearch.cs
```diff
@@ -604,14 +739,31 @@ private SearchParams toQuery(string[] criteria, string[] includes, int? pageSize
...
+        private (string path, IncludeModifier? modifier)[] stringToIncludeTuple(string[] includes)
+        {
+            if(includes != null && includes.Any())
+            {
+                var modifierIncludes = new List<(string, IncludeModifier?)>();
+                foreach (var i in includes)
+                {
+                    modifierIncludes.Add((i, null));
+                }
+                return modifierIncludes.ToArray();
```
ewoutkramer, line 751: includes.Select(i => (i, (IncludeModifier?)null)).ToArray()?

ewoutkramer: and even better (re: our discussion):   `includes.Select(i => (i, IncludeModifier.None)).ToArray()`.

mmsmits: Done!

PR #1455, opened 2020-07-31
Use MS code-generator for resources and datatypes (for R3)

ewoutkramer:
The Model classes are now being generated by the Microsoft code generator, instead of the T4 templates.


BradBarnich:
How does this work? what is doing the code generation?

ewoutkramer:
The generator can be found here: https://github.com/microsoft/fhir-codegen

I am working with @GinoCanessa to get it working with the new 2.0 API.

I also fixed it for the current 1.x branch, but unlikely we'll be introducing it there due to the fact that we don't plan any new versions in 1.x anymore...

ewoutkramer:
The reason for switching is that we here are Firely hardly dared touch the T4 templates anymore - they'd become full of duplicate code and very intricate to change. And so when Gino came up with a freshly designed generator, we were pretty happy to jump ship.

BradBarnich:
Oh we understand 😄 https://github.com/CareEvolution/fhir-net-api/blob/multiversions/src/Hl7.Fhir.Core/Model/Generate.csx

That's cool

ewoutkramer:
> Oh we understand 😄 https://github.com/CareEvolution/fhir-net-api/blob/multiversions/src/Hl7.Fhir.Core/Model/Generate.csx
> 
> That's cool

Oh, right, @michelemottini spoke to me about this, you have done that exercise before....

marcovisserFurore (approved):
I checked manually the generated files to the originals (a take some samples), and I found no anomalies.

@ewoutkramer : shall we add _GeneratorLog.cs to Git ignore?

PR #1457, opened 2020-08-03
Use MS code-generator for resources and datatypes (for R4)

ewoutkramer:
The Model classes are now being generated by the Microsoft code generator, instead of the T4 templates.


marcovisserFurore (approved):
I checked manually the generated files to the originals (a take some samples), and I found no anomalies.

@ewoutkramer : shall we add _GeneratorLog.cs to Git ignore?

PR #1477, opened 2020-09-09
Fix/1476 null pointer import child

### LINKED ISSUE #1476: NullPointerException in ElementNode.importChild
When trying to import a child which had a parent before and the child is on the first level below the root, a null pointer exception is thrown because ElementNode attempts to remove the wrong parent.

### LINKED ISSUE #1496: Adapt IAsyncResolver interface to use Result<T>
This is actually an old friend, related to #749.

Currently, Resolvers return null if *anything* goes wrong OR if the url cannot be resolved.  But as a called, we have no idea why.  Maybe the resolver had some internal inconsistency (i.e. two resources with the same canonical), maybe the artifact could not be retrieved from a store - in all cases the result of the resolve functions would be `null`.

It would be much better to have the resolve functions return `Result<T>`, so an explicit error can be returned instead. Note: throwing exceptions would not be appropriate, since having resolution failures is not really exceptional.

In any case, we did not have time to fix this in the 2.0 timeframe - it would also really break existing client code, so we need to add this as additional methods to the interface.  That's also a breaking change, but at least this puts the burden for migrating to the authors of the resolvers, not the clients.

alexzautke:
Fixes #1476

### DIFF
--- a/src/Hl7.Fhir.ElementModel.Tests/ElementNodeTests.cs
+++ b/src/Hl7.Fhir.ElementModel.Tests/ElementNodeTests.cs
@@ -292,5 +292,22 @@ namespace Hl7.FhirPath.Tests
             {
             }
         }        
+
+        [TestMethod]
+        public void TestImportChild()
+        {
+            var humanname = ElementNode.Root(provider, "HumanName", "name");
+            humanname.Add(provider, "given", "given");
+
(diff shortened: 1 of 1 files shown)


ewoutkramer:
@marcovisserFurore  - should we include this in b2?

marcovisserFurore (approved):

PR #1480, opened 2020-09-11
STU3: cleaning up solution

marcovisserFurore:
- updated the download script for the downloading the FHIR specification
- .xsd and .sch files are updated to version 3.0.1


src/Hl7.Fhir.Serialization.Tests/TestData/json-edge-cases.json
```diff
@@ -1,201 +1,168 @@
 {
```
ewoutkramer, line 1: Do you have an idea why this file has changed so much?  Is it just whitespace (or I seem some encoding of &nbsp; to 0x202f?

src/Hl7.Fhir.Specification/data/account.xsd
```diff
@@ -27,7 +27,7 @@
   POSSIBILITY OF SUCH DAMAGE.
   
 
-  Generated on Tue, Mar 21, 2017 06:09+1100 for FHIR v3.0.0 
+  Generated on Wed, Apr 19, 2017 07:44+1000 for FHIR v3.0.1
```
ewoutkramer, line 30: Ah we shipped the pre-technical correction XSDs....no one ever complained ;-)

ewoutkramer (approved):

src/Hl7.Fhir.Specification/data/account.xsd
marcovisserFurore: Yes, indeed. I also noticed this, when I was doing the downloads again

src/Hl7.Fhir.Serialization.Tests/TestData/json-edge-cases.json
marcovisserFurore: It is mostly white space. For example:
it was:
```
"identifier": [
        {
          "system": "urn:ietf:rfc:3986",
          "value": "2.16.840.1.113883.19.5"
        }
      ],
```
and now it is:
```
"identifier": [{
                "system": "urn:ietf:rfc:3986",
                "value": "2.16.840.1.113883.19.5"
            }],
```

But during the download, something went wrong, I see now. The family name of the contact should be `du Marché`, but is now `du MarchÃ©`. So I think the encoding is not right in the download phase. I will have a look at it.

PR #1482, opened 2020-09-11
Feature/Improve classmapping initialization (R3)

ewoutkramer:
Improvements to ClassMapping that we can undertake in the 2.0 timeframe.

* Only reflect properties for classes lazily
* Implement suggestions done by Brian P after trying out beta1.


src/Hl7.Fhir.Core.Tests/Introspection/FhirClassMappingTests.cs
```diff
@@ -34,5 +35,34 @@ bool isNested(Type testee)
...
+            var typesToTest = ModelInfo.FhirCsTypeToString.Keys;
+            Console.WriteLine($"Creating a mapping for all {typesToTest.Count} pocos.");
+
+            var sw = new Stopwatch();
+            sw.Start();
+            for (int i = 0; i < 10; i++)
+                foreach (var testee in typesToTest)
+                    createMapping(testee);
+            sw.Stop();
+            Console.WriteLine($"No props: {sw.ElapsedMilliseconds/10.0}ms");
```
marcovisserFurore, line 52: Why are you dividing it by 10? It is already milliseconds, right?

src/Hl7.Fhir.Core.Tests/Introspection/FhirClassMappingTests.cs
```diff
@@ -34,5 +35,34 @@ bool isNested(Type testee)
...
+
+            sw.Restart();
+            for (int i = 0; i < 10; i++)
+                foreach (var testee in typesToTest)
+                    createMapping(testee, touchProps: true);
+            sw.Stop();
+            Console.WriteLine($"With props: {sw.ElapsedMilliseconds/10.0}ms");
```
marcovisserFurore, line 59: Same here: dividing is not necessary, I think

marcovisserFurore (changes requested):
See my notes. 
Thanks for getting the warnings out of the way (comments at Age, etc.)

src/Hl7.Fhir.Core.Tests/Introspection/FhirClassMappingTests.cs
ewoutkramer: The test is run 10 times (see one line before)

src/Hl7.Fhir.Core.Tests/Introspection/FhirClassMappingTests.cs
```diff
@@ -43,20 +43,22 @@ public void TestPerformanceOfMappingAllClasses()
...
+
             var sw = new Stopwatch();
             sw.Start();
-            for (int i = 0; i < 10; i++)
+            for (int i = 0; i < numRepeats; i++)
                 foreach (var testee in typesToTest)
                     createMapping(testee);
             sw.Stop();
-            Console.WriteLine($"No props: {sw.ElapsedMilliseconds/10.0}ms");
+            Console.WriteLine($"No props: {sw.ElapsedMilliseconds/(float)numRepeats}ms");
```
marcovisserFurore, line 54: Thanks for this addition. I overlooked that this part was in a loop

marcovisserFurore (approved):

PR #1485, opened 2020-09-14
Let ClassMapping implement IStructureDefinitionSummary (R3)

ewoutkramer:
There is no real reason to have another class (PocoStructureDefinitionSummaryProvider) do this, as it is easy for ClassMapping to implement this interface. This means we no longer need to (non-common) PocoStructureDefinitionSummaryProvider, though we should probably keep it around for backwards compat.


src/Hl7.Fhir.Core.Tests/Introspection/FhirClassMappingTests.cs
```diff
@@ -34,5 +35,34 @@ bool isNested(Type testee)
...
+        [TestMethod]
+        public void TestPerformanceOfMappingAllClasses()
+        {
+            // just a random list of POCO types available in common
+            var typesToTest = ModelInfo.FhirCsTypeToString.Keys;
+            Console.WriteLine($"Creating a mapping for all {typesToTest.Count} pocos.");
+
+            var sw = new Stopwatch();
+            sw.Start();
+            for (int i = 0; i < 10; i++)
```
marcovisserFurore, line 48: This was in a previous PR changed to `for (int i = 0; i < numRepeats; i++)`. Now it changed back to `< 10`.

marcovisserFurore (changes requested):
Some of the changes in this PR are from another PR which has been merged already into develop-stu3. Why we see then still the differences? Do these changes not override the other PR?

src/Hl7.Fhir.Core.Tests/Introspection/FhirClassMappingTests.cs
ewoutkramer: This PR was based against dstu-3, not against the PR where I corrected it.

marcovisserFurore (approved):

marcovisserFurore:
The PR at Common (FirelyTeam/fhir-net-common/pull/78) has still status _Change Requested_

Issue #1237, opened 2020-02-03
R5: Fix interfaces in snapshot generation


mmsmits:
Fix snapshot generator to handle interface artifacts

https://chat.fhir.org/#narrow/stream/179177-conformance/topic/Element.2Ebase.20on.20elements.20implemented.20from.20interface.20classes

marcovisserFurore:
This should be solved as well for 2.0 r5

PR #1487, opened 2020-09-15
Refactor R5: use of extension structuredefinition-interface

marcovisserFurore:
We use now the `extension structuredefinition-interface` to identify the pattern/interface classes `CanonicalResource` and `MetadataResource`

### DIFF
--- a/src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
+++ b/src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
@@ -2651,7 +2651,7 @@ namespace Hl7.Fhir.Specification.Tests
             if (!result)
             {
                 Debug.Print("Expanded is not exactly equal to original... verifying...");
-                result = verifyElementBase(sd, expanded);
+                result = await verifyElementBase(sd, expanded);
             }
 
             // Core artifact snapshots are incorrect, e.g. url snapshot is missing extension element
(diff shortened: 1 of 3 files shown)


marcovisserFurore:
Solves #1237

src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
```diff
@@ -2800,9 +2800,9 @@ bool verifyElementBase(StructureDefinition original, StructureDefinition expande
                 else if (expanded.Kind == StructureDefinition.StructureDefinitionKind.Resource)
                 {
                     // [WMR 20190131] Fixed
-                    var baseDef = SnapshotGenerator.getBaseDefinition(expanded);
-                    bool isDerivedFromResource = baseDef == ModelInfo.CanonicalUriForFhirCoreType(FHIRAllTypes.Resource);
-                    bool isDerivedFromDomainResource = baseDef == ModelInfo.CanonicalUriForFhirCoreType(FHIRAllTypes.DomainResource);
+                    var baseDef = await _generator.getBaseDefinition(expanded);
```
ewoutkramer, line 2803: What's the difference between `SnapshotGenerator` and `_generator`? Is that the same instance?

ewoutkramer (approved):

src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
marcovisserFurore: `getBaseDefinition` was earlier in this PR a static method. I changed it to an instance method.

PR #1492, opened 2020-09-16
Let ClassMapping implement IStructureDefinitionSummary (R4)

ewoutkramer:
There is no real reason to have another class (PocoStructureDefinitionSummaryProvider) do this, as it is easy for ClassMapping to implement this interface. This means we no longer need to (non-common) PocoStructureDefinitionSummaryProvider, though we should probably keep it around for backwards compat.


marcovisserFurore (approved):
PR at Common (FirelyTeam/fhir-net-common/pull/78) has still the status _Request changes_

PR #1505, opened 2020-09-21
Add string encoding funcs + calendar months to FhirPath (R4)

ewoutkramer:
The latest sets of tests published by HL7 for FhirPath include changes to the way calendar months are treated (to be more precise: it no longer uses UCUM annotations to encode calendar durations as UCUM codes).

Also, a set of experimental string-encoding functions have been added that will probably become part of the next FhirPath normative release.


marcovisserFurore:
The `SearchExtractionAllExamples` is failing because:
```
Error processing file observation-example-sample-data(ekg).xml: Value cannot be null.
Parameter name: unit
```

This file is coming out `examples.zip` (R4)

PR #1509, opened 2020-09-24
Move parsers and serializers to common (R3)

ewoutkramer:
By removing the dependency on ModelInfo, I have been able to move classes for parsing and serializing (at least a major part) to the common assemblies. This is the low-hanging fruit, I'll see if there's more to move later to not make this PR overly big and complex.


src/Hl7.Fhir.Core/Model/ModelInfo.cs
```diff
@@ -473,6 +474,24 @@ public static string CanonicalUriForFhirCoreType(FHIRAllTypes type)
...
+        // The lazy here is not used for purposes of performance (delaying creation of the versioned provider, which
+        // is cheap anyway), but to use the framework to have a nicely multithread-aware singleton.
+        private static readonly Lazy<ModelInspector> _versionedProvider
+            = new Lazy<ModelInspector>(createVersionedProvider);
+
+        private static ModelInspector createVersionedProvider()
+        {
+            var result = new ModelInspector(Specification.FhirRelease.STU3);
+
+            result.Import(typeof(Resource).GetTypeInfo().Assembly);
```
marcovisserFurore, line 486: In fact you want to import from `Hl7.Fhir.Support.Poco` and `Hl7.Fhir.Core`. Should we not be more specific about this then? When Patient moves to Common, this code will not work for example..

marcovisserFurore (commented):
See my comment about loading assemblies below

src/Hl7.Fhir.Core/Model/ModelInfo.cs
ewoutkramer: Yes......using ModelInfo is obviously better.

marcovisserFurore (approved):

PR #1513, opened 2020-10-07
Validation test environment

mmsmits:


src/Hl7.Fhir.Specification.Tests/Validation/ValidationManifestTest.cs
```diff
@@ -0,0 +1,256 @@
...
+                                                    GenerateSnapshotForExternalProfiles = true,
+                                                    GenerateElementIds = true
+                                                };
+            settings.ResourceResolver = testResolver;
+
+           _testValidator = new Validator(settings);
+        }
+
+        [Ignore]
+        [TestMethod]
```
marcovisserFurore, line 45: Attribute `TestMethod` can be removed, because the `DataTestMethod` is already there

src/Hl7.Fhir.Specification.Tests/Validation/ValidationManifestTest.cs
```diff
@@ -0,0 +1,256 @@
...
+        {
+            RunTestCase(testCase);
+        } 
+
+        public static void RunTestCase(ValidationTestCase testCase)
+        {
+            Resource testResource = null;
+            if (testCase.FileName.EndsWith(".xml"))
+            {
+                var reader = XmlReader.Create(@"TestData\\validation-test-suite\\" + testCase.FileName);
```
marcovisserFurore, line 58: When you use `@` you don't have to escape `\` anymore. So this is the same:
```
var reader = XmlReader.Create(@"TestData\validation-test-suite\" + testCase.FileName);
```

src/Hl7.Fhir.Specification.Tests/Validation/ValidationManifestTest.cs
```diff
@@ -0,0 +1,256 @@
...
+            else if (testCase.FileName.EndsWith(".json"))
+            {
+                var json = File.ReadAllText(@"TestData\\validation-test-suite\\" + testCase.FileName);
+                testResource = new FhirJsonParser().Parse<Resource>(json);
+            }
+
+            Assert.IsNotNull(testResource);
+
+            var profileFilePath = testCase?.Profiles?.FirstOrDefault() ?? testCase?.ValidationProfile?.Source;
+            var profileUri = _dirSource.ListSummaries().Where(s => s.Origin.EndsWith(Path.DirectorySeparatorChar + profileFilePath)).Select(s => s.GetConformanceCanonicalUrl()).FirstOrDefault();
```
marcovisserFurore, line 70: Multiple profiles could exist, so why only select the first?

src/Hl7.Fhir.Specification.Tests/Validation/ValidationManifestTest.cs
marcovisserFurore, line 58: You could use also a textreader (which you do at Json as well), then you can write it shorter:

 ```c#
            var resourceText = File.ReadAllText(@$"TestData\validation-test-suite\{testCase.FileName}");
            var testResource = testCase.FileName.EndsWith(".xml") ?
                new FhirXmlParser().Parse<Resource>(resourceText) :
                new FhirJsonParser().Parse<Resource>(resourceText);
```

marcovisserFurore (changes requested):
Looks great. Too bad about the 600 (!!) extra files. But maybe we can solve this in the feature with a submodule to the repo of Grahame. For now it is fine.

Futhermore, I have some suggestions in your code:

marcovisserFurore (approved):
Giddy up!

PR #1567, opened 2020-12-14
STU3: fix: remove type specific constraints from polymorph elements

### LINKED ISSUE #1540: [SnapshotGenerator] Snapshot incorrectly adds rat-1 constraint to polymorphic element
The snapshot generated by Simplifier, Vonk and Torinox incorrectly add the constraint to the polymorphic element. 

The unit test that proves the bug: 

```
[Fact]
        public void ValidateMedicationAdministration()
        {
            var snapshotSettings = new SnapshotGeneratorSettings
            {
                GenerateSnapshotForExternalProfiles = true,
                ForceRegenerateSnapshots = true,
                GenerateAnnotationsOnConstraints = false,
                GenerateElementIds = false,
                GenerateExtensionsOnConstraints = false,
            };
            var source = new SnapshotSource(ZipSource.CreateValidationSource(), snapshotSettings);
            var settings = ValidationSettings.CreateDefault();
            settings.ResourceResolver = source;
            settings.GenerateSnapshot = true;
            settings.ResolveExternalReferences = true;
            settings.GenerateSnapshotSettings = snapshotSettings;
            var maXML = File.ReadAllText(Path.Combine("TestData", "validation", "MedicationAdministration-example.xml"));
            var node = FhirXmlNode.Parse(maXML);
            var ma = node.ToTypedElement(new StructureDefinitionSummaryProvider(source));
            var validator = new Validator(settings);
            var report = validator.Validate(ma);
            Assert.Equal(0, report.Errors);
        }
```

The snaphot snippet that shows the incorrectly added constraint:

```
<element id="MedicationAdministration.dosage.rate[x]">
      <path value="MedicationAdministration.dosage.rate[x]"/>
      <short value="Dose quantity per unit of time"/>
      <definition value="Identifies the speed with which the medication was or will be introduced into the patient.  Typically the rate for an infusion e.g. 100 ml per 1 hour or 100 ml/hr.  May also be expressed as a rate per unit of time e.g. 500 ml per 2 hours.  Other examples:  200 mcg/min or 200 mcg/1 minute; 1 liter/8 hours."/>
      <comment value="If the rate changes over time, and you want to capture this in MedicationAdministration, then each change should be captured as a distinct MedicationAdministration, with a specific MedicationAdministration.dosage.rate, and the date time when the rate change occurred. Typically, the MedicationAdministration.dosage.rate element is not used to convey an average rate."/>
      <min value="0"/>
      <max value="1"/>
      <type>
        <code value="Ratio"/>
      </type>
      <type>
        <code value="Quantity"/>
        <profile value="http://hl7.org/fhir/StructureDefinition/SimpleQuantity"/>
      </type>
      <condition value="ele-1"/>
      <constraint>
        <key value="ele-1"/>
        <severity value="error"/>
        <human value="All FHIR elements must have a @value or children"/>
        <expression value="hasValue() | (children().count() &gt; id.count())"/>
        <xpath value="@value|f:*|h:div"/>
      </constraint>
      <constraint>
        <key value="rat-1"/>
        <severity value="error"/>
        <human value="Numerator and denominator SHALL both be present, or both are absent. If both are absent, there SHALL be some extension present"/>
        <expression value="(numerator.empty() xor denominator.exists()) and (numerator.exists() or extension.exists())"/>
        <xpath value="(count(f:numerator) = count(f:denominator)) and ((count(f:numerator) &gt; 0) or (count(f:extension) &gt; 0))"/>
      </constraint>
      <mapping>
        <identity value="rim"/>
        <map value="n/a"/>
      </mapping>
      <mapping>
        <identity value="v2"/>
        <map value="N/A"/>
      </mapping>
      <mapping>
        <identity value="rim"/>
        <map value="RTO"/>
      </mapping>
      <mapping>
        <identity value="v2"/>
        <map value="RXA-12-Administered Per (Time Unit)"/>
      </mapping>
      <mapping>
        <identity value="rim"/>
        <map value=".rateQuantity"/>
      </mapping>
```

Additionally, .ForceRegenerateSnapshot in the .NET validator does not force the regeneration of the snapshot.

mmsmits:
fixes #1540

### DIFF
--- a/src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
+++ b/src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
@@ -6987,11 +6987,10 @@ namespace Hl7.Fhir.Specification.Tests
         public async T.Task TestInvariantsOnValueX()
         {
             var sd = await _testResolver.FindStructureDefinitionAsync("http://hl7.org/fhir/StructureDefinition/MedicationAdministration");
-            var generator = _generator = new SnapshotGenerator(_testResolver, _settings);
-
+          
             (_, var expanded) = await generateSnapshotAndCompare(sd);
 
-            dumpOutcome(generator.Outcome);
+            dumpOutcome(_generator.Outcome);
             dumpBaseElems(expanded.Snapshot.Element);
             Assert.IsNotNull(expanded);
             Assert.IsTrue(expanded.HasSnapshot);
--- a/src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
+++ b/src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
@@ -6899,6 +6899,29 @@ namespace Hl7.Fhir.Specification.Tests
         }
 
         [TestMethod]
+        public async T.Task TestInvariantsOnValueX()
+        {
+            var sd = await _testResolver.FindStructureDefinitionAsync("http://hl7.org/fhir/StructureDefinition/MedicationAdministration");
+            var generator = _generator = new SnapshotGenerator(_testResolver, _settings);
+
+            (_, var expanded) = await generateSnapshotAndCompare(sd);
+
+            dumpOutcome(generator.Outcome);
+            dumpBaseElems(expanded.Snapshot.Element);
+            Assert.IsNotNull(expanded);
+            Assert.IsTrue(expanded.HasSnapshot);
+
+            var nav = ElementDefinitionNavigator.ForSnapshot(expanded);
+            Assert.IsTrue(nav.JumpToFirst("MedicationAdministration.dosage.rate[x]"));
+            Assert.IsNotNull(nav.Current);
+
+            //verify that rate[x] contains ele-1 but not rat-1
+            Assert.IsTrue(nav.Current.Constraint.Any(c=> c.Key == "ele-1"));
+            Assert.IsFalse(nav.Current.Constraint.Any(c => c.Key == "rat-1"));
+
+        }
+
+        [TestMethod]
         public async T.Task TestReferenceTargetProfile()
         {
             // Verify that the snapshot generator correctly expands elements with a targetProfile (on ResourceReference itself)
@@ -7247,5 +7270,7 @@ namespace Hl7.Fhir.Specification.Tests
             newSliceSystem.Should().NotBeNull("The new slice 'newSlice' should be present in the snapshot");
             newSliceSystem.Fixed.Should().BeNull("No constraint elements from the base slice (BSN) should be present");
         }
+
+        
     }
 }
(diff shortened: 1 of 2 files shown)


src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
```diff
@@ -757,6 +763,29 @@ private async T.Task createNewElement(ElementDefinitionNavigator snap, ElementDe
...
+                if (nav.MoveToFirstChild())
+                {
+                    if(nav.Current.IsRootElement()) 
+                        newConstraints = nav.Current.Constraint;
+                }
+                //If there are any newly introduced constraints, remove them from the new element.
+                if (newConstraints != null)
+                {
+                    var keys = newConstraints.Select(c => c.Key);
+                    var removedConstraints = element.Constraint.Where(c => keys.Contains(c.Key))?.ToList();
```
marcovisserFurore, line 783: Another way of doing this, is:
```c#
element.Constraint.RemoveAll(c => keys.Contains(c.Key));
```

Although with big lists you should be carefull, but that is here not the case: more than 1000 constraints would be not realistic. When you deal with large list you could also do this:
```c#
var keys  = new HashSet<string>(newConstraints.Select(c => c.Key));
element.Constraint.RemoveAll(c => removedConstraints.Contains(c.Key));
```
That makes the complexity O(n+m) instead of O(n*m).

marcovisserFurore (commented):
Your unit test shows that the issue is resolved, which is great. Although I am still wondering why this is happening. For example, why is the constraint of `Quantity` (`qty-3`) not copied, but the constraint of `Ratio` (`rat-1`) is? And is this happening with other choice types as well? For example `value[x]` of `Observation`

mmsmits:
> 
> 
> Your unit test shows that the issue is resolved, which is great. Although I am still wondering why this is happening. For example, why is the constraint of `Quantity` (`qty-3`) not copied, but the constraint of `Ratio` (`rat-1`) is? And is this happening with other choice types as well? For example `value[x]` of `Observation`

Because when the snapshotgenerator creates a new element, it first creates a new type based on the first types in the typelist of the diff. After that, it merges in the other types. With this creation it copies over all constraints, which it doesn't do with merging. 
That's why qty-3 isn't copied.

src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
mmsmits: Thanks, I'll fix that!

src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
```diff
@@ -6983,6 +6983,28 @@ public async T.Task TestExtensionOnPrimitive()
             Assert.IsNotNull(expanded.Snapshot.Element.Where(e => e.Path == "Address.use")?.FirstOrDefault()?.DefinitionElement?.Value);
         }
 
+        [TestMethod]
+        public async T.Task TestInvariantsOnValueX()
+        {
+            var sd = await _testResolver.FindStructureDefinitionAsync("http://hl7.org/fhir/StructureDefinition/MedicationAdministration");
+            var generator = _generator = new SnapshotGenerator(_testResolver, _settings);
```
marcovisserFurore, line 6990: This is not needed, right?

mmsmits: Nope, don't know what happened there..

marcovisserFurore (approved):

PR #1607, opened 2021-01-26
Fix/1605 quantity different units

### LINKED ISSUE #1605: Cannot select quantities with different units within a single FHIRPath select
**Describe the bug**
Executing a FHIRPath select statement on an Observation resource containing a quantity in Observation.valueQuantity and Observation.component.valueQuantity with different units results in the following exception:

System.NotSupportedException: Comparing quantities with different units is not yet supported

These values can be selected for example by the following FHIRPath statement: "(Observation.value as Quantity) | (Observation.component.value as Quantity)". This selection is for example done in the SearchParameter "combo-value-quantity" defined on Observation.

The stack trace indicates that some part of the FHIRPath stack is triggering an equal comparison by adding the second value to the result set. 

**To Reproduce**
See unit test "TestSelectDifferentQuantities" in Hl7.FhirPath.R4.Tests.FhirPathTest (Branch: fix/1605-quantity-different-units)

**Expected behavior**
A result set containing both quantities.

**Version used:**
 - FHIR Version: [R4]
 - Version: [2.0.2]

marcovisserFurore:
fixes #1605

### DIFF
--- a/src/Hl7.FhirPath.R4.Tests/PocoTests/FhirPathTest.cs
+++ b/src/Hl7.FhirPath.R4.Tests/PocoTests/FhirPathTest.cs
@@ -274,5 +274,18 @@ namespace Hl7.FhirPath.R4.Tests
             Assert.IsNotNull(dataType, "Select should result an instance");
             Assert.IsInstanceOfType(dataType, input.GetType(), "dataType should be the same type of the input");
         }
+
+        [TestMethod]
+        public void TestSelectDifferentQuantities()
+        {
+            var obs = new Observation()
+            {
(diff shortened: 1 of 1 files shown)


alexzautke (approved):
Did you intentionally include the OperationsTests.cs in this PR in the common repo? The rest looks good to me.

marcovisserFurore:
I added `OperationsTests.cs` in Common because the bug can be reduced to `4 'g' < 8 '%'`.

alexzautke:
Ah even better!

marcovisserFurore:
Solved in hotfix 2.0.3

PR #1624, opened 2021-02-10
Devops/create parallel jobs

marcovisserFurore:
Run CI with parallel unit test jobs. The unittests run in release mode


src/Hl7.Fhir.Core.Tests/Hl7.Fhir.Core.Tests.csproj
```diff
@@ -1,7 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
   <PropertyGroup>
     <TargetFrameworks>netcoreapp2.1</TargetFrameworks>
-    <TargetFrameworks Condition=" '$(Configuration)' != 'Debug' ">$(TargetFrameworks);net45</TargetFrameworks>
```
ewoutkramer, line 4: Why don't we test on Framework 4.5 anymore?  I'm sure to save time?

ewoutkramer (changes requested):

src/Hl7.Fhir.Core.Tests/Hl7.Fhir.Core.Tests.csproj
marcovisserFurore: This has never worked before: in the solution were the unit testprojects deselected for the build in Release mode. So the unit tests never executed in Release mode.
Adding `net45` to the targetframework for the test projects in Release mode will not compile at the moment.

ewoutkramer: Ah right. I remembered that a committer had helped us to set up tests for all the platforms, but this must have just been in Debug mode (mostly because of all the internals that become visible in debug alone)

ewoutkramer (approved):

PR #1632, opened 2021-02-16
ValidatorManifest: Use of System.Text.Json with Poco objects

marcovisserFurore:
- Use of System.Text.Json with Poco objects


ewoutkramer:
Shall we close this PR?

Issue #1659, opened 2021-03-18
Accessing .AbsoluteUri Property on URI with UriKind.Relative


marcovisserFurore:
**Issue original from @thomas-eggemann**

I'm currently running into issues when processing responses from a HAPI FHIR Server.

As per the [Hl7 HTTP spec](https://www.hl7.org/fhir/http.html#create) location headers MAY be absolute or relative.

`[...] where [id] and [vid] are the newly created id and version id for the resource version. The Location header should be as specific as possible - if the server understands versioning, the version is included. If a server does not track versions, the Location header will just contain [base]/[type]/[id]. The Location MAY be an absolute or relative URL. `

Line 33 in [HttpToEntryExtension](https://github.com/FirelyTeam/firely-net-common/blob/develop/src/Hl7.Fhir.Support.Poco/Rest/HttpToEntryExtensions.cs) is reading the location- and contentlocation-headers by using the `.AbsoluteUri` property.

https://github.com/FirelyTeam/firely-net-common/blob/72029833777b73d1fa3d87de783504eaf18e62b8/src/Hl7.Fhir.Support.Poco/Rest/HttpToEntryExtensions.cs#L33

The server CAN return a relative URI e.g.: `/fhir/*/Bundle/{some_identifier_here}`
When a relative URI is returned in the location-/contentlocation-header the call to `.AbsoluteUri` causes an `InvalidOperationException` => _This operation is not supported for a relative URI._

Wouldn't it be sensible to allow relative URIs by replacing the access via `.AbsoluteUri` property with a call to `.ToString()` or even `.OriginalString`?

Suggested solution:
`Location = response.Headers.Location?.ToString() ?? response.Content.Headers.ContentLocation?.ToString(),`

Or maybe i'm getting something terribly wrong here.
If so, could someone explain to me why?

mmsmits:
Thanks for you thorough issue @thomas-eggemann. I investigated a bit and it seems like you are correct that we should change that.

PR #1662, opened 2021-03-22
Support relative URLs when parsing HttpResponseMessage to EntryResponse

mmsmits:
fixes: #1659


marcovisserFurore (approved):
Will be approved, when Common is approved

marcovisserFurore (changes requested):
Will be approved when corresponding Common is approved

marcovisserFurore (approved):

PR #1671, opened 2021-03-30
Can now use a discriminator to walk into a complex extension (for R3)

### LINKED ISSUE #1656: ElementDefinition.slicing.discriminator.path for complex Extensions
Use case: we want to add multiple, similar, complex extensions to a profile. They are differentiated by a fixed or pattern coding in one of the complex extension, in this case the `type` extension in ( http://hl7.org/fhir/extension-patient-proficiency.html).

We can't get a correct discriminator.path that gets into the complex extension. We tried: `Patient.communication.extension('http://hl7.org/fhir/StructureDefinition/patient-proficiency').extension('type').value`

- The .NET validator is complaining about that expression: Internal logic failure: Cannot walk into unknown StructureDefinition with canonical 'type' at 'http://hl7.org/fhir/StructureDefinition/patient-proficiency#Extension (Extension)'
- The JAVA validator also gives an error. But with not anything helpful.
- Zulip: https://chat.fhir.org/#narrow/stream/179166-implementers/topic/How.20to.20discriminate.20on.20a.20value.20in.20a.20complex.20extension.3F

Added some example R4 data. It has one example instance with the different extensions and the related (draft and unfinished) profiles.

[example data.zip](https://github.com/FirelyTeam/firely-net-sdk/files/6129392/example.data.zip)

ewoutkramer:
Fixed a bug where a discriminator walking into a complex extension would not find the complex child extension.

Created a (valid) example of a profile that constrains a complex extension inline and fixed the walker to look for fixUri instead of a targetProfile.profile.

## Related issues
Fixes #1656.

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Specification.Tests/Hl7.Fhir.Specification.Tests.csproj
+++ b/src/Hl7.Fhir.Specification.Tests/Hl7.Fhir.Specification.Tests.csproj
@@ -43,13 +43,7 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
   </ItemGroup>  
-
-  <ItemGroup>
-    <None Remove="TestData\validation\list-with-profile-slicing.xml" />
-    <None Remove="TestData\validation\observation-profile-for-discriminator-test.xml" />
-    <None Remove="TestData\validation\patient-sliced-complex-extension.xml" />
-  </ItemGroup>
-  
+ 
   <ItemGroup>
     <Content Include="..\Hl7.Fhir.Specification\data\profiles-resources.xml" Link="TestData\snapshot-test\profiles-resources.xml">
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
--- a/src/Hl7.Fhir.Specification.Tests/Hl7.Fhir.Specification.Tests.csproj
+++ b/src/Hl7.Fhir.Specification.Tests/Hl7.Fhir.Specification.Tests.csproj
@@ -47,6 +47,7 @@
   <ItemGroup>
     <None Remove="TestData\validation\list-with-profile-slicing.xml" />
     <None Remove="TestData\validation\observation-profile-for-discriminator-test.xml" />
+    <None Remove="TestData\validation\patient-sliced-complex-extension.xml" />
   </ItemGroup>
   
   <ItemGroup>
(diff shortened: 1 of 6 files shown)


src/Hl7.Fhir.Specification.Tests/Hl7.Fhir.Specification.Tests.csproj
```diff
@@ -47,6 +47,7 @@
   <ItemGroup>
```
marcovisserFurore, line 47: This ItemGroup is unnecessary, because it is already covered by the previous ItemGroup (with wildcards). So I would remove the whol ItemGroup here.

src/Hl7.Fhir.Specification/Specification/Navigation/StructureDefinitionWalker.cs
```diff
@@ -233,20 +233,35 @@ public IEnumerable<StructureDefinitionWalker> Resolve()
...
-                return FromCanonical(url);
-            else if (selection.Count == 1)
-                return new StructureDefinitionWalker(selection.Single(), _resolver);
-            else
-                throw new StructureDefinitionWalkerException($"extension('{url}') found multiple extension slices constraining the same extension, with is not allowed for the discriminator at '{Current.CanonicalPath()}'.");
+            return selection switch
+            {
+                // No children in the current profile -> we can still continue in the definition
+                // of the extension itself.
+                { Count: 0 } when isAbsoluteUri(url) => FromCanonical(url),
```
marcovisserFurore, line 243: wow, nice condition :)

marcovisserFurore (changes requested):
Small (non functional) remark:

src/Hl7.Fhir.Specification/Specification/Navigation/StructureDefinitionWalker.cs
ewoutkramer: Who needs if/else blocks anymore? Let's hope in C#10 they will allow blocks after the '=>' because, that's the only painful reason to use if/else if/else's now.

src/Hl7.Fhir.Specification.Tests/Hl7.Fhir.Specification.Tests.csproj
ewoutkramer: Ah, yes, I've seen VS do that before. I remove them from time to time manually...

marcovisserFurore (approved):

PR #1672, opened 2021-03-30
ScopedNode: AtResource can also be determined from IResourceTypeSupplier

marcovisserFurore:
## Description
In class `ScopedNode` the property `AtResource` can now also be determined from IResourceTypeSupplier when there is no `Definition` present.

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.ElementModel.Tests/ScopedNodeTests.cs
+++ b/src/Hl7.Fhir.ElementModel.Tests/ScopedNodeTests.cs
@@ -8,15 +8,16 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Text;
-using T=System.Threading.Tasks;
+using T = System.Threading.Tasks;
+
+#nullable enable
 
 namespace Hl7.Fhir.ElementModel.Tests
 {
     [TestClass]
     public class ScopedNodeTests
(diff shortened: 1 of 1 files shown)


src/Hl7.Fhir.ElementModel.Tests/ScopedNodeTests.cs
```diff
@@ -257,5 +271,59 @@ public async T.Task<Resource> ResolveByCanonicalUriAsync(string uri)
             public T.Task<Resource> ResolveByUriAsync(string uri) => throw new NotImplementedException();
         }
 
+        private class ElementNodeAdapter : ITypedElement, IResourceTypeSupplier
```
ewoutkramer, line 274: Was een wrapper rond een normale `ITypedElement` die `null` teruggeeft bij `Definition` niet goed genoeg?

ewoutkramer: Translation: wouldn't a wrapper around a normal `ITypedElement` returning `null` for `Definition` be sufficient?

ewoutkramer (changes requested):
I need to do this to progress. See my previous comment.

ewoutkramer (approved):

PR #1683, opened 2021-04-07
Reorganize slicing helper functions (for R3)

ewoutkramer:
## Description
For the new validation compiler, I have rewritten (read: simplified) the code to detect (re)slices. I have moved this new code to this repo since it may be of use to others. There was some comparable (older) code that I have now either replaced (FindMemberSlices) or kept intact (variants of IsResliceOf) - to avoid introducing regressions.

Also, I have fixed some pre-existing compiler warnings.

## FirelyTeam Checklist


src/Hl7.Fhir.Specification/Specification/Navigation/ElementNavigatorSlicingExtensions.cs
```diff
@@ -169,6 +170,7 @@ public static bool MoveToFirstReslice(this ElementDefinitionNavigator nav)
             return false;
         }
 
+#if CAN_THIS_OLD_CODE_BE_REMOVED
```
marcovisserFurore, line 173: I am not really a fan of this. Can we not make this internal function `FindMemberSlices(ElementDefinitionNavigator, bool)` deprecated? It is only used by `BucketFactory` and there you already use the new function `FindMemberSlices(ElementDefinitionNavigator)`

marcovisserFurore (changes requested):
Maybe deprecate the function instead of #ifdef

src/Hl7.Fhir.Specification/Specification/Navigation/ElementNavigatorSlicingExtensions.cs
ewoutkramer: Ok, I'll be brave and just remove it (it's internal after all).

marcovisserFurore (approved):

Issue #1681, opened 2021-04-05
Handle OperationOutCome resource in response


jiteshmadhu:
Hello,

We are using firely nuget  package  version 4  in our .Net project to push data through as FHIR standard to the server.

Now, our problem is server is always giving response in OperationOutCome resource in response jSON body. It gives same response (OperationOutcome) in all the cases means In case of status is OK and in case of status is Error.

So, how can we handle that using firely framework?

We have done following code to push the patient data to the server.
..... 
var client = new FhirClient("server end point url");
var pat = new Patient();
..... 
..... 
var Objres =  client.Create(pat):
.....

As per above code we are always receving NULL value in the Objres because server is giving response body as OperationOutCome and firely framework expecting response resource should be the same resource which was pushed.

So, how can we get OperationOutCome response body jSON?

Thanks in advance!
Jitesh

marcovisserFurore:
Hi Jitesh, thanks for using our Firely .NET SDK.

According to the FHIR standard, a FHIR server SHOULD always return the resource, see also http://hl7.org/fhir/http.html#create.

But when something goes wrong, for example the resource is not valid, then the Create method will throw an exception. In that exception you can see the OperationOutcome from the server. See here a snippet how to do that:

```c#
            using var client = new FhirClient(testEndpoint);
            try
            {
                var result = client.Create<Patient>(new Patient());
                Assert.IsNotNull(result);
            }
            catch (FhirOperationException ex)
            {
                if (!ex.Outcome.Success)
                {
                    // do something
                }
            }
```

brianpos:
The server can return the OperationOutcome like that, and can even be requested using the prefer header
```
Prefer: return=OperationOutcome
```
http://hl7.org/fhir/http.html#ops
From the C# code you can do this like this:
``` c#
            Hl7.Fhir.Rest.FhirClient clientFhir = new Hl7.Fhir.Rest.FhirClient(_baseAddress, false);
            clientFhir.PreferredReturn = Prefer.OperationOutcome;
```
I'm checking what the `client.Create<Patient>(pat)` would do with this... But it's perfectly valid to do it.

marcovisserFurore:
@mmsmits: can you check the last comment of Brian (update 2)?

brianpos:
Thanks, I would expect the outcome in the outcome property, and the header values put into the other props, Id locarion, etag etc.

mmsmits:
[...]

PR #1690, opened 2021-04-16
FhirClient: handle prefer operation outcome in resource response

mmsmits:
fixes: #1681


src/Hl7.Fhir.Core.Tests/Rest/FhirClientTests.cs
```diff
@@ -87,7 +87,7 @@ private static void CreateItems()
                         Family = "Donald",
                     }
                 },
-                Id = "pat1",
```
marcovisserFurore, line 90: Maybe a better approach to make the identifiers dependent on the FhirVersion, like for example:

```
"pat1" + ModelInfo.Version
```

Make this a variable and use them in the tests.

marcovisserFurore (changes requested):
Small unittest change

marcovisserFurore (approved):
Very nice !

PR #1700, opened 2021-04-30
Async support Json/Xml Parsers + Serializers

kennethmyhra:
## Description
Async support for Json Parsers related and Json/Xml Serializers

Non-async methods marked as obsolete, documentation refers to the corresponding Async method

Code in some of the smaller non-async methods are currently not modified since there was so little actual code affected

Corresponding PR: https://github.com/FirelyTeam/firely-net-common/pull/112

## Testing
Tested by running the unit tests, except for unit tests marked as IntegrationTests

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/ElementModel/PocoTypedElementTests.cs
+++ b/src/Hl7.Fhir.Core.Tests/ElementModel/PocoTypedElementTests.cs
@@ -9,6 +9,7 @@ using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using Tasks = System.Threading.Tasks;
 
 namespace Hl7.Fhir.Core.Tests.ElementModel
 {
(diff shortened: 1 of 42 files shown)


src/Hl7.Fhir.Specification.Tests/Hl7.Fhir.Specification.Tests.csproj
```diff
@@ -52,7 +52,7 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
     </Content>
     <Content Include="TestData\ResourcesInSubfolder.zip">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
marcovisserFurore, line 55: Are you sure? This can delay the compilation.

marcovisserFurore (changes requested):

src/Hl7.Fhir.Specification.Tests/Hl7.Fhir.Specification.Tests.csproj
kennethmyhra: Oh, that shouldn't have been changed. I'll set it back to PreserveNewest. I had some trouble with tests relying on that file failing so I changed the setting, but it should never have been committed

marcovisserFurore (changes requested):
Minor things

marcovisserFurore (approved):

PR #1701, opened 2021-05-03
Moved another set of classes to Common

ewoutkramer:
Moved some more classes to Common.

This is the relevant PR in Common: https://github.com/FirelyTeam/firely-net-common/pull/113


marcovisserFurore (approved):
Checked with PR https://github.com/FirelyTeam/firely-net-common/pull/113

Issue #1388, opened 2020-06-02
Produce ResourceReference attribute for choice types


darena-mhaque:
We perform additional validation for resource references so that a resource type that is not supposed to be in there doesn't get set. EG: Patient.link.other is a resource reference that can only be a resource of type Patient or RelatedPerson. This means that we shouldn't be able to set the `reference` property to `Organization/123` for example. We are currently performing the validation by obtaining the supported resource types by obtaining the `ReferencesAttribute` on that property using reflection.

The problem occurs when we come across a resource reference that is a choice value. EG: MedicationDispense.medicationReference. This property is a choice property (the other choice being medicationCodeableConcept). This property does not contain the `ReferencesAttribute` in the POCO class, so we can't use our current validation technique.

Is there some other way the library can provide the available resource types that can be set for a resource reference other than using the `ReferencesAttribute`? If not, can we somehow add this as a feature?

ewoutkramer:
It seems we should have generated the `ReferencesAttribute` here too - and the way to do that is to tune the .tt (T4 template files in the Hl7.Fhir.Core/Model directory) files!

Is this something you can do for us?

darena-mhaque:
Apologies for the late reply. Unfortunately, the only time I see that I can take out to look into this would be sometime early August...

ewoutkramer:
No problem - August is early enough for me ;-)

ewoutkramer:
I can see this is a neat little corner case - I had missed that this is specific to elements where `Reference` is one of the choice types. Since june we have actually moved to a new code-generator, and rebuilt it in that new technology. We did that so faithfully 1-on-1 that this is still an issue.  So, don't waste time on the T4 generator anymore, instead we're now using this: https://github.com/microsoft/fhir-codegen

PR #1713, opened 2021-05-06
Produce ResourceReference attributes for choice types [R5]

ewoutkramer:
Re-ran generator after adding support for better [ResourceReference] generation. Solves #1388


ewoutkramer:
Set to "draft" since we cannot generate the Statistic resource correctly (yet).

marcovisserFurore (approved):

PR #1721, opened 2021-05-12
Tests: FhirPath 'as' can return FHIR primtives

### LINKED ISSUE #1643: ITypedElement Select Type question
**Describe the bug**
`ConceptMap` Select `source-uri` return type is `string`?

**To Reproduce**
The `source` property of `Group` under `ConceptMap` is of `URI` type in `StructureDefinition`,`reference` type in `SearchParameter`,but `string` type in the SDK.

image

```csharp
typeElement.Select("ConceptMap.group.source").First().InstanceType; // uri
typeElement.Select("(ConceptMap.source as uri)").First().InstanceType; // System.String
```

kennethmyhra:
## Description
Unit Tests for the fix for the 'as' operator committed in FhirPath
library in the firely-net-common repository
Related PR: https://github.com/FirelyTeam/firely-net-common/pull/117

## Related issues
Fixes #1643.

## Testing
Tested with the unit tests in this PR

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/Model/FhirPathOperationTests.cs
+++ b/src/Hl7.Fhir.Core.Tests/Model/FhirPathOperationTests.cs
@@ -1,0 +1,48 @@
+﻿using Hl7.Fhir.Model;
+using Hl7.Fhir.FhirPath;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Linq;
+
+namespace Hl7.Fhir.Core.Tests.Model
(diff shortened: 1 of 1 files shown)


ewoutkramer:
Thanks @kennethmyhra - I have replaced this PR with #1725.  We keep all (poco-based) tests for FhirPath in the R4 release, since most of the HL7 tests are written against that version (or later). So I had to cherry pick this PR unto the R4 branch.

Issue #1498, opened 2020-09-17
FhirDateTime.ToDateTimeOffset() TimeSpan parameter should be optional


joellemm:
`FhirDateTime.ToDateTimeOffset()` requires a `TimeSpan zone` parameter, but this parameter should be optional. The `zone` parameter is only used for a `dto.ToOffset()` call _after_ the actual conversion is completed. This is a .NET library method that just shifts the DateTimeOffset to the specified time zone. 

`zone` should not be required parameter because the user may want the converted `DateTimeOffset` to simply remain in the time zone it is already in. **In fact, requiring the zone parameter makes this method destructive**: We lose the information of what time zone this `FhirDateTime` was originally displayed in because we are required to specify a new zone which it will be offset to.

Here is the code for reference.

```
        public DateTimeOffset ToDateTimeOffset(TimeSpan zone)
        {
            if (this.Value == null) throw new InvalidOperationException("FhirDateTime's value is null");

            // ToDateTimeOffset() will convert partial date/times by filling out to midnight/january 1 UTC
            // When there's no timezone, the UTC is assumed
            var dto = PrimitiveTypeConverter.ConvertTo<DateTimeOffset>(this.Value);

            return dto.ToOffset(zone);
        }
```

I noticed this because I upgraded my project fhir-net-api dependency from 0.96.0 to 1.9.0, and this broke a test.

This is quite a simple change that should not affect any existing use cases. I've created a commit, and all tests still seem to be passing. I just don't appear to have permissions to create a pull request. **If you let me know what I have to do to get permission, I can submit the pull request.**

joelmm:
@marcovisserFurore any update on this ticket?

Here is a unit test illustrating the problem:
```
[Test]
public void FhirDateTime_Conversion_TimeZoneInformationLoss()
{
  var fdt = new FhirDateTime(new DateTimeOffset(2021, 3, 18, 12, 22, 35, 999, new TimeSpan(-4, 0, 0)));

  Assert.AreEqual("2021-03-18T12:22:35.999-04:00", fdt.Value);

  var dto = fdt.ToDateTimeOffset();

  // Fails: Actual value is "2021-03-18T16:22:35.9990000+00:00"
  // The DateTimeOffset has been offset to UTC+0 instead of the original UTC-4
  Assert.AreEqual("2021-03-18T12:22:35.9990000-04:00", dto.ToString("o"));

  // We have lost the information of what timezone the original FhirDateTime was in
}
```
[...]

PR #1724, opened 2021-05-17
Keep existing timezone by default in FhirDateTime.ToDateTimeOffset - STU3

mmsmits:
fixes: #1498


marcovisserFurore (changes requested):
Like I mentioned in Common already:

the following is failing, which is actually correct:

```c#
  fdt = new FhirDateTime("2021-03-18T12:22:35.1234+04:00");
  Assert.AreEqual("2021-03-18T12:22:35.1234+04:00", fdt.Value);
  Assert.IsTrue(fdt.TryToDateTimeOffset(out var dto6));
```

marcovisserFurore (approved):

PR #1734, opened 2021-05-21
Changes for STU3/4/5 that belong to the common-types-parsing PR in common.

ewoutkramer:


ewoutkramer:
Forget about it. I hadn't fetched locally, so thought this branch wasn't pushed yet :-(

PR #1740, opened 2021-05-31
Add backward compatibility to parse datetime for date fields

### LINKED ISSUE #1691: Backward parsing compatibility for resources with Date
**Describe the bug**
I'm working on service which uses currently 1.9 libraries of SDK. We parse incoming resources, and save them into database as plain text (we do other things to extract search information) 1.9 library doesn't have validation for date/datetime difference. 
So we can easily accept something like this:
```
{
  "resourceType": "Patient",
  ...
  "birthDate": "1991-02-03T11:22:33Z"
}
```

We working on moving to 3.0 libraries, and at some point in our pipeline (let's say user requests object from DB) we take plain text from database and deserialize it into object.  Since 2.* libraries has validation on top of date fields I get back parsing exception:
```
resource = JsonParser.Parse<Resource>(str);

Error occurred when parsing model: 'Type checking the data: Literal '1932-09-24T11:22:33Z' cannot be parsed as a date. (at Resource.birthDate[0])'.
```

- I can't scan all databases and fix date representation.

-  I can try to put try catch block around parser check for exception type, parse exception by location, and date content, apply regex to extract date out of datetime, convert string to json, find exact node by location, replace value, and do it in a loop because some resource like `Evidence` can have multiple dates. That solution feels extremely brittle for me, I rely on exception type, exception message format, any change in them would cause me to rewrite code.

- I would like to have ParserSetting option to automatically convert DateTime to Date which would be turn off by default:
something like this:
```
new ParserSettings() { AutomaticDateConversion = true }
```

I don't mind to send PR with such changes, but I would like to hear would you be interested in accepting that change?

https://github.com/microsoft/fhir-server/pull/1743#issuecomment-797827925

**To Reproduce**
```
           var oldValidResource = @"{
  ""resourceType"": ""Evidence"",
  ""approvalDate"": ""1991-02-03T11:22:33Z"",
  ""date"": ""1991-02-03T11:22:33Z"",
  ""lastReviewDate"": ""1991-02-04T11:22:33Z""
}";
FhirJsonParser JsonParser = new FhirJsonParser(DefaultParserSettings.Settings);
var resource = JsonParser.Parse<Resource>(oldValidResource);
```

**Expected behavior**
I would like to have setting which I can specify to make sure I can parse previously valid resource into object.

**Version used:**
 - FHIR Version: [All]
 - Version: [2.*]

ewoutkramer:
Fixes #1691.

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/Serialization/SerializationTests.cs
+++ b/src/Hl7.Fhir.Core.Tests/Serialization/SerializationTests.cs
@@ -616,5 +616,21 @@ namespace Hl7.Fhir.Tests.Serialization
             Assert.IsTrue(json.ContainsKey("issued"));
             Assert.IsTrue(json.ContainsKey("status"));
         }
+
+        [TestMethod]
+        public void AllowParseDateWithDateTime()
+        {
+            var patientJson = "{ \"resourceType\": \"Patient\", \"birthDate\": \"1991-02-03T11:22:33Z\" }";
+            var parser = new FhirJsonParser();
+
+            //   Assert.ThrowsException<StructuralTypeException>(() => parser.Parse<Patient>(patientJson));
+
+#pragma warning disable CS0618 // Type or member is obsolete
+            parser = new FhirJsonParser(new ParserSettings { TruncateDateTimeToDate = true });
+#pragma warning restore CS0618 // Type or member is obsolete
+            var patient = parser.Parse<Patient>(patientJson);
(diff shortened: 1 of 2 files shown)


ewoutkramer:
Note: This won't build until we approved https://github.com/FirelyTeam/firely-net-common/pull/125 and made this PR reference the resulting commit.

marcovisserFurore (approved):

PR #1745, opened 2021-05-31
Merge R4 changes into R4B

ewoutkramer:


src/Hl7.Fhir.Core/Properties/AssemblyInfo.cs
```diff
@@ -1,9 +1,11 @@
+using Hl7.Fhir.Introspection;
 using System;
 using System.Runtime.CompilerServices;
 
 // This assembly is not fully CLSCompliant, but this triggers compiler warnings to avoid the issue as described here, when mixing C# and VB 
 // https://msdn.microsoft.com/en-us/library/ms235408(v=vs.90).aspx 
 [assembly: CLSCompliant(true)]
+[assembly: FhirModelAssembly(Hl7.Fhir.Specification.FhirRelease.R4)]
```
marcovisserFurore, line 8: Should this not be R4B?

marcovisserFurore (changes requested):
See my comment below

src/Hl7.Fhir.Core/Properties/AssemblyInfo.cs
ewoutkramer: Yes, but it has no practical consequences, so I let it slide. But I'll correct it now anyway, since this will create confusion somewhere in the future when we're not in a mood for nasty surprises.

marcovisserFurore (approved):

PR #1763, opened 2021-06-17
Fix IDeepcomparable for empty repeating elements in patterns

### LINKED ISSUE #1762: DeepComparable.Matches does not work for empty repeating elements in pattern.
**Describe the bug**
When an instance is matched against a pattern, and the pattern has *no* value for a repeating element where the instance has, `Matches` will return `false` instead of `true`.

**To Reproduce**
```csharp
  [Test]
    public static void MatchingParticipant_Fails()
    {
      var encounter = new Encounter
      {
        Participant = new List<Encounter.ParticipantComponent>
        {
          new Encounter.ParticipantComponent
          {
            Individual = new ResourceReference("Practitioner/1234")
          }
        }
      };

      var pattern = new Encounter();

      Assert.IsTrue(encounter.Matches(pattern)); // FAILS
    }
```

**Expected behavior**
The above test should pass.

**Version used:**
 - FHIR Version: all
 - Version: 1.x,2.x,3.x

**Additional context**
See this discussion thread for more background and a solution. https://chat.fhir.org/#narrow/stream/179171-dotnet/topic/.2EMatches.20not.20working.20on.20Resources.20with.20BackboneElement

mmsmits:
fixes #1762

### DIFF
--- a/build/azure-pipelines.yml
+++ b/build/azure-pipelines.yml
@@ -275,11 +275,11 @@ stages:
               displayName: 'GitHub release notes (create)'  
               inputs:
                 gitHubConnection: 'GitHub Fhir-net-api'
-                repositoryName: '$(Build.Repository.Name)'
-                action: 'create'
-                target: '$(Build.SourceVersion)'
                 tagSource: userSpecifiedTag
                 tag: '$(Build.SourceBranchName)'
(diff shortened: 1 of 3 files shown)


mmsmits:
wrong target branch oops

PR #1764, opened 2021-06-17
Fix IDeepcomparable for empty repeating elements in patterns

### LINKED ISSUE #1762: DeepComparable.Matches does not work for empty repeating elements in pattern.

mmsmits:
fixes #1762

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/Model/DeepMatchTest.cs
+++ b/src/Hl7.Fhir.Core.Tests/Model/DeepMatchTest.cs
@@ -70,7 +70,7 @@ namespace Hl7.Fhir.Tests.Model
         }
 
         [TestMethod]
-        public void MatchingParticipant_Fails()
+        public void MatchingEmptyRepeatingPattern()
         {
             var encounter = new Encounter
             {
--- a/src/Hl7.Fhir.Core.Tests/Model/DeepMatchTest.cs
+++ b/src/Hl7.Fhir.Core.Tests/Model/DeepMatchTest.cs
@@ -85,7 +85,7 @@ namespace Hl7.Fhir.Tests.Model
 
             var pattern = new Encounter();
 
-            Assert.IsTrue(encounter.Matches(pattern)); // FAILS
+            Assert.IsTrue(encounter.Matches(pattern)); 
         }
 
     }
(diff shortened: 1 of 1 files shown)


marcovisserFurore:
I think we should create a test project in the near future for Common.Poco as well. A lot of 'simple' tests are coming now in the specific Fhir version projects with the possibility of maintenance. I will put this item on the backlog. 
Btw, I am not sure whether this specific issue could be tested in common or not. I think it could.

src/Hl7.Fhir.Core.Tests/Model/DeepMatchTest.cs
```diff
@@ -68,5 +69,24 @@ public void CheckCompareListChanged()
...
+                    new Encounter.ParticipantComponent
+                    {
+                    Individual = new ResourceReference("Practitioner/1234")
+                    }
+                }
+            };
+
+            var pattern = new Encounter();
+
+            Assert.IsTrue(encounter.Matches(pattern)); // FAILS
```
ewoutkramer, line 88:
```suggestion
            Assert.IsTrue(encounter.Matches(pattern)); 
```

ewoutkramer (changes requested):

src/Hl7.Fhir.Core.Tests/Model/DeepMatchTest.cs
ewoutkramer: Looks a bit confusing to say something fails when it should not.

ewoutkramer (approved):

PR #1765, opened 2021-06-17
Make the location for XSD schemas configurable

ewoutkramer:
When using the XML validation features of the validator, the user should be able to configure where to get the necessary XSD schemas from.

### DIFF
--- a/src/Hl7.Fhir.Specification/Validation/XmlValidationExtensions.cs
+++ b/src/Hl7.Fhir.Specification/Validation/XmlValidationExtensions.cs
@@ -6,15 +6,14 @@
  * available at https://raw.githubusercontent.com/FirelyTeam/firely-net-sdk/master/LICENSE
  */
 
-using Hl7.Fhir.FhirPath;
+#nullable enable
+
 using Hl7.Fhir.Model;
 using Hl7.Fhir.Serialization;
 using Hl7.Fhir.Specification.Source;
 using Hl7.Fhir.Support;
 using System;
 using System.Collections.Generic;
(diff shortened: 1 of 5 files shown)


src/Hl7.Fhir.Specification/Validation/XmlValidationExtensions.cs
```diff
@@ -81,33 +79,33 @@ internal static OperationOutcome ValidatedParseXml(this Validator me, XmlReader
...
             {
-                result.AddIssue($"Parsing of Xml into a FHIR Poco failed: {e.Message}", Issue.XSD_CONTENT_POCO_PARSING_FAILED, (string)null);
+                result.AddIssue($"Parsing of Xml into a FHIR Poco failed: {e.Message}", Issue.XSD_CONTENT_POCO_PARSING_FAILED, default(string));
                 poco = null;
             }
 
             return result;
-        }
 
+            static OperationOutcome validateXml(XDocument instance, XmlSchemaSet xsdSchemas)
```
marcovisserFurore, line 96: missing here an access modifiers (internal probably)

marcovisserFurore (changes requested):
A very minor thing

src/Hl7.Fhir.Specification/Validation/XmlValidationExtensions.cs
ewoutkramer: Haha, took me a while to notice that this is a nested function - so no access modifiers possible!

marcovisserFurore (approved):

PR #1771, opened 2021-06-21
Update external nuget packages [STU3]

mmsmits:


marcovisserFurore (changes requested):
The meaning was to add a prop file as well for the test projects. So the storage of the versions are  maintained in 1 single file. Like we do with the 'normal' projects (`firely-net-sdk.props`).

Such a file should be in Common _and_ in the main project.

marcovisserFurore (changes requested):
I see that the build server is creating the compiled test artifact, but it is not running it.

Can you add the following snippet at line 138 in azure-pipelines.yaml (stage test, job: `Tests common test projects`):
```
**/common/**/publish/Hl7.Fhir.Support.Poco.Tests.dll
```

marcovisserFurore (approved):

PR #1783, opened 2021-06-28
R4: 1283 - Repair of constraint con-3 of Condition

### LINKED ISSUE #1283: R4 Condition validation failed (rule con-3)
Rule con-3 has error in fhir spec instead of
category.select($this='problem-list-item').empty()
shoud be
category.select($this.**code**='problem-list-item').empty()
source https://chat.fhir.org/#narrow/stream/179166-implementers/topic/.24this.20Fhirpath.20interpretation

marcovisserFurore:
## Description
There was incorrect FhirPath expression for `con-3`. This should be fixed by the standard, but in the meantime we fix this in our definition files.

In addition, we encountered another error in the `CopyTo` function of `ValidationSettings`: the member `ConstraintBestPractices` was missing during copying.

## Related issues
Fixes #1283 

## Testing
Tested by `BasicValidationTests.ConditionCon3ConstraintTest`

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Specification/Validation/ValidationSettings.cs
+++ b/src/Hl7.Fhir.Specification/Validation/ValidationSettings.cs
@@ -26,7 +26,7 @@ namespace Hl7.Fhir.Validation
     public class ValidationSettings
     {
         public StructureDefinitionSummaryProvider.TypeNameMapper ResourceMapping { get; set; }
-        
+
         [Obsolete("Use the CreateDefault() method, as using this static member may cause threading issues.")]
         public static readonly ValidationSettings Default = new ValidationSettings();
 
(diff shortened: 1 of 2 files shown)


src/Hl7.Fhir.Specification/Validation/ValidationSettings.cs
```diff
@@ -123,6 +123,7 @@ public void CopyTo(ValidationSettings other)
         {
             if (other == null) throw Error.ArgumentNull(nameof(other));
 
+            other.ConstraintBestPractices = ConstraintBestPractices;
```
ewoutkramer, line 126: oops ;-)

marcovisserFurore: Indeed!

ewoutkramer (approved):

PR #1791, opened 2021-07-01
Cleaned up package references

### LINKED ISSUE #1539: Don't reference [System/Microsoft].Net.Http for .NET 4.5+ and don't reference System.ValueTuple for .NET 4.7+
We cannot reference the `Hl7.Fhir.STU3` NuGet without pulling in `System.ValueTuple` and `System.Net.Http`.

The csproj files should add the `net47` target and change the `Condition` attributes to correctly exclude `System.ValueTuple` for `net47` and exclude `System.Net.Http`/`Microsoft.Net.Http` for `net45` and `net47`.

This also applies to the `common` submodule.

Adding `<ExcludeAssets>all</ExcludeAssets>` to the `<PackageReference />`.

**Additional context**
Immo has a [Twitter thread](https://twitter.com/terrajobst/status/997262020108926976) explaining the history of the Http NuGet packages and Framework and clearly states that now they shouldn't be used. They're only kept updated for people who took a dependency on them before since they don't want to break existing code.

@joperezr also has [a comment](https://github.com/microsoft/dotnet/issues/889#issuecomment-430466436) that is pretty clear on it as well.

They should be removed where they can.

`ValueTuple` is not necessary in .NET 4.7+ since it is built-in so this obviously should not be a dependency in NuGet either for this platform, but currently only `net45` is supported so `net47` should be added to exclude this dependency.

Thank you.

mmsmits:
fixes #1539

### DIFF
--- a/src/Hl7.Fhir.Core/Hl7.Fhir.Core.csproj
+++ b/src/Hl7.Fhir.Core/Hl7.Fhir.Core.csproj
@@ -1,56 +1,46 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0">
-  <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net45;netstandard1.6</TargetFrameworks>
-  </PropertyGroup>
+	<PropertyGroup>
+		<TargetFrameworks>netstandard2.0;net45;netstandard1.6</TargetFrameworks>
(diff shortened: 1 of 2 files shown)


marcovisserFurore (approved):
Well done!

marcovisserFurore (approved):

PR #1806, opened 2021-07-08
Move target from netstandard2.1 to net5.0 - STU3

mmsmits:


src/Hl7.Fhir.Specification/Validation/XmlValidationExtensions.cs
```diff
@@ -6,7 +6,9 @@
  * available at https://raw.githubusercontent.com/FirelyTeam/firely-net-sdk/master/LICENSE
  */
 
-#nullable enable
+#nullable enable annotations
+#nullable disable warnings
```
marcovisserFurore, line 10: At the bottom of this file, there is a `#nullable restore` directive. Should this be changed as well?

src/Hl7.Fhir.Specification/Validation/XmlValidationExtensions.cs
```diff
@@ -6,7 +6,9 @@
  * available at https://raw.githubusercontent.com/FirelyTeam/firely-net-sdk/master/LICENSE
  */
 
-#nullable enable
+#nullable enable annotations
```
marcovisserFurore, line 9: Changing line 100 to
```
instance.Validate(xsdSchemas, (o, args) => { result.AddIssue(ToIssueComponent(args)); });
```

will remove the nullable warning and we don't have to disable the nullable warnings.
To be honest: I don't know why this solves the warning, but I followed the example [here](https://docs.microsoft.com/en-us/dotnet/api/system.xml.xmldocument.validate?view=net-5.0#system-xml-xmldocument-validate(system-xml-schema-validationeventhandler)).

marcovisserFurore (changes requested):
In `FallbackTerminologyService` there is still a warning, which we would handle I think. So also my other comments:

src/Hl7.Fhir.Specification/Validation/XmlValidationExtensions.cs
marcovisserFurore: No, because:

> #nullable restore: Restores the nullable annotation context and nullable warning context to the project settings.

marcovisserFurore (approved):

PR #1810, opened 2021-07-13
STU3: Behavior of PocoStuctureDefinitionSummaryProvider is now the same as StuctureDefinitionSummaryProvider

### LINKED ISSUE #1576: IStructureDefinitionSummary differs between PocoSDSP and StructureDefinitionSDSP
**Describe the bug**
We currently ship two implementations of `IStructureDefinitionSummaryProvider`: `PocoStructureDefinitionSummaryProvider` and `StructureDefinitionSummaryProvider`.

One would expect that the `IStructureDefinitionSummary` information provided by both for the same resource would yield the same results. This is however not the case.

**To Reproduce**
1. Retrieve an `IStructureDefinitionSummary` from both providers for Observation
2. Compare the `IsSummary` property for the elements `modifierExtension` and `id`
3. It will be `true` for the info returned by the `PocoSDSP` and `false` for the `SDSP`.
4. Retrieve an `IStructureDefinitionSummary` from both provider for Observation.referenceRange
5. The BackboneElement will be `abstract` for `ISDSP` and not abstract for the `PocoSDSP`.
6. The same is true for `Order` in all elements.

**Expected behavior**
The information should be the same, maybe except for `Order`, since that should just retain the same sequence.

**Version used:**
- 2.0.1

**Additional context**
The differences exist because the `InSummary` for 'id' (called `ElementId` in the POCO) is hardwired to be "true", while it is false in the StructureDefinition. The same is true for `modifierExtension` (which is surprising - shouldn't they indeed be in summary?).

marcovisserFurore:
## Description
The behavior of `PocoStuctureDefinitionSummaryProvider` is now the same as `StuctureDefinitionSummaryProvider`. The solution was to remove the attribute InSummary=true for the following pocos elements:
- CarePlan.activity.detail.status
- ConceptMap.group.element.target.equivalence
- Group.characteristic.exclude
- List.entry.deleted
- MedicationRequest.substitution.allowed

Those elements had this `InSummary` attribute while it was absent in the FHIR specification. Therefor the attribute `InSummary` has been removed.

The FhirAttribite `IsModifier` has been added to Poco elements according to the FHIR specification. `_summary=true` is now depedent on `IsModifier` or `InSummary`.

See also https://github.com/FirelyTeam/firely-net-common/pull/138
See also PR https://github.com/FirelyTeam/fhir-codegen/pull/14 in FirelyTeam/fhir-codegen

## Related issues
Resolves #1576.

## Testing
See unittest `Hl7.Fhir.Specification.Tests.PocoAndSdSummaryProvidersShouldBeEqual()`

## FirelyTeam Checklist


ewoutkramer (approved):

marcovisserFurore:
Don't pull this PR until we have a consensus on Zulip: https://chat.fhir.org/#narrow/stream/294899-dotnet.2Fcommitters/topic/Summary.20elements

ewoutkramer (dismissed):

PR #1826, opened 2021-07-27
Snapshot generator: corrected cardinality of element with a derived type

### LINKED ISSUE #1824: Wrong element cardinality
Derived profile contains wrong Min cardinality even though it was not changed in regards to the base profile.

Original bug was reported for Forge (https://firely.atlassian.net/browse/FOR-497). Additional information can be found there.

FHIR version: R4 (may also exist in other Fhir versions but not tested) 
SDK version: 3.4.0

Unit test to reproduce error (directory source folder contents: [resourceProfiles.zip](https://github.com/FirelyTeam/firely-net-sdk/files/6878126/resourceProfiles.zip)):

        [TestMethod]
        public void TestObservationCodeCardinality()
        {
            // Arrange
            var zipSource = ZipSource.CreateValidationSource();
            var dirSource = new DirectorySource("resourceProfiles\\FOR-497");
            var resolver = new CachedResolver(new MultiResolver(zipSource, dirSource));
            var snapshotGenerator = new SnapshotGenerator(resolver, SnapshotGeneratorSettings.CreateDefault());
            var sd = resolver.ResolveByCanonicalUri("http://hl7.org/fhir/StructureDefinition/Observation") as StructureDefinition;
            var sut = resolver.ResolveByCanonicalUri("http://hl7.org/fhir/uv/ips/StructureDefinition/Observation-results-uv-ips") as StructureDefinition;

            // Act
            var elements = snapshotGenerator.Generate(sut);

            // Assert
            snapshotGenerator.Outcome.Should().BeNull();

            const string codeId = "Observation.code";

            var sdCode = sd.Snapshot.Element.Single(x => x.ElementId == codeId);
            var sutCode = elements.Single(x => x.ElementId == codeId);

            sutCode.Max.Should().Be(sdCode.Max);
            sutCode.Min.Should().Be(sdCode.Min);
        }

marcovisserFurore:
## Description
When you were using a profile with an element that has again a profiled type. then the snapshot generator used the cardinality of that profiled type, and not the cardinality of the element. So

For example when you profile the element `Observation.code` and you use a derived type for this element, for example `CodeableConceptProfile`, then the cardinality of `CodeableConceptProfile` was used for the element  `Observation.code`. In this case the minimum cardinality of `Observation.code` went from `1`  to `0`, which is incorrect. This has been solved in this PR.

## Related issues
Fixes #1824 

## Testing
Unit test `SnapshotGeneratorTest2.CheckCardinalityOfProfiledType()`

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
+++ b/src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
@@ -34,14 +34,14 @@
 // [WMR 20180409] Resolve contentReference from core resource/datatype (StructureDefinition.type)
 #define FIX_CONTENTREFERENCE
 
-using System;
-using System.Collections.Generic;
-using System.Linq;
 using Hl7.Fhir.Model;
 using Hl7.Fhir.Specification.Navigation;
 using Hl7.Fhir.Specification.Source;
-using System.Diagnostics;
 using Hl7.Fhir.Utility;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
 using T = System.Threading.Tasks;
 
 namespace Hl7.Fhir.Specification.Snapshot
@@ -441,7 +441,7 @@ namespace Hl7.Fhir.Specification.Snapshot
             // e.g. Extension : BaseDefinition => Element : Element.extension => Extension
             // Then snapshot root is already generated and annotated to differential.Element[0]
             //Debug.WriteLineIf(diffRoot.HasSnapshotElementAnnotation(), $"[{nameof(SnapshotGenerator)}.{nameof(generate)} (before merge)] diff root is annotated with cached snapshot root...");
-            
+
             await merge(nav, diff).ConfigureAwait(false);
 
             result = nav.ToListOfElements();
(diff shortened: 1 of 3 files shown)


src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
```diff
@@ -1167,6 +1167,9 @@ private async T.Task<bool> mergeTypeProfiles(ElementDefinitionNavigator snap, El
                             // Rebase before merging
                             var rebasedRootElem = (ElementDefinition)typeRootElem.DeepCopy();
                             rebasedRootElem.Path = diff.Path;
+                            // MV 20210727: copy cardinality from base (so do not use the cardinality of the type). See issue #1824
```
ewoutkramer, line 1170: Doesn't this mean that if you have an extension definition (which is a profiled Extension), that it will ignore the extension's cardinality now too?  Isn't it safer (for regressions) to take the higher one of the "min" and the lower one of the "max", so that you can still override this cardinality in a profile (at least for Extension)?

ewoutkramer (changes requested):

src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
marcovisserFurore: So, I had a lot of thinking about this and I was experimenting with Extension as well and I could not find an error there.

I think the cardinality of a type (what does that actually mean?)  should influence the cardinality of an element. The cardinality of `Observation.code` is `1..1`. When you profile this and you change the type of `Observation.code` to a profiled `CodeableConcept`  with cardinality `2..*` (again, what does that mean?), then the cardinality of the element `Observation.code` should still be `1..1` to my opinion.

ewoutkramer (approved):

PR #1827, opened 2021-07-28
R4: Combined discriminators can now also be used when one of them cannot discriminate

### LINKED ISSUE #1686: When using combined discriminators, can one of them not discriminate?
**Describe the bug**
The profile at https://github.com/FHIR/fhir-test-cases/blob/master/validator/mixed-type-slicing-profile.xml declares a slice with a dual discriminator, both using a `type` discriminator:

``` xml
<discriminator>
   <type value="type"/>
   <path value="content"/>
</discriminator>
<discriminator>
  <type value="type"/>
  <path value="content.ofType(Reference).resolve()"/>
</discriminator>
```

One of the slices that follow (starting at id `Communication.payload:String`) is *not* a reference, and therefore, the
second of the two discriminators has "nothing" to work on.  The SDK complains about this, and refuses this setup.

**Expected behavior**
This is the discussion: the disciminator *combination* is still unique (as long as the types of the slices differ), however, in the String slice, the second discriminator does not discriminate. Is this valid?  I can see it could be, it would certainly make the usecase tested in the linked profile possible.  We should probably update the SDK so it validates whether the combination of discriminators is valid, not a single one of the component discriminators.

**Version used:**
 - FHIR Version: R4
 - Version: 3.1

marcovisserFurore:
## Description
Using combined discriminators will now also work when one of the discriminators does not discriminate. 
Also solved in this PR:
- `resolve()` in a discriminator expression will also work when it is executed on the Reference definition (the targetProfiles are copied into the SDWalker

## Related issues
Fixes #1686.

## Testing
Describe how this change was tested.

## FirelyTeam Checklist


src/Hl7.Fhir.Specification/Specification/Navigation/StructureDefinitionWalker.cs
```diff
@@ -141,9 +163,9 @@ public IEnumerable<StructureDefinitionWalker> Expand()
             else if (Current.Current.Type.Count >= 1)
             {
                 return Current.Current.Type
-                    .Select(t => t.GetTypeProfile())
+                    .Select(t => (profile: t.GetTypeProfile(), targetProfiles: t.TargetProfile))
```
ewoutkramer, line 166: Multiple targetProfiles will always cause an ambiguous discriminator, so I think we should only allow one single target profile?

ewoutkramer (changes requested):

src/Hl7.Fhir.Specification/Specification/Navigation/StructureDefinitionWalker.cs
marcovisserFurore: And throw an exception when there are more than 1 target profiles? In that case we can also make the private field `_targetProfiles` a string instead of an array of strings.

marcovisserFurore: Hmm, that is not possible, because this code is part of the method `Expand()` which is used elsewhere as well.

marcovisserFurore: Looking at the code for the second time, I think the initial code is correct. Allowing multiple targetProfiles will result in multiple `StructureDefinitionWalker`'s in function `Resolve()`, which is still allowed, because the profile author can decide to concatenate other functions (like `first()`) on to these walkers.

In the method `DiscriminatorFactory.walkToCondition` is the final check whether there are more than 1 conditions. In that case an exception will be thrown.

ewoutkramer (approved):

PR #1831, opened 2021-07-30
Changes to ClassMapping to support new serializer (for R3-R5)

ewoutkramer:
Added benchmark project, and included benchmark test for ModelInspector

### DIFF
--- a/src/Benchmarks/ModelInspectorBenchmarks.cs
+++ b/src/Benchmarks/ModelInspectorBenchmarks.cs
@@ -1,0 +1,66 @@
+﻿using BenchmarkDotNet.Attributes;
+using Hl7.Fhir.Introspection;
+using Hl7.Fhir.Model;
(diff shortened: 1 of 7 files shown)


src/Benchmarks/ModelInspectorBenchmarks.cs
```diff
@@ -0,0 +1,66 @@
+﻿using BenchmarkDotNet.Attributes;
+using Hl7.Fhir.Introspection;
+using Hl7.Fhir.Model;
+using System;
+
+namespace Benchmarks
```
marcovisserFurore, line 6: Should we prefix the namespace with `Firely`? Now it is so generic.

marcovisserFurore (changes requested):
Prefix namespace?

marcovisserFurore (approved):

PR #1848, opened 2021-08-06
Don't Inherit invariants and conditions from datatypes [STU3]

### LINKED ISSUE #1836: [SnapshotGenerator] Invariants shouldn't be inherited from the datatype
**Describe the bug**
Currently, all invariants (elementDefinition.constraint) are being copied from the element's datatype definition. This is redundant, since the element already states it conforms to the datatype (including constraints), and makes the snapshot unnecessarily big, and snapshot generation unnecessarily slow

**To Reproduce**
- Generate any snapshot. 
- All elements automatically have constraint "ele-1".

**Expected behavior**
- Elements should only state the constraints that are introduced on that element.

**Version used:**
- FHIR Version: [STU3, R4, R5]
- Version: at least 1.9.0 +

**Additional context**
Java snapgen doesn't repeat the constraints on all elements either

mmsmits:
fixes #1836

For discussion: I now have specifically fixed this for invariants and conditions, but should there be anything inherited at all from the datatype? I can't think of something....


mmsmits:
Closed after discussion: this isn't necessary, and would differ from the java snapshot generator

PR #1861, opened 2021-08-23
Return correct match when parsing similar polymorphic elements in a single resource [new serializer]

### LINKED ISSUE #1858: FhirXmlParser -> Bundle -> type checking failure for Ingredient
Hi guys!

When an Ingredient is present in the Bundle and it contains both _presentation[x]_ and _presentationHighLimit[x]_ nodes (see structure image from http://hl7.org/fhir/2021May/ingredient.html below)

image

in that order (_pres..[x]_ then _pres..HighLimit[x]_), we are seeing an error whilst attempting to parse the Bundle XML:

`Type checking the data: Choice element 'presentationHighLimitRatio' is suffixed with unexpected type 'HighLimitRatio' (at Bundle.entry[15].resource[0].substance[0].strength[0].presentationHighLimitRatio[0])`

**To Reproduce**
Have an entry in your Bundle similar to 

image

where _strength_ node has both types of _presentations_ then attempt to parse it with the use of 

`new FhirXmlParser().Parse<Bundle>(...)`

**Expected behavior**
Both _presentation[x]_ and _presentationHighLimit[x]_ should be recognised as valid elements given that they are post-fixed with a correct / allowed value.

**Version used:**
 - "Hl7.Fhir.R5" Version="3.4.0"
 - "Hl7.Fhir.Specification.R5" Version="3.4.0"

Many thanks for your help with this!

mmsmits:
fixes #1858


marcovisserFurore:
Closing this PR, because this change is also already in PR https://github.com/FirelyTeam/firely-net-sdk/pull/1851.

PR #1877, opened 2021-09-14
Do not allow FhirPath navigating with ChoiceTypes that include a typename

### LINKED ISSUE #1785: Validator rejects valid resource: Instance failed constraint xy (FhirPath constraint).
**Description**
Being the first time here first of all I would like to say thank you for providing and supporting firely.sdk! It helps us a lot.

Using the validator to validate bundle resources I encountered a problem. The validator rejected a valid resource as invalid, stating that a certain FirePath constraint would not be met:

```
[ERROR] Instance failed constraint angabeAUseitErstbescheinigung "Die Angabe arbeitsunfaehig seit ist nicht gesetzt, muss aber bei einer Erstbescheinigung gesetzt werden."(further diagnostics: (         entry.where(resource is Composition).resource.type.coding.where(system = 'https://fhir.kbv.de/CodeSystem/KBV_CS_EAU_AU_Type').code = 'ERST'          or         entry.where(resource is Composition).resource.type.coding.where(system = 'https://fhir.kbv.de/CodeSystem/KBV_CS_EAU_AU_Type').code = 'ERST_END'         )implies(         entry.select(resource as Condition).where(meta.profile = 'https://fhir.kbv.de/StructureDefinition/KBV_PR_EAU_Condition_AU|1.0.2').onsetPeriod.start.exists()        )) (at Bundle)
```
This problem only occurs when validating Poco-based resources (poco -> ITypedElement). If validation is based on deserialized xml (ISourceNode -> ITypedElement) all is fine.

Looking into it it seems to me that the element node names used when validating poco-based resources (provided by PocoElementNode instances) in certain cirmcumstances are wrong. In my case a childnode named "onset" was shown. However the FirePath constraint demands the existance of a childnode named "onsetPeriod", thus failing the validation.

Attached you find a project showing the problem (and a work around when using the ISourceNode approach). Moreover I added a screenshot showing the "onset" childnode name seen in a debugging session.

**To Reproduce**
See attached solution.

[FirelyValidatorPocoElementNodeProblem.zip](https://github.com/FirelyTeam/firely-net-sdk/files/6734543/FirelyValidatorPocoElementNodeProblem.zip)

**Screenshot**
ScreenshotChildNodeName

**Version used:**
 - FHIR Version: R4
 - Version: 3.3.0

marcovisserFurore:
## Description
Navigating in FhirPath cannot be done anymore with element names that contains choice element with a type name. For example `parameter[0].valueString` is not allowed anymore, while `parameter[0].value` is still allowed. `value` is here a Choice element.

## Related issues
Fixes #1785.

## Testing
Unit test: `Hl7.Fhir.Support.Tests.FhirPathTests.NavigateWithChoiceTypes`

## FirelyTeam Checklist


ewoutkramer (dismissed):

ewoutkramer:
> @Ewout Kramer condition.onsetPeriod should not be supported. That's a bug and surely there's a test case for that?

It's a bug, and should not be supported. This was allowed a long time ago (around 2016, or DevDays 1 I think), so I think it's time to say bye bye ;-)

PR #1879, opened 2021-09-20
Include packaging functionalities [STU3]

mmsmits:


marcovisserFurore (approved):
The project `Firely.Fhir.Packages` has been added to the SDK and included in the Azure Pipelines. Also it can now be compiled against `netstandard1.6`.

The following steps that we still need to make are:
- reduce the number of public methods/classes
- polish the code to make it publishable.

marcovisserFurore (changes requested):
Some very small things...

PR #1882, opened 2021-09-23
ToTypedElementOnSourceNode: caching children and value

### LINKED ISSUE #482: PocoNavigator.Value is too expensive
The Value property of IElementNavigator is accessed pretty frequently by all the supporting infrastructure (Validator, FhirPath evaluator, serializer) and therefore needs to be fast. The current implementation will frequently call out to methods that parse and serialize raw values - which takes too much time.

I propose we cache this property in the PocoNavigator.

marcovisserFurore:
## Description
Describe the changes in this PR.

## Related issues
Resolves #482 

## Testing
Describe how this change was tested.

## FirelyTeam Checklist


ewoutkramer (commented):
Should we include the /results directory? I did not do it for the serializer/deserializer, but maybe it's nice as a reference (though useless across different machines....)

marcovisserFurore:
> Should we include the /results directory? I did not do it for the serializer/deserializer, but maybe it's nice as a reference (though useless across different machines....)

This is not relevant anymore

PR #1883, opened 2021-09-23
R4: Validate resources with targetprofiles and child definitions

### LINKED ISSUE #1654: The validator is not resolving certain references
**Describe the bug**
We have identified some cases in which the validator is not calling the IResourceResolver to resolve certain relative references.

**To Reproduce**

Both this resources have a reference to `Patient/UKCore-Patient-RichardSmith-Example`:
https://simplifier.net/ukcore/ukcore-careteam-weightmanagementteam-example
https://simplifier.net/ukcore/ukcore-medicationstatement-amoxicillin-example

However, validating them the outcome is different:
1. Gives warnings:  https://simplifier.net/ukcore/ukcore-careteam-weightmanagementteam-example/$validate
2. Doesn't give warnings: https://simplifier.net/ukcore/ukcore-medicationstatement-amoxicillin-example/$validate

Our investigation concluded that in the case of the resource that **doesn't give warnings** the validator is not calling the IResourceResolver to resolve the reference.

**Expected behavior**
Both validations should attempt to resolve the reference to `Patient/UKCore-Patient-RichardSmith-Example`.

**Screenshots**

**Version used:**
 - FHIR Version: R4 (haven't tried with STU3)
 - Version: Tried with both **1.9.0** and **2.0.2**.

**Additional context**

marcovisserFurore:
## Description
A profile that had a target profile on a reference type *and* other child definitions on that reference, then the validator did not resolve the reference and skipped the whole validation of that referenced resource. This has been resolved now. 

## Related issues
Fixes #1654 

## Testing
See `BasicValidationTests.ValidateWithTargetProfileAndChildDefinitions`

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Specification/Validation/TypeRefValidationExtensions.cs
+++ b/src/Hl7.Fhir.Specification/Validation/TypeRefValidationExtensions.cs
@@ -7,7 +7,6 @@
  */
 
 using Hl7.Fhir.ElementModel;
-using Hl7.Fhir.Introspection;
 using Hl7.Fhir.Model;
 using Hl7.Fhir.Rest;
 using Hl7.Fhir.Support;
@@ -26,7 +25,7 @@ namespace Hl7.Fhir.Validation
 
             validator.Trace(outcome, "Validating against constraints specified by the element's defined type", Issue.PROCESSING_PROGRESS, instance);
 
-            if(definition.Type.Any(tr => tr.Code == null))
+            if (definition.Type.Any(tr => tr.Code == null))
                 validator.Trace(outcome, "ElementDefinition contains a type with an empty type code", Issue.PROFILE_ELEMENTDEF_CONTAINS_NULL_TYPE, instance);
 
             // Check if this is a choice: there are multiple distinct Codes to choose from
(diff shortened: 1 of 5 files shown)


src/Hl7.Fhir.Specification/Validation/TypeRefValidationExtensions.cs
```diff
@@ -102,14 +101,30 @@ OperationOutcome validate()
...
-                // If this is a reference, also validate the reference against the targetProfile
-                if (ModelInfo.FhirTypeNameToFhirType(tr.Code) == FHIRAllTypes.Reference)
-                    result.Add( validator.ValidateResourceReference(instance, tr) );
+                result.Add(validateTargetProfiles(validator, instance, tr));
 
                 return result;
             }
         }
 
+        internal static OperationOutcome ValidateTargetProfiles(this Validator validator, ElementDefinition definition, ScopedNode instance)
```
ewoutkramer, line 110: This one file now has a method `ValidateTargetProfiles()` and a method `validateTargetProfiles()` with almost the same signature.

ewoutkramer: I think the method starting with lowercase is the original behaviour, so it is actually not validating the target profile?

ewoutkramer (changes requested):

src/Hl7.Fhir.Specification/Validation/TypeRefValidationExtensions.cs
marcovisserFurore: In the method `ValidateChildConstraints` I wanted only to validate the targetprofiles, which is part of the method `ValidateTypeReferences`. To not duplicate code I wrote a new method `ValidateTargetProfiles` which is used by the ChildConstraints and `ValidateTypeReferences`.
I have now decided to use the method `ValidateTypeReferences` in `ValidateChildConstraints`. That does also validate the targetprofiles and the declared profiles. The issue was only about targetprofiles, but to validate the declared profiles as well does not hurt.

ewoutkramer (approved):

PR #1917, opened 2021-11-23
SnapshotGenerator: continue merging diff child constraints with multiple types

### LINKED ISSUE #1898: Snapshot generator Generate method removes or ommits an extension when the element type has a custom profile
This bug was reported for Forge:
https://firely.atlassian.net/browse/FOR-580

The snapshot generator Generate method removes or ommits an extension when the element type has a custom profile.

Unit test is available in this branch:
https://github.com/FirelyTeam/firely-net-sdk/tree/forge/FOR-580

Expected behavior is that the snapshot generator does not remove the extension.

marcovisserFurore:
## Description
Continue merging diff child constraints in the case of multiple types.

## Related issues
Resolves #1898.

## Testing
See unit test `Hl7.Fhir.Specification.Tests.SnapshotGeneratorTest2.ContinueMergingChildConstraintsMultipleTypes`

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
+++ b/src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
@@ -966,19 +966,13 @@ namespace Hl7.Fhir.Specification.Snapshot
 
             var diffTypes = diff.Current.Type;
             var distinctTypeCodeCnt = diffTypes.DistinctTypeCodes().Count;
-            if (distinctTypeCodeCnt == 0)
+            if (distinctTypeCodeCnt != 1)
             {
                 // Element has no type constraints, nothing to merge
                 // return true to continue merging child constraints
-                return true;
-            }
-            else if (distinctTypeCodeCnt > 1)
-            {
+                // OR
                 // Element specifies multiple type codes, cannot expand children
-                // return false to prevent merging child constraints
-
-                // return false;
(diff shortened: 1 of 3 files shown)


src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
```diff
@@ -976,7 +976,10 @@ private async T.Task<bool> mergeTypeProfiles(ElementDefinitionNavigator snap, El
             {
                 // Element specifies multiple type codes, cannot expand children
                 // return false to prevent merging child constraints
-                return false;
+
+                // return false;
+                // [MV 20211123] return true to continue merging diff child constraints
+                return true;
```
mmsmits, line 982: But` if` and `else if` return true.
But I get that you might want to keep this, in case something else breaks with this commit.

I think Michel's todo` [WMR 20171004] `above explains what should happen.

mmsmits (changes requested):

src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
marcovisserFurore: Better would be:
```c#
if (distinctTypeCodeCnt != 1 )
{ 
  return true;
}
else
{
// the rest 
}
```
I will change that

mmsmits (approved):

PR #1932, opened 2021-12-29
FiveWs autogenerated files

itye-msft:
## Description
This PR is a continuation move to add support for the FiveWs property mappings.
It depends on an open PR on Common https://github.com/FirelyTeam/firely-net-common/pull/178
Once the common PR is merged, this PR will pass.

This PR contains:

- A list of auto generated files using CodeGen (with guidance of Gino)
- Unit test

## FirelyTeam Checklist


marcovisserFurore:
Hi @itye-msft, I'm working on the FiveWs at the moment. We would like it to pull it for Firely .NET SDK  4.0. But I have some questions about this:
- why is the property `FiveWs` of `FhirElement` a string array? As far as I know a Fhir element cannot map to more than 1 FiveW elements. Also arrays as attribute arguments is not CLS-compliant. I would prefer that the property `Fivews` of  `FhirElement` is of type `string`, which would be a [codegen](https://github.com/microsoft/fhir-codegen) change.
- I see some mappings that are not valid FiveWs elements: like `AuditEvent.recorded` is mapped to `when.recorded`. That should be `recorded` I think. See also [here](http://hl7.org/fhir/fivews.html).

GinoCanessa:
@marcovisserFurore , likely the array came from `AllergyIntolerance.patient`, which has an obviously incorrect mapping of both `FiveWs.subject[x]` and `FiveWs.subject`.  Changing to a single value makes sense - would you like me to make the change in code-gen?

For the other mapping issue, I will add some internal consistency checking for the values.

marcovisserFurore:
Thanks @GinoCanessa, that would be great when you can make the change in codegen. I'll make sure that the needed changes (single string value) in the SDK are made.

GinoCanessa:
@marcovisserFurore , I have pushed up the changes to code-gen:
* FhirElement.FiveWs is now a string.
* CSharpFirely2 uses a primitive string instead of an array
* CSharpFirely2 now includes 5Ws mapping by default (can disable via language flag 'w5')
* FiveWs.subject[x] (incorrect flag) is now filtered for and not present
* Any FiveWs mapping that does not begin with the proper prefix is filtered.

The changes are on the `main` branch, but also in `sequence/v1`, which will be the long-term home of this series.  I have made the mirroring changes on `dev` and `sequence/v2`.  The reason for the new sequence is a large refactor to package loading.  It was needed for better support of IG packages, CI builds, the `~/.fhir` cache, R4B, etc.

There is no rush to update as the v2 sequence is still under pretty heavy development.  As of now, there are no breaking changes, so it should be painless.

marcovisserFurore:
Wow, thank you @GinoCanessa. I will update this PR (or better, I create a new one) so that `FiveWs` are also available in the SDK.

marcovisserFurore:
This PR has been replaced by https://github.com/FirelyTeam/firely-net-sdk/pull/1973. Thanks @itye-msft and @GinoCanessa for helping out with this issue.

marcovisserFurore:
> @marcovisserFurore , I have pushed up the changes to code-gen:
> 
> * FhirElement.FiveWs is now a string.
> * CSharpFirely2 uses a primitive string instead of an array
> * CSharpFirely2 now includes 5Ws mapping by default (can disable via language flag 'w5')
> * FiveWs.subject[x] (incorrect flag) is now filtered for and not present
> * Any FiveWs mapping that does not begin with the proper prefix is filtered.
> 
> The changes are on the `main` branch, but also in `sequence/v1`, which will be the long-term home of this series. I have made the mirroring changes on `dev` and `sequence/v2`. The reason for the new sequence is a large refactor to package loading. It was needed for better support of IG packages, CI builds, the `~/.fhir` cache, R4B, etc.
> 
> There is no rush to update as the v2 sequence is still under pretty heavy development. As of now, there are no breaking changes, so it should be painless.

@GinoCanessa, I updated the SDK now with the FiveWs attributes. For that I changed the codegen slightly because I did not like the empty string assignment for every attribute which does not have a FiveWs mapping. What should I do with this change? Commit it to `main`,  `sequence/v1`, `dev` or `sequence/v2`? Just let me know and I will add the PR to the right branch. Again many thanks for the quick support.

PR #1934, opened 2022-01-04
1095 - Move useful ElementDefinition ext methods to a single place (R3)

ewoutkramer:
Collected useful ElementDefinition extension methods to a single ElementDefinitionExtensions class in the model assembly.


src/Hl7.Fhir.Specification/Validation/TypeRefExtensions.cs
```diff
@@ -6,38 +6,9 @@
...
  */
 
-using System.Linq;
-using Hl7.Fhir.Introspection;
-using Hl7.Fhir.Model;
-using Hl7.Fhir.Support;
-using System.Collections.Generic;
-using Hl7.Fhir.Utility;
-
 namespace Hl7.Fhir.Validation
```
marcovisserFurore, line 9: Should this file not removed then?

marcovisserFurore (changes requested):
I think file TypeRefExtensions.cs can be removed as well, right?

marcovisserFurore (approved):

mmsmits (approved):

PR #1969, opened 2022-02-15
Validator checks if invariants don't result in empty anymore [STU3]

### LINKED ISSUE #1929: FhirPath invariants should only be met is the result is `true`, not `false` or `empty`
**Describe the bug**
Our validator invokes the FhirPath engine for evaluating invariants by calling the `Predicate` function. This function returns `true` or `false`, which is supposed to mean that the invariant has been met or not.

This discussion: https://chat.fhir.org/#narrow/stream/179166-implementers/topic/Period.20Datatype.20Invariant states that invariants are considered to be "met" when the result is true, not false or empty, however our current code treats the fhirpath statement as a predicate, and as such considers `empty` to be `true` as well.

**Expected behavior**
We might need to add a new (extension) method to run FhirPath that implements the desired "constraint" functionality, so that only a `true` outcome results in `true`.  The validator then should call this new function, not `Predicate()`.

This is generally speaking a breaking change, but since most FhirPath statements in the spec are prefixed by `<data>.exists(),not() or <data>.someconstraint` the differences should be minimal (which is the reason we never noticed this "bug" in the first place).

**Version used:**
Any.

mmsmits:
## Description
Validator uses new IsTrue() FhirPath evaluation for invariants instead of Predicate()

## Related issues
Fixes #1929

## FirelyTeam Checklist


src/Hl7.Fhir.Core/FhirPath/FhirPathExtensions.cs
```diff
@@ -48,6 +48,12 @@ public static bool Predicate(this Base input, string expression, FhirEvaluationC
             return inputNav.Predicate(expression, ctx ?? FhirEvaluationContext.CreateDefault());
         }
 
+        public static bool IsTrue(this Base input, string expression, FhirEvaluationContext ctx = null)
+        {
+            var inputNav = input.ToTypedElement();
+            return inputNav.Predicate(expression, ctx ?? FhirEvaluationContext.CreateDefault());
```
ewoutkramer, line 54: Why isn't this a call to `IsTrue` instead of `Predicate`?

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #1978, opened 2022-02-18
Merge package functionalities in R4B

mmsmits:


marcovisserFurore (changes requested):
Can you add an extra unittest for resolving the currency with packages?

marcovisserFurore (approved):

PR #1979, opened 2022-02-18
Merge package functionalities in R5

mmsmits:


src/Hl7.Fhir.Core.Tests/Model/ModelTests.cs
```diff
@@ -760,5 +760,12 @@ public void TestHumanNameFluentInitializers()
             pat.Name.FirstOrDefault().Family.Should().Be("Levin");
             pat.Name.FirstOrDefault().Suffix.Should().ContainSingle("The 7th");
         }
+
+        [TestMethod]
+        public void TestCorrectCurrencyDescription()
```
marcovisserFurore, line 765: Can you add some instructions in any case this unittest fails? Because after generating the pocos we have to manually correct the Money class.

marcovisserFurore (changes requested):
Can you add an extra unittest for resolving the currency with packages?

marcovisserFurore (approved):

PR #1990, opened 2022-02-24
3-way merge of cardinality of an element

### LINKED ISSUE #1981: Snapshot generator calculates invalid max cardinality for extensions
A bug has been reported for Forge:
https://firely.atlassian.net/browse/FOR-611

When adding a new extension to a profile, save it and reload it, the max cardinality is set to "*" while "1" is expected.

I have created a unit test in this branch (based on 3.8.0 R4 release) to reproduce the problem:
https://github.com/FirelyTeam/firely-net-sdk/tree/forge/FOR-611

The expected behavior is that the BaseElement and derived element contains the correct max cardinality.

FHIR Version: R4 (may apply to other versions as well)
SDK Version: 3.6.0 and newer (works as expected in version 3.4.0!)

marcovisserFurore:
## Description
In PR #1826 we introduced a bug in existing functionality. We resolved that now by a 3-way merge of the cardinality of
- snapshot
- differential
- external profile type

The most constrained value will win. Let's say we have the following cardinalities on an element:
- snapshot: `0..*`
- differential: `1..*`
- external profile type (like an extension): `1..1`

Then the result cardinality of the element would be: `1..1`

This behavior is also applied to the "normal" merge of `Min`: in the past the Min value of the differential was copied to the snapshot (when it exists). Now  both Min values (snap and diff) are compared first, and the highest value wins.

## Related issues
Resolves #1981.

## Testing
- `SnapshotGeneratorTest2.TestMinMax`
- `SnapshotGeneratorTest2.CardinalityOfExtension`

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefnMerger.cs
+++ b/src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefnMerger.cs
@@ -6,14 +6,13 @@
  * available at https://raw.githubusercontent.com/FirelyTeam/firely-net-sdk/master/LICENSE
  */
 
+using Hl7.Fhir.Model;
+using Hl7.Fhir.Specification.Navigation;
+using Hl7.Fhir.Utility;
 using System;
 using System.Collections.Generic;
 using System.Linq;
-using Hl7.Fhir.Model;
-using Hl7.Fhir.Specification.Navigation;
-using Hl7.Fhir.Support;
 using System.Reflection;
-using Hl7.Fhir.Utility;
 
 namespace Hl7.Fhir.Specification.Snapshot
 {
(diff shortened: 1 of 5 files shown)


ewoutkramer (approved):

src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefnMerger.cs
```diff
@@ -278,9 +308,9 @@ internal FhirString mergeMax(FhirString snap, FhirString diff)
                     return snap;
             }
 
-            private FhirString deepCopyAndRaiseOnConstraint(FhirString elt)
+            private T deepCopyAndRaiseOnConstraint<T>(T elt) where T : PrimitiveType
```
mmsmits, line 311: Nice!

mmsmits (approved):

Issue #1986, opened 2022-02-23
Remove core packages from SDK


mmsmits:
- [x] Constructor CorePackageSource with package filepaths
- [x] XSD validator with specification.zip
- [x] Core packages included in unit tests
- [x] Remove packages from specification project

PR #1995, opened 2022-02-28
Removes CorePackages from the specification project and adds Package client

mmsmits:
fixes: #1987 #1986


marcovisserFurore (changes requested):
See https://github.com/FirelyTeam/firely-net-common/pull/189

PR #1999, opened 2022-03-02
Fix for PR #1990

### LINKED ISSUE #1981: Snapshot generator calculates invalid max cardinality for extensions

Rob5045:
PR #1990 did not solve the issue (checked with additional unit tests). That is because the use of the new mergeMin and MergeMax methods in ElementDefnMerger is enough to fix both #1824 and #1981. The additonal code in mergeTypeProfiles to set the Min/Max of the rebasedRootElem is no longer valid.

## Related issues
Resolves #1981 and #1824.

## Testing
Additional unit tests were added for CardinalityOfExtension.

### DIFF
--- a/src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefnMerger.cs
+++ b/src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefnMerger.cs
@@ -104,9 +104,8 @@ namespace Hl7.Fhir.Specification.Snapshot
                 // Mappings are cumulative, but keep unique on full contents
                 snap.Mapping = mergeCollection(snap.Mapping, diff.Mapping, (a, b) => a.IsExactly(b));
 
-                //snap.MinElement = mergePrimitiveAttribute(snap.MinElement, diff.MinElement);
+                // Note that max is not corrected when max < min! constrainMax could be used if that is desired.
                 snap.MinElement = mergeMin(snap.MinElement, diff.MinElement);
-                //snap.MaxElement = mergePrimitiveAttribute(snap.MaxElement, diff.MaxElement);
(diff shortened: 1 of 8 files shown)


src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefnMerger.cs
```diff
@@ -300,12 +299,31 @@ internal FhirString mergeMax(FhirString snap, FhirString diff)
...
                 }
-                else
-                    return snap;
+
+                return snap;
+            }
+
+            /// <summary>
+            /// Make sure max >= min
+            /// </summary>
```
marcovisserFurore, line 313: This method is not used (yet). Perhaps some extra documentation where it should be used.

marcovisserFurore (changes requested):
Very nice unittests. Thanks for that.

Some extra lines of code and we are good to go

src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefnMerger.cs
Rob5045: Done

marcovisserFurore (approved):

PR #2007, opened 2022-03-07
Add TestValidationPerformanceUsCore (R4)

alexzautke:


src/Hl7.Fhir.Specification.Tests/Validation/BasicValidationTests.cs
```diff
@@ -1390,6 +1390,49 @@ public void ValidateWithTargetProfileAndChildDefinitions()
...
+            {
+                var jsonBundle = await File.ReadAllTextAsync(Path.Combine("TestData", "us-core-instances", instance));
+                var parser = new FhirJsonParser();
+                var bundle = await parser.ParseAsync<Bundle>(jsonBundle);
+                Assert.NotNull(bundle);
+
+                var sw = new Stopwatch();
+                sw.Start();
+                var result = _validator.Validate(bundle);
+                sw.Stop();
```
cknaap, line 1429: "When a [Stopwatch](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch?view=net-6.0) instance measures more than one interval, the [Stop](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch.stop?view=net-6.0) method is equivalent to pausing the elapsed time measurement. A subsequent call to [Start](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch.start?view=net-6.0) resumes measuring time from the current elapsed time value. Use the [Reset](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch.reset?view=net-6.0) method to clear the cumulative elapsed time in a [Stopwatch](https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch?view=net-6.0) instance."
https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch.stop?view=net-6.0

src/Hl7.Fhir.Specification.Tests/Validation/BasicValidationTests.cs
```diff
@@ -1390,6 +1390,49 @@ public void ValidateWithTargetProfileAndChildDefinitions()
...
+                ResourceResolver = resolver,
+                GenerateSnapshot = true,
+                EnableXsdValidation = false,
+                Trace = false,
+                ResolveExternalReferences = false
+            };
+
+            _validator = new Validator(ctx);
+
+            var instances = new[] { "us-core-instance-1.json",
```
cknaap, line 1409: If you turn this into a theory, with the instances as data, the stopwatchtimings are also correct.
You would however need to have the CachedResolver in a fixture then, or it would keep generating snapshots all the time.

src/Hl7.Fhir.Specification.Tests/Validation/BasicValidationTests.cs
```diff
@@ -1390,6 +1390,49 @@ public void ValidateWithTargetProfileAndChildDefinitions()
...
+                                    "us-core-instance-200.json",
+                                    "us-core-instance-300.json",
+                                    "us-core-instance-500.json",
+                                    "us-core-instance-700.json"};
+
+            foreach(var instance in instances)
+            {
+                var jsonBundle = await File.ReadAllTextAsync(Path.Combine("TestData", "us-core-instances", instance));
+                var parser = new FhirJsonParser();
+                var bundle = await parser.ParseAsync<Bundle>(jsonBundle);
```
cknaap, line 1423: Interesting to see whether using ITypedElement directly on ISourceNode might be faster than on a POCO.

src/Hl7.Fhir.Specification.Tests/Validation/BasicValidationTests.cs
```diff
@@ -1390,6 +1390,49 @@ public void ValidateWithTargetProfileAndChildDefinitions()
...
 
+        [Fact]
+        public async void TestValidationPerformanceUsCore()
+        {
+            var resolver = new MultiResolver(new CachedResolver(new DirectorySource(Path.Combine("TestData", "us-core-profiles"))), new CachedResolver(new ZipSource("specification.zip")));
+
+            var ctx = new ValidationSettings()
+            {
+                ResourceResolver = resolver,
+                GenerateSnapshot = true,
```
cknaap, line 1401: Generating snapshots will slow down the first validations, as the CachedResolver won't have them cached yet.

We can look into using Benchmark.net for more accurate timings.

cknaap (changes requested):

alexzautke:
Please note that I'm not sure if the actual validation results are correct in some cases. Worth to check this as well.

ewoutkramer:
What shall we do with this PR?  It has helped tremendously to locate the performance problems in bundle validation - and will likely be of use in the future for the same purpose. Plus: this was - I think - quite some work to produce.  However, I don't want to pull it as it is an integration tests with about 90 files - too much to include forever in the SDK project. 

Would closing it + keeping the branch be a solution?  We can keep the branch for quite a while, but will we remember we still have it?

marcovisserFurore:
I am not in favor to pull these 90 files into the repository as well. Let's close this PR and keep the branch.

Issue #2033, opened 2022-04-04
Wrong Namespace for R4 Subsetted Tag


brianpos:
https://github.com/FirelyTeam/firely-net-sdk/blob/35fb12f9b486800cdb1455daad0093caaeff07ac/src/Hl7.Fhir.Core/Serialization/BaseFhirSerializer.cs#L76

This should be using http://terminology.hl7.org/CodeSystem/v3-ObservationValue
as per https://www.hl7.org/fhir/v3/ObservationValue/cs.html

Hoping this is all and it isn't somewhere in the common area too.

PR #2037, opened 2022-04-11
#2033 Wrong namespace for the SUBSETTED type

brianpos:
## Description
Update the System value for the SUBSETTED type for the MarkSubset routine that is used when marking a resource that was serialized with a subset flag

## Related issues
Resolves [issue #2033 ].

## Testing
The existing unit tests for formatting with _summary=true were updated to also check for the correct system value.


ewoutkramer:
@marcovisserFurore - I am find with this change, but it is indeed breaking - can we still include it with 4.0?  5.0 is another year out or so.

ewoutkramer (approved):

brianpos:
> @marcovisserFurore - I am fine with this change, but it is indeed breaking - can we still include it with 4.0? 5.0 is another year out or so.

I guess we could work out if there is a settings object that would be appropriate to include this on?
(so we can choose to opt in or out - as it stands, this isn't really FHIR conformant)

ewoutkramer:
Yeah, it's incorrect, so I think we should change it in 4.0.  We should also document it in the wiki (https://github.com/FirelyTeam/firely-net-sdk/wiki/Breaking-changes-in-4.0).

marcovisserFurore:
Well, we are about to release 4 very soon. This week or next week. We had already 2 betas for release 4 to get the community ready. So I am not very excited to add this change to the final release of 4. 
But I guess I have not so much choice... Indeed version 5 will be somewhere next year, and I am not in favor to switch from old behavior to new behavior by a setting object.
So I will try to do my best to add this to 4 release.

ewoutkramer:
> Indeed version 5 will be somewhere next year, and I am not in favor to switch from old behavior to new behavior by a setting object.

Yes, that's exactly what I thought too....

marcovisserFurore:
This PR is been replaced by PR #2047 because of a wrong base branch.

Issue #2029, opened 2022-03-30
HttpMessageHandler Passed To Constructor Of FhirClient Gets Disposed When FhirClient Is Disposed


AMehlem:
**Describe the bug**
We are using the firely-net-sdk in a project and are experiencing port exhaustion when the server-web-api using the sdk is under high load. To deal with the issue, I tried to reuse the HttpMessageHandler that gets passed into the constructor of the FhirClient. However, the HttpMessageHandler gets disposed with the HttpClientRequester, that is created in the FhirClient's constructor. 

**Expected behavior**
If I am not mistaken, a class instance passed into a constructor should in general not be disposed by the class using it. The BaseFhirClient should therefore only dispose the HttpMessageHandler if the class BaseFhirClient or the FhirClient created it. If the HttpMessageHandler was passed as a constructor parameter, the handler should not be disposed with the BaseFhirClient.

**Version used:**
 - FHIR Version: R4, maybe other versions are also affected
 - Version: 3.8.0

Kind regards,
Ashwani Mehlem

brianpos:
May also be related to #2036

brianpos:
Have confirmed this behaviour, and have a simple fix for it.

brianpos:
Along with this common module version
https://github.com/brianpos/firely-net-common/commit/69eb790f31826d6c08c16773c64421c2e0a74716

marcovisserFurore:
Good idea to use the boolean `disposeHandler` of the `HttpClient` constructor. It saves us a flag in the `FhirClient` to remember not dispose the injected handler.  

We will create a PR for this.

AMehlem:
Thanks @brianpos, your fix solves the problem with the port exhaustion as the message handler can then be reused.

As a side note: It would be even easier if the constructor of the class FhirClient would take an HttpClient or an IHttpClientFactory as a parameter instead of an HttpMessageHandler, as this would allow using the HttpClientFactory-Pattern (see https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-6.0) already mentioned in issue #2036. 
The main benefit of this is that the HttpClientFactory handles the lifetime of the message handlers, which avoids DNS-Issues (which can occur when the message handlers live too long) as well as port exhaustion and performance problems (which may occur when too many message handlers are created in a short period of time).

PR #2040, opened 2022-04-11
Do not dispose of a shared HttpMessageHandler passed into the constructor

brianpos:
## Description
When creating the new HttpClient inside the constructor indicate that if it is a shared MessageHandler and therefore that it should not be disposing it when it is finished with it.
As the change in the common lib has a default parameter this is not a truly breaking change, and in the R4 (version level) functionality the public surface has not changed this is not tagged as a breaking change.

## Related issues
Resolves [issue #2029].

## Testing
The unit test `CallsCallbacksHttpClient` was used to verify this functionality, as it fails when the fix is not applied.


marcovisserFurore (approved):
I will also make a PR for `develop-stu3`

brianpos:
This also requires that common layer change to be merged in too - just reminding.

PR #2048, opened 2022-04-20
Use IssueComponent.Expression instead of IssueComponent.Location

### LINKED ISSUE #1692: Validator reports error location in deprecated `location` element.
**Describe the bug**
The validator returns the location of an error in the `location` element of an `IssueComponent`, but this element is deprecated

**Expected behavior**
The validator should use the `expression` element for the location of the message, and preferable also set `location` to the same value for backwards compatibility.

**Version used:**
 - FHIR Version: All
 - Version: 3.x

marcovisserFurore:
## Description
 `IssueComponent.Location` is deprecated. We use now  `IssueComponent.Expression`

## Related issues
Resolves #1692.

## Testing
Existing tests

## See also
PR FirelyTeam/firely-net-common/pull/196

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Specification/Schema/Binding.cs
+++ b/src/Hl7.Fhir.Specification/Schema/Binding.cs
@@ -149,7 +149,12 @@ namespace Hl7.Fhir.Specification.Schema
 
 
                 var outcome = (await svc.ValueSetValidateCode(parameters).ConfigureAwait(false)).ToOperationOutcome();
-                foreach (var issue in outcome.Issue) issue.Expression = new string[] { location };
+                foreach (var issue in outcome.Issue)
+                {
+                    issue.Expression = new string[] { location };
+                    // Location is deprecated, but we set this for backwards compatibility
+                    issue.Location = new string[] { location };
+                }
                 return outcome;
             }
             catch (FhirOperationException tse)
--- a/src/Hl7.Fhir.Specification/Schema/Binding.cs
+++ b/src/Hl7.Fhir.Specification/Schema/Binding.cs
@@ -138,7 +138,7 @@ namespace Hl7.Fhir.Specification.Schema
                 Coding coding = null, CodeableConcept cc = null, bool? abstractAllowed = null, string context = null)
         {
             try
-            {                
+            {
                 var parameters = new ValidateCodeParameters()
                     .WithValueSet(canonical)
                     .WithCode(code: code, system: system, display: display, context: context)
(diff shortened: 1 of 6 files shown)


src/Hl7.Fhir.Specification/Schema/Binding.cs
```diff
@@ -149,14 +149,14 @@ internal async Task<OperationOutcome> ValidateCode(ITypedElement source, Element
 
 
                 var outcome = (await svc.ValueSetValidateCode(parameters).ConfigureAwait(false)).ToOperationOutcome();
-                foreach (var issue in outcome.Issue) issue.Location = new string[] { location };
+                foreach (var issue in outcome.Issue) issue.Expression = new string[] { location };
```
ewoutkramer, line 152: In common, you have set both for backwards compat - here we do not do that, so older code that still looks to Location will fail now, right?

src/Hl7.Fhir.Specification/Validation/FpConstraintValidationExtensions.cs
```diff
@@ -72,7 +72,7 @@ public static OperationOutcome ValidateFp(this Validator v, string structureDefi
                         Code = issue.Type,
                         Details = issue.ToCodeableConcept(text),
                         Diagnostics = constraintElement.GetFhirPathConstraint(), // Putting the fhirpath expression of the invariant in the diagnostics
-                        Location = new string[] { instance.Location }
+                        Expression = new string[] { instance.Location }
```
ewoutkramer, line 75: For backwards compat, we should set both Location and Expression here.

src/Hl7.Fhir.Specification/Validation/TypeRefValidationExtensions.cs
```diff
@@ -195,7 +195,7 @@ OperationOutcome validate()
                 // Prefix each path with the referring resource's path to keep the locations
                 // interpretable
                 foreach (var issue in childResult.Issue)
-                    issue.Location = issue.Location.Concat(new string[] { instance.Location });
+                    issue.Expression = issue.Expression.Concat(new string[] { instance.Location });
```
ewoutkramer, line 198: For backwards compat, we should set both Location and Expression here.

ewoutkramer (changes requested):
I think not all changes are backwards compatible (like you did in common).

ewoutkramer (approved):

Issue #2036, opened 2022-04-06
Introduce HttpClientFactory in HttpClient


ewoutkramer:
In (closed) issue #2032 , the discussion centered around re-using instances of the HttpClient.  Currently, every instance of the FhirClient creates an instance of the HttpClient, but Microsoft recommends having a single HttpClient per application - which means the FhirClient should re-use one single instance.

**Describe the solution you'd like**
@brianpos suggests using a class called `HttpClientFactory` but it is unclear to me how this solves the problem.

**Describe alternatives you've considered**

**Additional context**
See issue the discussion in #2032 for more info.

brianpos:
https://docs.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests?msclkid=580e3af7b5ee11ec8c6cb7bf16e81260
which also links back to https://www.aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/

brianpos:
Noting through those posts it makes the point that it's not the actual HttpClient that is the issue, its the HttpMessageHandler that is the problem, and that's the thing you need to hang onto and the Factory helps manage things better.

brianpos:
I think with these changes to include the HttpClient via the constructor, then we don't need to do the HttpClientHandlerFactory stuff, and if someone wants to use that stuff, they can and we don't need to be involved.
(and we get the other unit testing benefits too as I've mentioned in the PRs.)

vankampenp:
The HttpClientFactory creates and manages HttpClient instances and takes an available message handler from the pool. The HttpClient should be included in the constructor rather than the factory, so the consumer can reuse the same HttpClient.

brianpos:
@ewoutkramer,  Now that the 3.8.3 and 4 beta release have the httpclient in the constructor of the fhirclient, I think we can close this issue as it enables the HttpClientFactory to be used without change in this library

PR #2049, opened 2022-04-22
Support providing the HttpClient direct to the FhirClient

brianpos:
## Description
Support providing the HttpClient direct to the FhirClient instead of the HttpClientHandler (as another alternative)

This then supports creating unit tests as described here for webapi based projects
https://www.hanselman.com/blog/minimal-apis-in-net-6-but-where-are-the-unit-tests

## Related issues
Resolves/Contributes to #2046 #2036 (may be able to now push the HttpClientFactory work to the callers, and not be required in the FirelySDK at all)

## Testing
Additional unit test included (and also included in my server to verify that can unit test as shown)

### DIFF
--- a/src/Hl7.Fhir.Core/Rest/FhirClient.cs
+++ b/src/Hl7.Fhir.Core/Rest/FhirClient.cs
@@ -6,7 +6,6 @@
 * available at https://raw.githubusercontent.com/FirelyTeam/firely-net-sdk/master/LICENSE
 */
 
-using Hl7.Fhir.Model;
 using Hl7.Fhir.Serialization;
 using Hl7.Fhir.Specification;
 using System;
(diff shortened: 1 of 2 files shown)


src/Hl7.Fhir.Core/Rest/FhirClient.cs
```diff
@@ -19,32 +19,44 @@ namespace Hl7.Fhir.Rest
...
         /// <param name="endpoint">
         /// The URL of the server to connect to.<br/>
         /// If the trailing '/' is not present, then it will be appended automatically
         /// </param>
         /// <param name="settings"></param>
         /// <param name="messageHandler"></param>
+        /// <param name="httpClient"></param>
         /// <param name="provider"></param>
-        public FhirClient(Uri endpoint, FhirClientSettings settings = null, HttpMessageHandler messageHandler = null, IStructureDefinitionSummaryProvider provider = null) : base(endpoint, settings, provider)
+        public FhirClient(Uri endpoint, FhirClientSettings settings = null, HttpMessageHandler messageHandler = null, HttpClient httpClient = null, IStructureDefinitionSummaryProvider provider = null) : base(endpoint, settings, provider)
```
marcovisserFurore, line 41: I think it is better to create an overload of the constructor with only the `HttpClient` as a parameter and not both (`HttpMessageHandler` and  `HttpClient`). Then the check for on line 46 is also not needed.

```c#
public FhirClient(Uri endpoint, FhirClientSettings settings = null, HttpClient httpClient = null, IStructureDefinitionSummaryProvider provider = null) : base(endpoint, settings, provider)
{
 // etc.
}
```

Of couse you need the other overload with Uri as a string as well:
```c#
 public FhirClient(string endpoint, FhirClientSettings settings = null, HttpClient httpClient = null, IStructureDefinitionSummaryProvider provider = null)
            : this(new Uri(endpoint), settings, httpClient, provider)
        {
        }
```

marcovisserFurore (changes requested):
I am not sure we are going to pull this for the next release, but for now these are my comments:

src/Hl7.Fhir.Core/Rest/FhirClient.cs
brianpos: 2 constructors to 4 is why I did like this, but no drama, can switch it over pretty quick and get it back in.

brianpos:
Made the changes as requested @marcovisserFurore

src/Hl7.Fhir.Core/Rest/FhirClient.cs
```diff
@@ -38,19 +38,42 @@ public partial class FhirClient : BaseFhirClient
...
             // Expose default request headers to user.
             RequestHeaders = requester.Client.DefaultRequestHeaders;
         }
 
+        /// <summary>
+        /// Creates a new client using a default endpoint
+        /// If the endpoint does not end with a slash (/), it will be added.
+        /// </summary>
+        /// <remarks>
+        /// If the messageHandler, or httpClient is provided then it must be disposed by the caller
```
marcovisserFurore, line 59:
```suggestion
        /// If the httpClient is provided then it must be disposed by the caller
```
```suggestion
        /// If the messageHandler, or httpClient is provided then it must be disposed by the caller
```

src/Hl7.Fhir.Core/Rest/FhirClient.cs
```diff
@@ -38,19 +38,42 @@ public partial class FhirClient : BaseFhirClient
...
+        /// Only one of the messageHandler or httpClient can be provided
```
marcovisserFurore, line 60:
```suggestion
```
```suggestion
        /// Only one of the messageHandler or httpClient can be provided
```

marcovisserFurore, line 60:
```suggestion
```
```suggestion
        /// Only one of the messageHandler or httpClient can be provided
```

src/Hl7.Fhir.Core/Rest/FhirClient.cs
```diff
@@ -19,15 +18,16 @@ namespace Hl7.Fhir.Rest
...
+        //disables warning that OnBeforeRequest and OnAfterResponse are never used.
 #pragma warning disable CS0067
 
         /// <summary>
         /// Creates a new client using a default endpoint
         /// If the endpoint does not end with a slash (/), it will be added.
         /// </summary>
         /// <remarks>
-        /// If the messageHandler is provided then it must be disposed by the caller
+        /// If the messageHandler, or httpClient is provided then it must be disposed by the caller
```
marcovisserFurore, line 29:
```suggestion
        /// If the messageHandler is provided then it must be disposed by the caller
```
```suggestion
        /// If the messageHandler, or httpClient is provided then it must be disposed by the caller
```

marcovisserFurore, line 29:
```suggestion
        /// If the messageHandler is provided then it must be disposed by the caller
```
```suggestion
        /// If the messageHandler, or httpClient is provided then it must be disposed by the caller
```

src/Hl7.Fhir.Core/Rest/FhirClient.cs
```diff
@@ -19,15 +18,16 @@ namespace Hl7.Fhir.Rest
...
+        /// Only one of the messageHandler or httpClient can be provided
```
marcovisserFurore, line 30:
```suggestion

```
```suggestion
        /// Only one of the messageHandler or httpClient can be provided
```

marcovisserFurore, line 30:
```suggestion

```
```suggestion
        /// Only one of the messageHandler or httpClient can be provided
```

marcovisserFurore (changes requested):
Some documentation adjustments needed

src/Hl7.Fhir.Core/Rest/FhirClient.cs
brianpos: Updated the comments on the 4 constructors

marcovisserFurore (approved):

PR #2052, opened 2022-04-23
Add unit test to prove that type slices are not correctly constrained

alexzautke:
If the value[x] element is constrained to a single type, but the parent profile contains an explicit slice of a different type, that slice is copied to the snapshot as well and not removed.

This PR is based on PR #2051


marcovisserFurore (approved):

marcovisserFurore:
Thanks @alexzautke, I will move this PR to draft and plan it for the next release. We can use this branch to add the solution to it.

PR #2054, opened 2022-04-25
Handle and generate absolute contentRefernces

mmsmits:
Fixes #2054

### DIFF
--- a/src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
+++ b/src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
@@ -399,6 +399,7 @@ namespace Hl7.Fhir.Specification.Snapshot
                 // [WMR 20170424] Inherit existing base components, generate if missing
                 await ensureBaseComponents(snapshot.Element, structure.BaseDefinition, false).ConfigureAwait(false);
 
+
                 // [WMR 20170208] Moved to *AFTER* ensureBaseComponents - emits annotations...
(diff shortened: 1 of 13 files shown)


src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
```diff
@@ -890,6 +884,11 @@ private async T.Task mergeElement(ElementDefinitionNavigator snap, ElementDefini
                 // Now, recursively merge the children
                 await merge(snap, diff).ConfigureAwait(false);
 
+                // [MS 20220425] Make sure newly added contentReferences (from bases, types, and contentReferences) are absolute.
+                // Except when the it's a core profile (Derivation is not a Constraint)
+                if (snap.StructureDefinition.Derivation == StructureDefinition.TypeDerivationRule.Constraint)
+                    ensureAbsoluteContentReferences(snap, snap.StructureDefinition.BaseDefinition);
```
ewoutkramer, line 890: Why cannot we call `ensureAbsoluteContentReference()` with `defn.ContentReferenceElement` here directly?  Why using the navigate to the first child first? I think `defn` is already the right child here?

ewoutkramer, line 890: Using `BaseDefinition` will I think not work when we have profiles on profiles (since BaseDefinition is then the Url of the parent profile, not the original core profile). Michel has some code in this file that actually finds the base url of the core type, not of the direct parent!

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #2055, opened 2022-04-26
Support providing the HttpClient direct to the FhirClient - STU3

marcovisserFurore:
## Description
Support providing the `HttpClient` direct to the `FhirClient` instead of the `HttpClientHandler` (as another alternative)

This then supports creating unit tests as described here for webapi based projects
https://www.hanselman.com/blog/minimal-apis-in-net-6-but-where-are-the-unit-tests

## Related issues
Resolves/Contributes to #2046 #2036 (may be able to now push the `HttpClientFactory` work to the callers, and not be required in the FirelySDK at all)

## Testing
Additional unit test included (and also included in my server to verify that can unit test as shown)

## Credits
@brianpos

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/Rest/FhirClientTests.cs
+++ b/src/Hl7.Fhir.Core.Tests/Rest/FhirClientTests.cs
@@ -1383,7 +1383,7 @@ namespace Hl7.Fhir.Tests.Rest
 
         [TestMethod]
         [TestCategory("FhirClient"), TestCategory("IntegrationTest")]
-        public void CallsCallbacksHttpClient()
+        public void CallsCallbacksHttpClientHandler()
         {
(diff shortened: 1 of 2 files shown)


mmsmits (approved):

oldefezziwig:
@marcovisserFurore, @brianpos - if this overload of the FhirClient constructor is now included in 4.x of the fhir-net-sdk, are there plans to update the documentation to show how this can be done, i.e. somewhere in here perhaps? https://docs.fire.ly/projects/Firely-NET-SDK/client/setup.html#fhirclient-communication-options

Asking because the current documentation still shows the original way to instantiate a FhirClient (i.e. passing just the FHIR endpoint) unless I'm missing something and this is documented somewhere already?

Issue #2073, opened 2022-05-04
POST/PUT without body throws NullReferenceException


cdaven:
**Describe the bug**

When making a POST or PUT call without a resource through `FhirClient`, `HttpClientRequester.ExecuteAsync()` will throw `NullReferenceException`.

Below, `requestMessage.Content` is null and throws an exception:

```csharp
if (requestMessage.Method == HttpMethod.Post || requestMessage.Method == HttpMethod.Put)
{
    outgoingBody = await requestMessage.Content.ReadAsByteArrayAsync().ConfigureAwait(false);
}
```

**To Reproduce**

```csharp
var fhirClient = new FhirClient(endpoint, settings, messageHandler, provider);
await fhirClient.OperationAsync(fhirClient.Endpoint, "ping")
```

The Swedish EHM have defined a "ping" operation without resource, and claims that it should work with either POST or GET. See https://simplifier.net/swedishnationalmedicationlist/nllping

Also, `OperationAsync()` arguments have default values `parameters = null` and `useGet = false`, which leads to this exception.

**Expected behavior**

I would expect a POST or PUT without body to either be prohibited and a better exception being thrown, or that it should work without any exceptions. Probably the latter.

**Version used:**

 - FHIR Version: R4
 - Version: 3.8.3

ewoutkramer:
According to the W3C; an empty body on a POST is ok, so we should be able to handle it. I guess we can improve on the NullReferenceException - but what would be more appropriate?  It feels like an empty body should result in a "null" returned as received resource?  And what if the mediatype is not suggesting a FHIR payload?

cdaven:
If you agree that "an empty body on POST is ok", you don't need to throw any exceptions. Simply add a null check for `requestMessage.Content` in the `if` clause above. It shouldn't matter which content-type you're using if the content is null...?

PR #2102, opened 2022-05-25
Check for null reference when trying to use a FHIR operation without a body [stu3]

mmsmits:
fixes: #2073

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/Rest/FhirClientMockTests.cs
+++ b/src/Hl7.Fhir.Core.Tests/Rest/FhirClientMockTests.cs
@@ -296,7 +296,7 @@ namespace Hl7.Fhir.Core.Tests.Rest
 
             var parameters = await client.OperationAsync(new Uri("http://example.com/fhir/$ping")) as Parameters;
 
-            ((FhirString)parameters.Parameter.FirstOrDefault().Value).Value.Should().Be("connected");
+            parameters.Parameter.FirstOrDefault().Value.Should().BeOfType<FhirString>().Which.Value.Should().Be("connected");
 
         }
     }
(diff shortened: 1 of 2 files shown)


src/Hl7.Fhir.Core.Tests/Rest/FhirClientMockTests.cs
```diff
@@ -270,7 +270,33 @@ public async System.Threading.Tasks.Task TestUnauthorizedWithANonFhirJsonBody()
...
+                        "SendAsync",
+                        ItExpr.Is<HttpRequestMessage>(h => h.RequestUri == new Uri("http://example.com/fhir/$ping")),
+                        ItExpr.IsAny<CancellationToken>())
+                     .ReturnsAsync(response);
+
+            using var client = new FhirClient("http://example.com/fhir/", new FhirClientSettings { VerifyFhirVersion = false }, mock.Object);
+
+            var parameters = await client.OperationAsync(new Uri("http://example.com/fhir/$ping")) as Parameters;
 
+            ((FhirString)parameters.Parameter.FirstOrDefault().Value).Value.Should().Be("connected");
```
marcovisserFurore, line 299: To avoid the cast you can do this:
```c#
parameters.Parameter.FirstOrDefault().Value.Should().BeOfType<FhirString>().Which.Value.Should().Be("connected");
```

marcovisserFurore (approved):
See my comment, but can be merged when you leave it like this

src/Hl7.Fhir.Core.Tests/Rest/FhirClientMockTests.cs
mmsmits: Wow sweet

PR #2112, opened 2022-06-15
Initialization of empty element constraint sources can be skipped for snapshot generator output intended for the differential

### LINKED ISSUE #2111: If an element.constraint.source points to itself then the source should only be set for the snapshot, not the differential
**Source**
https://chat.fhir.org/#narrow/stream/179177-conformance/topic/STU3.20Qustionnaire.20snapshot.20generation/near/279683046

**Version used:**
 - FHIR Version: R4 and later
 - Version: 4.0.0

Rob5045:
## Description
Added SnapshotIntendedUse enum and IntendedUse property to SnapshotGeneratorSettings so that initialization of empty element constraint sources can be skipped for snapshot generator output intended for the differential.

## Related issues
Fixes #2111 

## Testing
Unit test ElementConstraintSourceTest was added.

## FirelyTeam Checklist


Rob5045:
After discussion with @mmsmits and @ewoutkramer this pull request is withdrawn and the issue is solved in Forge

PR #2113, opened 2022-06-16
Added IntendedUse property to SnapshotGeneratorSettings for consistency with other FHIR versions of the SDK.

Rob5045:
## Related issues
Relates to #2112 

## Testing
N/A

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGeneratorSettings.cs
+++ b/src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGeneratorSettings.cs
@@ -45,6 +45,7 @@ namespace Hl7.Fhir.Specification.Snapshot
             other.GenerateExtensionsOnConstraints = GenerateExtensionsOnConstraints;
             other.GenerateAnnotationsOnConstraints = GenerateAnnotationsOnConstraints;
             other.GenerateElementIds = GenerateElementIds;
+            other.IntendedUse = IntendedUse;
             // other.MergeTypeProfiles = MergeTypeProfiles;
         }
 
@@ -69,7 +70,7 @@ namespace Hl7.Fhir.Specification.Snapshot
         /// to elements and properties in the snapshot that are constrained by the differential with respect to the base profile.
         /// <br />
         /// Note that this extension only applies to the containing profile and should NOT be inherited by derived profiles.
-        /// The FHIR API snapshot generator explicitly removes and re-generates these extensions for each profile.
+        /// The FHIR SDK snapshot generator explicitly removes and re-generates these extensions for each profile.
         /// The <seealso cref="SnapshotGeneratorExtensions"/> class provides utility methods to read and/or remove the generated extensions.
         /// </summary>
         public bool GenerateExtensionsOnConstraints { get; set; } = false; // MarkChanges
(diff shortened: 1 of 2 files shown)


src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGeneratorSettings.cs
```diff
@@ -81,6 +82,13 @@ public void CopyTo(SnapshotGeneratorSettings other)
         /// <summary>Enable this setting to automatically generate missing element id values.</summary>
         public bool GenerateElementIds { get; set; } = true;
 
+        /// <summary>
+        /// Indicates the intended use of the snapshot generator output: snapshot or differential.
+        /// Snapshot generator business logic may vary depending on the intended use.
+        /// The default value is set to a backwards compatible setting (i.e. snapshot and differential).
+        /// </summary>
+        public SnapshotIntendedUse IntendedUse { get; set; } = SnapshotIntendedUse.BackwardsCompatible;
```
ewoutkramer, line 90: This requires a lot of context to understand (I did not read the discussions leading up to this). I guess that if I don't get it, no one will (except for you two ;-)).  What are the differences? Why would a snapshot generator touch the differential? (just the first questions that came up).  Is there a way to make this setting a bit more explict? Something like  naming it `FixDiscrepancyBetweenXAndIButYouShouldntWorryAboutThis`?  Is this really about intended use?

Ok, maybe I should now read the discussion leading up to this ;-)

Rob5045: There is a specific bug in R4 that needs fixing but we tried to make the solution as generic as possible. Any feedback is of course welcome!

ewoutkramer: It's just that this is part of the public interface, and there's no way to understand what it does based on the current public xmldoc comment.  Maybe expand on it by documenting the `enum` ?

Rob5045: How about:

```
    /// <summary>
    /// Indicates the intended use of the snapshot generator output.
    /// Snapshot generator business logic may vary depending on the intended use.
    /// For example: if an "element.constraint.source" references the resource itself,
    /// then the source value should be set for the snapshot but not for the differential.
    /// </summary>
    public enum SnapshotIntendedUse
    {
        /// <summary>
        /// Indicates that the snapshot generator output is intended to be used for the snapshot component of a structure definition.
        /// </summary>
        Snapshot = 1 << 0,
        /// <summary>
        /// Indicates that the snapshot generator output is intended to be used for the differential component of a structure definition.
        /// </summary>
        Differential = 1 << 1,
        /// <summary>
        /// This is the (default) backwards compatibility setting.
        /// </summary>
        BackwardsCompatible = Snapshot | Differential
    }
```

Rob5045: @ewoutkramer I can imagine that you don't like this solution but then we would have to restart the discussion with Grahame (https://chat.fhir.org/#narrow/stream/179177-conformance/topic/STU3.20Qustionnaire.20snapshot.20generation/near/279683046)

mmsmits: From what I understand, is that Forge uses the result of the SnapshotGenerator to fill missing (and incorrent?) elements in the differential. It then behaves differently then it would when it creating a snapshot.

ewoutkramer: Maybe that is better. I am careful about adding properties to the settings (they only tend to grow in sieze) that are of no use to the general public.  A carefully documented member (maybe even documented as "internal use only") may be better.

Rob5045: @ewoutkramer @mmsmits To correct for this in Forge is not hard to do either. Who is responsible for making this correction: Forge or the SDK?

ewoutkramer: The discussion on Zulip seems to be about snapshotting behaviour that others outside of Forge should be able to do, I think it is the responsibility of the SDK. But I might be wrong, I don't see the connection with this issue?

PR #2159, opened 2022-07-21
Validate codes against implicit valuesets in local terminology service

mmsmits:


src/Hl7.Fhir.Specification.Tests/Source/TerminologyTests.cs
```diff
@@ -948,6 +963,59 @@ public async Task<Resource> ResolveByCanonicalUriAsync(string uri)
             public Task<Resource> ResolveByUriAsync(string uri) => throw new NotImplementedException();
         }
 
+        private class OnlyCodeSystemResolver : IAsyncResourceResolver, IConformanceSource
```
ewoutkramer, line 966: I do this regularly too (creating test implementations), but shouldn't we use Moq for this?

ewoutkramer (approved):

src/Hl7.Fhir.Specification.Tests/Source/TerminologyTests.cs
mmsmits: Good point, let's try to use Moq for these kinds of things next time.

PR #2169, opened 2022-08-01
STU3: Remove binding after Type slicing and the slice does not contain a bindable type

### LINKED ISSUE #2132: Incorrectly added bindings on Boolean element in snapshot
Snapshots generated with Forge 29.0 (Firely .NET SDK 4.0.0) incorrectly adds a binding on `<element id="MedicationStatement.dosage.asNeeded[x]:asNeededBoolean">`

And those type slices get a binding from the slice intro, but an element of type boolean can not have a binding on it. 

Please see the attached resource and base resource. 

It is about the code here:
```
<element id="MedicationStatement.dosage.asNeeded[x]:asNeededBoolean">
      <path value="MedicationStatement.dosage.asNeeded[x]" />
      <sliceName value="asNeededBoolean" />
      <short value="Take &quot;as needed&quot; (for x)" />
      <definition value="Indicates whether the Medication is only taken when needed within a specific dosing schedule (Boolean option), or it indicates the precondition for taking the Medication (CodeableConcept)." />
      <comment value="Can express &quot;as needed&quot; without a reason by setting the Boolean = True.  In this case the CodeableConcept is not populated.  Or you can express &quot;as needed&quot; with a reason by including the CodeableConcept.  In this case the Boolean is assumed to be True.  If you set the Boolean to False, then the dose is given according to the schedule and is not &quot;prn&quot; or &quot;as needed&quot;." />
      <min value="0" />
      <max value="1" />
      <base>
        <path value="Dosage.asNeeded[x]" />
        <min value="0" />
        <max value="1" />
      </base>
      <type>
        <code value="boolean" />
      </type>
      <condition value="ele-1" />
      <constraint>
        <key value="ele-1" />
        <severity value="error" />
        <human value="All FHIR elements must have a @value or children" />
        <expression value="hasValue() or (children().count() &gt; id.count())" />
        <xpath value="@value|f:*|h:div" />
        <source value="http://hl7.org/fhir/StructureDefinition/Element" />
      </constraint>
      <isSummary value="true" />
      <binding>
        <extension url="http://hl7.org/fhir/StructureDefinition/elementdefinition-bindingName">
          <valueString value="MedicationAsNeededReason" />
        </extension>
        <strength value="example" />
        <description value="A coded concept identifying the precondition that should be met or evaluated prior to consuming or administering a medication dose.  For example &quot;pain&quot;, &quot;30 minutes prior to sexual intercourse&quot;, &quot;on flare-up&quot; etc." />
        <valueSet value="http://hl7.org/fhir/ValueSet/medication-as-needed-reason" />
      </binding>
      <mapping>
        <identity value="rim" />
        <map value="n/a" />
      </mapping>
      <mapping>
        <identity value="v2" />
        <map value="TQ1-9" />
      </mapping>
      <mapping>
        <identity value="rim" />
        <map value=".outboundRelationship[typeCode=PRCN].target[classCode=OBS, moodCode=EVN, code=&quot;as needed&quot;].value=boolean or codable concept" />
      </mapping>
    </element>
```

The java validator picks this up by throwing: 

```
`Error @ StructureDefinition.snapshot.element[121].binding: The element MedicationStatement.dosage.asNeeded[x] has a binding, but no bindable types are present [boolean]

  Error @ StructureDefinition.snapshot.element[24].binding: The element MedicationStatement.medication[x] has a binding, but no bindable types are present [Reference]

  Error @ StructureDefinition.snapshot.element[25].binding: The element MedicationStatement.medication[x] has a binding, but no bindable types are present [Reference]`
```

[KBV_PR_Base_MedicationStatement.zip](https://github.com/FirelyTeam/firely-net-sdk/files/9009786/KBV_PR_Base_MedicationStatement.zip)

marcovisserFurore:
## Description
Remove binding from the snapshot after Type slicing and the slice does not contain a bindable type.

## Related issues
Resolves #2132 .

## Testing
See unit test BindingRemovedAfterTypeSlicing()

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
+++ b/src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
@@ -1702,17 +1702,7 @@ namespace Hl7.Fhir.Specification.Snapshot
 
             result = snap.ReturnToBookmark(lastSlice);
             // Copy the original (unmerged) slice base element to snapshot
-            if (result)
-            {
-                result = snap.InsertAfter((ElementDefinition)sliceBase.Current.DeepCopy());
-
-                // snapshot has a binding, but not a bindable type
-                if (result && snap.Current.Binding is not null && !diff.Current.Type.Any(t => ModelInfo.IsBindable(t.Code)))
-                {
-                    // remove the binding then
-                    snap.Current.Binding = null;
-                }
-            }
+            result = snap.InsertAfter((ElementDefinition)sliceBase.Current.DeepCopy());
             // Recursively copy the original (unmerged) child elements, if necessary
             if (result && sliceBase.HasChildren)
             {
--- a/src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
+++ b/src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
@@ -1702,7 +1702,17 @@ namespace Hl7.Fhir.Specification.Snapshot
 
             result = snap.ReturnToBookmark(lastSlice);
             // Copy the original (unmerged) slice base element to snapshot
-            if (result) { result = snap.InsertAfter((ElementDefinition)sliceBase.Current.DeepCopy()); }
+            if (result)
+            {
+                result = snap.InsertAfter((ElementDefinition)sliceBase.Current.DeepCopy());
+
+                // snapshot has a binding, but not a bindable type
+                if (result && snap.Current.Binding is not null && !diff.Current.Type.Any(t => ModelInfo.IsBindable(t.Code)))
+                {
+                    // remove the binding then
+                    snap.Current.Binding = null;
+                }
+            }
             // Recursively copy the original (unmerged) child elements, if necessary
             if (result && sliceBase.HasChildren)
             {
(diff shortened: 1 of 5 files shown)


mmsmits:
I think you fixed to for a too specific case (only on type slicing). 
This issue already occurs when you just remove all bindable types from a choice type (just a constraint, not a slice).
I added the unit test to prove this.

src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
```diff
@@ -1702,7 +1702,17 @@ private void addSliceBase(ElementDefinitionNavigator snap, ElementDefinitionNavi
 
             result = snap.ReturnToBookmark(lastSlice);
             // Copy the original (unmerged) slice base element to snapshot
-            if (result) { result = snap.InsertAfter((ElementDefinition)sliceBase.Current.DeepCopy()); }
+            if (result)
```
mmsmits, line 1705: I think we need to fix this in an earlier stage (when merging probably) to also tackle the case where you just remove the bindable types without slicing.

mmsmits (changes requested):

src/Hl7.Fhir.Specification/Specification/Snapshot/SnapshotGenerator.cs
marcovisserFurore: Thanks for the unittest which proves this

mmsmits (approved):

PR #2181, opened 2022-08-03
[SnapshotGenerator] Fix invalid constraint sources [R4]

### LINKED ISSUE #2118: Snapshot generator fills constraint.source incorrectly
**Describe the bug**
When generating snapshots in Forge or with Firely Terminal the source of constraints are being filled with values of very different profiles.

**To Reproduce with Firely Terminal**
* Install core R4 package: `fhir install hl7.fhir.r4.core`
* Bake that package (ask FT to calculate the snapshots): `fhir bake --package hl7.fhir.r4.core`
* Open `StructureDefinition-Observation.json` from the core package in your cache (`~/.fhir/packages/hl7.fhir.r4.core#4.0.1/package/StructureDefinition-Observation.json`) and search for `ref-1`
* See that `ref-1` source is fileld with `http://hl7.org/fhir/StructureDefinition/Account`, while I would expect it to come from `http://hl7.org/fhir/StructureDefinition/Reference` (https://simplifier.net/packages/hl7.fhir.r4.core/4.0.1/files/81189/~xml)

```
{
                        "key": "ref-1",
                        "severity": "error",
                        "human": "SHALL have a contained resource if a local reference is provided",
                        "expression": "reference.startsWith('#').not() or (reference.substring(1).trace('url') in %rootResource.contained.id.trace('ids'))",
                        "xpath": "not(starts-with(f:reference/@value, '#')) or exists(ancestor::*[self::f:entry or self::f:parameter]/f:resource/f:*/f:contained/f:*[f:id/@value=substring-after(current()/f:reference/@value, '#')]|/*/f:contained/f:*[f:id/@value=substring-after(current()/f:reference/@value, '#')])",
                        "source": "http://hl7.org/fhir/StructureDefinition/Account"
                    }
```
* Note that the snapshot provided in the core package also has no mention of `ref-1`: https://simplifier.net/packages/hl7.fhir.r4.core/4.0.1/files/83132/~xml

FT/SDK version info:
```
❯ fhir --version
Firely Terminal 2.5.0-beta-7
FHIR 4.0.1
HL7.FHIR SDK 3.7.0
```

**To Reproduce with Forge**
* Open https://simplifier.net/acmebaseprofilesr4 locally in Forge 29.0.0 (SDK 4.0.0) (e.g. download Zip and open that folder with Forge)
* Open AcmeBaseObservationLab.StructureDefinition.xml
* In the Element Tree open subject and look at the subject element, see that Forge has updated the Constraint.source for `ref-1` without user interaction. You can see the same in the differential in the XML tab. Forge does this (incorrect) correction based on the snapshot generator.

image

**Version used:**
 - FHIR Version: R4
 - SDK Version: At least 3.7.0 - 4.0.0

**Additional context**
Internal slack conversation: https://firelyteam.slack.com/archives/CKXSWMDL6/p1655910480739109

mmsmits:
fixes #2118


ewoutkramer (commented):
I see you use `diff?.StructureDefinition?.Url ?? _stack.CurrentProfileUri` a lot.

Just curious.....when is StructureDefinition.Url ever null?  Will it ever use _stack.CurrentProfileUri?

mmsmits:
> I see you use `diff?.StructureDefinition?.Url ?? _stack.CurrentProfileUri` a lot.
> 
> Just curious.....when is StructureDefinition.Url ever null? Will it ever use _stack.CurrentProfileUri?
It will, sometimes the entire `StructureDefinition` property of an `ElementDefinitionNavigator` is null, so Url as well. That was the entire reason for the fallback to `_stack.CurrentProfileUri`

ewoutkramer (approved):

PR #2182, opened 2022-08-03
[SnapshotGenerator] Fix invalid constraint sources [STU3]

mmsmits:
fixes: 2118

Needed to backport some logic from R4 and add an nullable argument to a public function to fix this in STU3

### DIFF
--- a/src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefnMerger.cs
+++ b/src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefnMerger.cs
@@ -430,7 +430,7 @@ namespace Hl7.Fhir.Specification.Snapshot
                             onConstraint(mergedItem);
                         }
                     }
-                    if (!string.IsNullOrEmpty(source)
+                    if (!string.IsNullOrEmpty(source))
                     {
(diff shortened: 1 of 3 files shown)


ewoutkramer (commented):
See my comment on the R4 version ;-)

src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefnMerger.cs
```diff
@@ -429,10 +430,26 @@ internal static FhirString constrainMax(FhirString max, UnsignedInt minValue)
...
+                    if (!string.IsNullOrEmpty(source))
+                    {
+                        InitializeConstraintSource(result, source);
+                    }
+
                 }
                 return result;
             }
 
+            internal static void InitializeConstraintSource(IEnumerable<ElementDefinition.ConstraintComponent> constraints, string source)
```
ewoutkramer, line 442: Is it true this is not needed in the R4 version?

mmsmits: This is backported from R4 ;)

ewoutkramer (approved):

PR #2184, opened 2022-08-04
2105 - Improve error messaging when any/all validations fail

### LINKED ISSUE #2105: Unresolvable extension leads to invalid validation error
**Describe the bug**
Validating the the attached bundle will result in an OperationOutcome containing the following error (next to other errors):

```
        {
            "severity": "error",
            "code": "invalid",
            "details": {
                "coding": [
                    {
                        "system": "http://hl7.org/fhir/dotnet-api-operation-outcome",
                        "code": "1003"
                    }
                ],
                "text": "The declared type of the element (Group-Fhir4.0) is incompatible with that of the instance ('Patient-Fhir4.0')"
            },
            "location": [
                "Bundle.entry[0].resource[0]",
                "Bundle.entry[3].resource[0].subject[0]"
            ]
        },
```

If you remove all extensions from the Patient resource (first resource in the bundle), the validation will succeed. This indicates that the error does not have anything to do with the reference listed in the OperationOutcome.

**Expected behavior**
Error message should explain better why it's being reported or erroneous error message is removed from the OperationOutcome. 

**Screenshots**

**Version used:**
 - FHIR Version: [e.g.R4]
 - Version: [e.g. 3.x]

ewoutkramer:
Tried to improve the error messages when lists of candidate profiles (ANY validation) fails.

Resolves #2105


src/Hl7.Fhir.Specification.Tests/OutcomeReportingTest.cs
```diff
@@ -53,15 +53,15 @@ public void Setup()
         private OperationOutcome _report;
         private string _location;
 
-        [TestMethod]
+        [TestMethod, Ignore("The way we used hierarchy was broken, piloting another strategy now")]
```
marcovisserFurore, line 56: If the test is broken should we not fix it? Or otherwise remove it, when it is not applicable anymore?

marcovisserFurore (changes requested):
See my comments and the branch could be merged because of conflicts

marcovisserFurore (approved):

Issue #2185, opened 2022-08-04
Allign invariant FHIRPath patches with Java Validator


mmsmits:
The Java validator patches some incorrect FHIRPath expressions on the fly during validations.
Let's align our validators a little bit more to also correct the same expressions.

Java patches are here:
https://github.com/hapifhir/org.hl7.fhir.core/blob/master/org.hl7.fhir.validation/src/main/java/org/hl7/fhir/validation/instance/utils/FHIRPathExpressionFixer.java

ewoutkramer:
Definitely connected with #2186 I guess ;-)

ewoutkramer:
Funny, we are now also employing (internally) an IResourceResolver that patches the structuredefinitions - but its not part of the SDK yet.  Should be make it public?

marcovisserFurore:
We should create also unittest for these corrections. We could use this construction:

```c#
        [Fact]
        public async T.Task ValidateInvariantCorrections()
        {
            await invariantValidation(FHIRAllTypes.ElementDefinition, "eld-16", new ElementDefinition { SliceName = "ASlice" });
            await invariantValidation(FHIRAllTypes.StructureDefinition, "sdf-0", new StructureDefinition { Name = "AName" });
        }

        private async T.Task invariantValidation(FHIRAllTypes type, string key, Base poco)
        {
            var sd = await _asyncSource.FindStructureDefinitionForCoreTypeAsync(type);
            var expression = sd.Snapshot.Element
                .SelectMany(elem => elem.Constraint)
                .SingleOrDefault(ce => ce.Key == key)?.Expression
                .Should().BeOfType<string>().Subject;

            poco.Predicate(expression).Should().BeTrue();
        }
```

mmsmits:
We already do this right?

PR #2186, opened 2022-08-04
Repair some invariants during validation

### LINKED ISSUE #2013: Patch FP constraints because of wrong interpretation of `matches()`
**Describe the bug**
This is not a bug per se, but as a result of a recent discussion on how the FP function `matches()` work, it's become clear that a few FP constraints in the HL7 FHIR spec are incorrect. To be exact: `matches()` will match any *substring* of a string with a pattern, while both Grahame and I had assumed this will match the *whole* string. This in turn has led to a few incorrect FP constraints being present in the spec.

**To Reproduce**
In fact, the .NET SDK has implemented `matches()` correctly (as the .NET Regex matches does the "right" thing), it's just that some constraints have assumed that the regex matches the whole string, rather than a substring.

At this moment the constraints `sdf-0`, `eld-19` and `eld-20` are in error.

**Expected behavior**
To fix this, we need to wrap a small number of constraints in `^` and `$` to make them work correctly. This will ensure the constraints match the WHOLE string, not a substring.

**Additional context**
See https://chat.fhir.org/#narrow/stream/179266-fhirpath/topic/Matches.20question for the full discussion.

### LINKED ISSUE #2153: Manually correct Invariants ref-1, sdf-24 and sdf-25
**Describe the bug**
The invariants `sdf-24` and `sdf-25` are causing problems because of negative values. This should be manually corrected in the validator (and FhirPath?)

See also the HL7 issue [FHIR-37761](https://jira.hl7.org/browse/FHIR-37761)

**Version used:**
 - FHIR Version: R4B
 - Version: 4.1.0

### LINKED ISSUE #2174: Questionnaire validation failing constraint que-7 "If the operator is 'exists', the value must be a boolean"
**Describe the bug**
When validating a Questionnaire resource containing an item that uses the 'Enabled When' feature with operator 'exists', validation always fails giving the following error:
> [ERROR] Instance failed constraint que-7 "If the operator is 'exists', the value must be a boolean"(further diagnostics: operator = 'exists' implies (answer is Boolean)) (at ...)

**To Reproduce**
Create the following minimal Questionnaire resource:
```
var instance = new Questionnaire
{
    Status = PublicationStatus.Unknown,
    Item = new List<Questionnaire.ItemComponent>
    {
        new Questionnaire.ItemComponent
        {
            LinkId = "1",
            Type = Questionnaire.QuestionnaireItemType.Text,
            EnableWhen = new List<Questionnaire.EnableWhenComponent>
            {
                new Questionnaire.EnableWhenComponent
                {
                    Question = "0",
                    Operator = Questionnaire.QuestionnaireItemOperator.Exists,
                    Answer = new FhirBoolean(true)
                }
            }
        }
    }
};
```

Validate using the following snippet:
```
var resolver = new CachedResolver(
    new MultiResolver(
        ZipSource.CreateValidationSource(System.IO.Path.Combine(System.IO.Path.GetDirectoryName(typeof(Program).Assembly.Location), "specification.zip"))));
var validator = new Validator(new ValidationSettings
{
    ResourceResolver = resolver,
});
var outcome = validator.Validate(instance);
```

**Expected behavior**
Validation should pass constrint que-7, since the Answer element contains an instance of FhirBoolean.

**Version used:**
 - FHIR Version: R4
 - Version: 4.1.0

marcovisserFurore:
## Description
The following invariants are corrected during validation:
- `eld-19`: changed the expression in `matches()` so that the *whole* string is matched 
- `eld-20`: changed the expression in `matches()` so that the *whole* string is matched
- `sdf-0`: changed the expression in `matches()` so that the *whole* string is matched
- `sdf-24`: do not use `$this` in the expression, see also https://jira.hl7.org/browse/FHIR-37761
- `sdf-25`: do not use `$this` in the expression, see also https://jira.hl7.org/browse/FHIR-37761
- `ref-1`: Reference with only a display should succeed. See https://jira.hl7.org/browse/FHIR-37793
- `que-7`: use the correct FHIR boolean, see also https://jira.hl7.org/browse/FHIR-25390

## Related issues
Resolves #2013 
Resolves #2153 
Resolves #2174 

## Testing
Unfortunately no unit tests, because I could not find an easy way to test this.

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Specification/Validation/FpConstraintValidationExtensions.cs
+++ b/src/Hl7.Fhir.Specification/Validation/FpConstraintValidationExtensions.cs
@@ -38,12 +38,6 @@ namespace Hl7.Fhir.Validation
                 // of FP up, which could do comparisons between quantities.
                 if (v.Settings.ConstraintsToIgnore.Contains(constraintElement.Key)) continue;
 
-                // This constraint got repaired in R4B - pre-apply it for other R3+ here as well.
-                if (constraintElement.Key == "ref-1" && constraintElement.Expression == "reference.startsWith('#').not() or (reference.substring(1).trace('url') in %resource.contained.id.trace('ids'))")
-                {
-                    constraintElement.Expression = "reference.startsWith('#').not() or (reference.substring(1).trace('url') in %rootResource.contained.id.trace('ids'))";
-                }
-
                 // The following constraints will be repaired in R4B - pre-apply it for other R3+ here as well.
                 constraintElement.Expression = constraintElement switch
                 {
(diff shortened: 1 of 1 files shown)


mmsmits (approved):
I miss a unit test, but I understand that's rather hard for these patches

marcovisserFurore:
Correct review. I have now idea to do this. I will add this for issue #2185

Issue #2211, opened 2022-08-26
Internal error in snapshot generator when opening profile


Rob5045:
**Describe the bug**
When generating a snapshot an internal error is thrown.

**To Reproduce**
1. Download project for profile that cannot be opened:
https://simplifier.net/Geboortezorg-STU3/bc-Birth
2. Open Forge 29.1 for STU3
3. Open the profile

**Expected behavior**
Profile can be opened

**Screenshots**
![image](https://user-images.githubusercontent.com/81365724/186925392-8ae7766f-3632-4622-bb02-fab8f47457b3.png)

**Version used:**
 - FHIR Version: STU3
 - Version: 4.2.0 & 4.2.1

**Additional context**
The fix for https://github.com/FirelyTeam/firely-net-sdk/issues/2053 seems to have introduced this error. The new logic introduces conflicting actions:

![image](https://user-images.githubusercontent.com/81365724/186926592-a852de43-e5c2-4a6b-bce4-b3f82dbffa4f.png)

mmsmits:
Ah yeah, introducing a remove option was bound to give these kinds of errors. 
Thanks for reporting!

PR #2220, opened 2022-09-02
Do removals in snapshot generation after all other actions. [STU3]

mmsmits:
fixes: #2211


src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
```diff
@@ -8267,5 +8267,138 @@ private void logExtensions(string title, IEnumerable<Extension> extensions, int
                     Debug.WriteLine($"{new string(' ', level * 3)}{extension.Url} : {extension.Value}");
             }
         }
+
+        [TestMethod]
+        public async T.Task TestMergingAPreviouslyRemovedElement()
+        {
+            var baseStructureDefinition = new StructureDefinition()
```
marcovisserFurore, line 8274: To prevent large unit test files, it would be better to add these profiles to the folder TestData and use them here. You can use then the already existing `_testResolver` as well.

marcovisserFurore (changes requested):

marcovisserFurore (approved):

PR #2221, opened 2022-09-02
[SnapshotGenerator] Do removals in snapshot generation after all other actions. [R4]

mmsmits:
fixes: #2211

### DIFF
--- a/src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
+++ b/src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
@@ -9787,128 +9787,20 @@ namespace Hl7.Fhir.Specification.Tests
             }
         }
 
+        //Tests Github issue #2211, see TestData/Issue-2211 for test artifacts.
         [TestMethod]
         public async T.Task TestMergingAPreviouslyRemovedElement()
         {
-            var baseStructureDefinition = new StructureDefinition()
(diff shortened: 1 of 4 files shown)


src/Hl7.Fhir.Specification.Tests/Snapshot/SnapshotGeneratorTest.cs
```diff
@@ -9786,5 +9786,129 @@ private void logExtensions(string title, IEnumerable<Extension> extensions, int
                     Debug.WriteLine($"{new string(' ', level * 3)}{extension.Url} : {extension.Value}");
             }
         }
+
+        [TestMethod]
+        public async T.Task TestMergingAPreviouslyRemovedElement()
+        {
+            var baseStructureDefinition = new StructureDefinition()
+            {
+                Url = "http://fire.ly/fhir/StructureDefiniton/ObservationBaseLimitedChoiceTypes",
```
marcovisserFurore, line 9795: To prevent large unit test files, it would be better to add these profiles to the folder TestData and use them here.

marcovisserFurore: You can use then the already existing `_testResolver` as well.

marcovisserFurore (changes requested):
Can you move the profiles to the TestData as resources?

marcovisserFurore (approved):

PR #2225, opened 2022-09-06
Bugfix/2106 terminologyserviceexpand

### LINKED ISSUE #2106: FhirClient TerminologyServiceExtensions Expand incorrect parameter
**Describe the bug**
The fhir client's ExpandValueSet extension method uses the STU3 parameter `idenfitier` and not the R4 parameter name `url`
(several servers have some backward compatability code to handle the old parameter name - ontoserver doesn't)
https://github.com/FirelyTeam/firely-net-sdk/blob/38e47ed37b98c0bc2b7479cd61c469101c4c3928/src/Hl7.Fhir.Core/Rest/FhirClientTermSvcExtensions.cs#L99
http://hl7.org/fhir/r4/valueset-operation-expand.html

**To Reproduce**
```
var server = new FhirClient("https://r4.ontoserver.csiro.au/fhir");
server.ExpandValueSet(new FhirUri("http://hl7.org/fhir/ValueSet/administrative-gender"));
```

**Expected behavior**
This should return the set of administrative genders, however reports that the `url` parameter has not been provided.

**Version used:**
 - FHIR Version: R4+
 - Version: 3.8.3

**Additional context**

brianpos:
## Description
Update the parameter to be conformant with the R4+ specification
Update the URL of the CSIRO test server to use the R4 server, not STU3 endpoint.

_(Not sure if you'd mark this as a breaking change, but it's now conformant with the R4 spec, would have been working with many terminology servers as they've been tolerant of the old parameter name)_

## Related issues
Resolves #2106

## Testing
Unit test InvokeExpandExistingValueSet updated to also cover this code path.

### DIFF
--- a/src/Hl7.Fhir.Core.Tests/Rest/OperationsTests.cs
+++ b/src/Hl7.Fhir.Core.Tests/Rest/OperationsTests.cs
@@ -1,7 +1,7 @@
-﻿/* 
+﻿/*
  * Copyright (c) 2014, Firely (info@fire.ly) and contributors
  * See the file CONTRIBUTORS for details.
- * 
+ *
  * This file is licensed under the BSD 3-Clause license
  * available at https://raw.githubusercontent.com/FirelyTeam/firely-net-sdk/master/LICENSE
  */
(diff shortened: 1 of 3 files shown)


ewoutkramer:
It is certainly a change in behaviour of the client towards the server - but it was really a bug on our side. So I think we don't have to wait for a major release. @marcovisserFurore  - what do you think?

ewoutkramer (dismissed):

marcovisserFurore:
I think this PR not a breaking change, because the external terminology servers are tolerant with the old parameter, and should also support the new `url` parameter. So let's pull it for the next release.

marcovisserFurore (approved):

Issue #2190, opened 2022-08-08
ResourceIdentity error with reference XXX/_history/1


marcovisserFurore:
When creating a new ResourceIdentity based on `XXX/_history/1`, the SDK throws a `System.IndexOutOfRangeException`, because the `_history` component isn't in the correct place (or either the resource type or version are missing). 

Expected behaviour: Don't crash, but return a ResourceIdentity without any components (because it's an invalid identity).

PR #2227, opened 2022-09-06
Fix index out of bounds error when trying to parse illegal history resource identifier

mmsmits:
fixes: #2190


src/Hl7.Fhir.Core.Tests/Rest/TestResourceIdentifier.cs
```diff
@@ -249,7 +249,7 @@ public void TestIsResourceId()
             Assert.IsFalse(ResourceIdentity.IsRestResourceIdentity("http://nu.nl/fhir/crap"));
             Assert.IsFalse(ResourceIdentity.IsRestResourceIdentity("x/y/z/4"));
             Assert.IsFalse(ResourceIdentity.IsRestResourceIdentity("urn:oid:1.2.3.4.5"));
+            Assert.IsFalse(ResourceIdentity.IsRestResourceIdentity("XXX/_history/1"));
```
marcovisserFurore, line 252: Unfortunately, this unit test does not test the parsing. It only checks if the uri matches the pattern. 
Please add an new testmethod like `TestBuild`.

marcovisserFurore (approved):

Issue #2183, opened 2022-08-03
Validator ignores valueString for value type sliced to be Quantity


njnikodym:
**Describe the bug**
The validator does not disallow providing a valueString for a value that is type sliced to be a Quantity

**To Reproduce**
The STU3 profile http://nictiz.nl/fhir/StructureDefinition/zib-BloodPressure ([nictiz.fhir.nl.stu3.zib2017 | HCIM BloodPressure - SIMPLIFIER.NET](https://simplifier.net/packages/nictiz.fhir.nl.stu3.zib2017/2.1.8/files/440622/~overview)) has sliced Observation.component, where the average blood pressure value is type sliced to be a Quantity:

See successful validation of https://simplifier.net/snippet/wardweistra/37 

**Expected behavior**
Validator giving back an error

**Screenshots**
![image](https://user-images.githubusercontent.com/68131234/182647746-291d9dc4-92bd-44d8-8110-8b6dccd53cbe.png)
![image](https://user-images.githubusercontent.com/68131234/182647769-4a2af26f-5e86-4739-abd1-ca9128b744ff.png)

**Version used:**
 - FHIR Version: STU3

njnikodym:
@mmsmits @ewoutkramer Can I update the customer that this issue is resolved?

PR #2236, opened 2022-09-13
[Validator] Also validate the instance type when the definition has children

mmsmits:
fixes: #2183

### DIFF
--- a/src/Hl7.Fhir.Specification/Validation/TypeRefValidationExtensions.cs
+++ b/src/Hl7.Fhir.Specification/Validation/TypeRefValidationExtensions.cs
@@ -19,7 +19,7 @@ namespace Hl7.Fhir.Validation
 {
     internal static class TypeRefValidationExtensions
     {
-        internal static OperationOutcome ValidateType(this Validator validator, ElementDefinition definition, ScopedNode instance, ValidationState state)
+        internal static OperationOutcome ValidateType(this Validator validator, ElementDefinition definition, ScopedNode instance, ValidationState state, bool validateProfiles)
         {
             var outcome = new OperationOutcome();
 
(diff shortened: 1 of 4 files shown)


src/Hl7.Fhir.Specification/Validation/TypeRefValidationExtensions.cs
```diff
@@ -32,46 +32,63 @@ internal static OperationOutcome ValidateType(this Validator validator, ElementD
...
                         else
                         {
                             var choiceList = String.Join(",", choices.Select(t => "'" + t + "'"));
                             validator.Trace(outcome, $"Type specified in the instance ('{instance.InstanceType}') is not one of the allowed choices ({choiceList})",
-                                     Issue.CONTENT_ELEMENT_HAS_INCORRECT_TYPE, instance);
+                                        Issue.CONTENT_ELEMENT_HAS_INCORRECT_TYPE, instance);
                         }
                     }
                     else
+                    {
```
ewoutkramer, line 65: This block should move to line 41 (so just under `var instanceType = ModelIInfo.FhirTypeNameToFhirType(..)` : it's applicable to both choice and non-choice elements.

src/Hl7.Fhir.Specification/Validation/TypeRefValidationExtensions.cs
```diff
@@ -32,46 +32,63 @@ internal static OperationOutcome ValidateType(this Validator validator, ElementD
...
+                }
+
+
             }
-            else if (choices.Count() == 1)
+            else
             {
-                // Only one type present in list of typerefs, all of the typerefs are candidates
-                outcome.Include(validator.ValidateTypeReferences(typeRefs, instance, state));
+                validator.Trace(outcome, "ElementDefinition is a choice or contains a polymorphic type constraint, but the instance does not indicate its actual type",
```
ewoutkramer, line 90: This should not be a reachable code path when choices.Count() = 1.

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #2269, opened 2022-10-14
Big common base (can be merged)

marcovisserFurore:
## Description
First PR to move some code to a FHIR agnostic library. 

In this PR:
- The existing `Hl7.Fhir.R5` project is renamed to `Hl7.Fhir.Core` (PackageId) and all Fhir non-agnostic code has been removed to satellite projects, like  `Hl7.Fhir.[R4|R4B|R5].Core`.
- In the satellite projects only `Model` and shim classes remain.
- All the satellite projects contain a directory `Specification\data`, which is needed to create the `specification.zip`.
- The constructor for `BaseFhirClient` requires a `ModelInspector` argument, which is provided by the derived `FhirClient`.
- Almost all the unit tests use the FhirPackageSource (instead of the `specification.zip`)
  
Todo:
- Refactor `ModelInfo_new`
- Get rid of `IConformanceResource` in order to move `ConceptMap` and `NamingSystem` to the satellite projects
- Get the unittests working again
- Move the poco's to support.Poco


src/Hl7.Fhir.Core.Tests/ElementModel/PocoTypedElementTests.cs
```diff
@@ -21,7 +21,7 @@ public class PocoTypedElementTests
         public void PocoTypedElementPocoRoundtrip()
         {
             var patient = new Patient();
-            var actual = patient.ToTypedElement().ToPoco<Patient>();
+            var actual = patient.ToTypedElement().ToPoco<Patient>(ModelInfo.ModelInspector);
```
ewoutkramer, line 24: Isn't this ToPoco<T> a shim that can do this `ModelInfo.ModelInspector` itself?  I think you showed me that yesterday.

src/Hl7.Fhir.Core.Tests/Rest/RequesterTests.cs
```diff
@@ -280,7 +279,7 @@ public async Tasks.Task TestBundleToEntryRequest()
             };
             bundleComponent.AddAnnotation(InteractionType.Search);
 
-            var entryRequest = await bundleComponent.ToEntryRequestAsync(_Settings);
+            var entryRequest = await bundleComponent.ToEntryRequestAsync(_Settings, ModelInfo.Version);
```
ewoutkramer, line 282: What was version specific about `ToEntryRequestAsync` (I guess I will see later ;-))

src/Hl7.Fhir.Core/Model/ResourceNames.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Hl7.Fhir.Model
+{
+    internal class ResourceNames
```
ewoutkramer, line 3: These aren't all resources - FhirTypeNames is maybe a better name?

src/Hl7.Fhir.Core/Rest/BaseFhirClient.cs
```diff
@@ -1111,12 +1114,41 @@ private void verifyServerVersion()
...
-            else if (!ModelInfo.CheckMinorVersionCompatibility(serverVersion))
+            else if (!CheckMinorVersionCompatibility(_fhirVersion, serverVersion))
             {
-                throw Error.NotSupported($"This client supports FHIR version {ModelInfo.Version} but the server uses version {serverVersion}");
+                throw Error.NotSupported($"This client supports FHIR version {_fhirVersion} but the server uses version {serverVersion}");
             }
 
         }
+
+        private static bool CheckMinorVersionCompatibility(string version, string externalVersion)
```
ewoutkramer, line 1124: I would not mind keeping this function as a public utility method (somewhere outside the current ModelInfo of course).

src/Hl7.Fhir.Core/Rest/TypedEntryResponseToBundle.cs
```diff
@@ -56,11 +56,11 @@ public static Bundle.EntryComponent ToBundleEntry(this TypedEntryResponse entry,
                 }
                 else
                 {
-                    if(entry.TypedElement != null)
+                    if (entry.TypedElement != null)
                     {
                         try
                         {
-                            result.Resource = new BaseFhirParser(parserSettings).Parse<Resource>(entry.TypedElement);
+                            result.Resource = new BaseFhirParser(parserSettings).Parse<Resource>(entry.TypedElement, inspector);
```
ewoutkramer, line 63: I think we should introduce a protected `Parse()` method on the base fhir client - since that now already has the inspector, so we don't need to pass it around here. Also, by centralizing the Parse() in the base, we can more easily swap out different parsers/serializers (thinking ahead to using the other parsers here).

src/Hl7.Fhir.Core/Serialization/BaseFhirParser.cs
```diff
@@ -32,9 +33,10 @@ public BaseFhirParser(ParserSettings settings = null)
 #pragma warning restore CS0618 // Type or member is obsolete
             };
 
-        public Base Parse(ITypedElement element) => element.ToPoco(buildPocoBuilderSettings(Settings));
+        // TODO BIG_COMMON
+        //public Base Parse(ITypedElement element) => element.ToPoco(buildPocoBuilderSettings(Settings));
 
-        public T Parse<T>(ITypedElement element) where T : Base => element.ToPoco<T>(buildPocoBuilderSettings(Settings));
+        public T Parse<T>(ITypedElement element, ModelInspector inspector) where T : Base => element.ToPoco<T>(inspector, buildPocoBuilderSettings(Settings));
```
ewoutkramer, line 39: Shouldn't the inspector be a private member on BaseFhirParser, initialized by the constructor?

src/Hl7.Fhir.Core/Serialization/JsonSerializationOptionsExtensions.cs
```diff
@@ -25,26 +25,26 @@ public static class JsonSerializerOptionsExtensions
         /// <summary>
         /// Initialize the options to serialize using the JsonFhirConverter, producing compact output without whitespace.
         /// </summary>
-        public static JsonSerializerOptions ForFhir(this JsonSerializerOptions options) => options.ForFhir(typeof(ModelInfo).Assembly);
+        public static JsonSerializerOptions ForFhir(this JsonSerializerOptions options) => options.ForFhir(typeof(ModelInfoNEW).Assembly);
```
ewoutkramer, line 28: I think these methods will have to become shims.

src/Hl7.Fhir.R5.Core/ElementModel/PocoBuilderExtensions.cs
```diff
@@ -0,0 +1,20 @@
...
+ * available at https://github.com/FirelyTeam/firely-net-sdk/blob/master/LICENSE
+ */
+
+
+using Hl7.Fhir.Model;
+using Hl7.Fhir.Serialization;
+
+namespace Hl7.Fhir.ElementModel
+{
+    public static class PocoBuilderExtensions
```
ewoutkramer, line 15: Why is this shim only present in the R5 project and not in the others?

src/Hl7.Fhir.R5.Core/ElementModel/PocoBuilderExtensions.cs
```diff
@@ -0,0 +1,20 @@
+﻿/* 
+ * Copyright (c) 2018, Firely (info@fire.ly) and contributors
```
ewoutkramer, line 2: R5.Core contains quite a few files that are not present in the others (HumanName.cs, IConformaceResource.cs, Initializer.cs, Questionnaire.cs), I cannot quickly find out why this is?

src/Hl7.Fhir.Specification/Specification/Source/WebResolver.cs
```diff
@@ -24,25 +24,25 @@ public class WebResolver : IResourceResolver, IAsyncResourceResolver
...
-        /// <summary>Create a new <see cref="WebResolver"/> instance that supports a custom <see cref="FhirClient"/> implementation.</summary>
+        /// <summary>Create a new <see cref="WebResolver"/> instance that supports a custom <see cref="BaseFhirClient"/> implementation.</summary>
         /// <param name="fhirClientFactory">
-        /// Factory function that should create a new <see cref="FhirClient"/> instance for the specified <see cref="Uri"/>.
-        /// If this parameter equals <c>null</c>, then the new instance creates a default <see cref="FhirClient"/> instance.
+        /// Factory function that should create a new <see cref="BaseFhirClient"/> instance for the specified <see cref="Uri"/>.
+        /// If this parameter equals <c>null</c>, then the new instance creates a default <see cref="BaseFhirClient"/> instance.
         /// </param>
-        public WebResolver(Func<Uri, FhirClient> fhirClientFactory)
+        public WebResolver(Func<Uri, BaseFhirClient> fhirClientFactory)
```
ewoutkramer, line 37: This is a change in the public interface.

src/Hl7.Fhir.Core/Model/ResourceNames.cs
marcovisserFurore: I agree

src/Hl7.Fhir.R5.Core/ElementModel/PocoBuilderExtensions.cs
marcovisserFurore: Because this shim is linked to the other projects (R4 and R4B). In this way we don't have to copy this shims to the other satellite projects.

src/Hl7.Fhir.Specification/Specification/Source/WebResolver.cs
marcovisserFurore: Yes indeed. We can also decide to create a common `WebResolver` and move this `WebResolver` as a shim back to the satellites.

src/Hl7.Fhir.R5.Core/ElementModel/PocoBuilderExtensions.cs
ewoutkramer: I should have known that, you showed me ;-)

src/Hl7.Fhir.Core/Rest/TypedEntryResponseToBundle.cs
marcovisserFurore: Should we not make something injectable? Like an extra parameter in the `BaseFhirClient`? I investigated this a bit and we need the following:
- parsing:
  - json string => Resource
  - xml string => Resource
- serializing:
  - Resource => json string => byte[]
  - Resource => xml string => byte[]

ewoutkramer: Yes, I think so too.

src/Hl7.Fhir.Core.Tests/ElementModel/PocoTypedElementTests.cs
marcovisserFurore: Solved

src/Hl7.Fhir.Core.Tests/Rest/RequesterTests.cs
marcovisserFurore: It is used for verifying the FhirVersion during communication.

src/Hl7.Fhir.Core/Rest/BaseFhirClient.cs
marcovisserFurore: I used `SemVersion` now for this functionality.

src/Hl7.Fhir.Core/Serialization/JsonSerializationOptionsExtensions.cs
marcovisserFurore: Moved to the satellites projects as a shim.

src/Hl7.Fhir.Core/Rest/TypedEntryResponseToBundle.cs
marcovisserFurore: This will be an extra task (and another PR). It is on the todo list

src/Hl7.Fhir.Specification/Specification/Source/WebResolver.cs
marcovisserFurore: I've created a `CommonWebSource`.

src/Hl7.Fhir.Core/ElementModel/PocoBuilderExtensions.cs
```diff
@@ -16,14 +16,17 @@ namespace Hl7.Fhir.ElementModel
 {
     public static class PocoBuilderExtensions
     {
-        public static Base ToPoco(this ISourceNode source, Type pocoType = null, PocoBuilderSettings settings = null) =>
-            new PocoBuilder(ModelInspector.ForAssembly(pocoType.Assembly), settings).BuildFrom(source, pocoType);
+        public static Base ToPoco(this ISourceNode source, ModelInspector inspector, Type pocoType = null, PocoBuilderSettings settings = null) =>
```
ewoutkramer, line 19: This I do not understand. Let's discuss tomorrow ;-)

marcovisserFurore: It goes wrong with `bundle.ToTypedElement()` . `Bundle`'s type information can still be found in the big common, but... `Bundle` can also contain other resources, such as `Patient`. At that moment (if we are in the middle of the big common), then type information of `Patient` cannot be retrieved. It's somewhere in the satellites that common doesn't know about. Hence this overload.

ewoutkramer: Well, that's a good one. But then we have not solved it, since this is true for other things than `Bundle`, like `Parameters`, where `Parameters.value[x]` and `Parameters.resource` can contain instances from a different satellite than common. And how would the user know when to call this overload instead of the other?

marcovisserFurore: Solved in PR #2272

ewoutkramer:
Still todo:

- ~~Class SchemaCollection to satellite~~
- SearchParams in all satellites
- Remove enum ResourceType out of big-common
- Move all Open Types to Common
- Organize all of the libraries
- Change Codegen for generating classes

PR #2272, opened 2022-10-20
Update 2 on big common 

marcovisserFurore:
## Description
- `ModelInspector` implements `IModelInfo`. Using this `IModelInfo` in big-common.
- Removed the enum `FhirAllTypes` in big-common

## Still to do:
- Class `SchemaCollection` to satellite
- `SearchParams` in all satellites
- Remove enum `ResourceType` out of big-common
- Move all Open Types to Common
- Organize all of the libraries
- Change Codegen for generating classes


src/Hl7.Fhir.Core.Tests/Introspection/ModelInspectorExtensionsTests.cs
```diff
@@ -0,0 +1,114 @@
...
+            ModelInspectorExtensions.ForInstance(instance).Should().Be(expectedResult);
+        }
+
+        private static IEnumerable<object[]> getTestData()
+        {
+            yield return new object[] { new Patient(), STATELLITE_INSPECTOR };
+            yield return new object[] { new Bundle(), MIDDLE_INSPECTOR };
+            yield return new object[] { new FhirString(), COMMON_INSPECTOR };
+
+            yield return new object[] { new Bundle()  {
```
ewoutkramer, line 29: So cool to see this working!

src/Hl7.Fhir.Core/ElementModel/TypedElementExtensions.cs
```diff
@@ -6,14 +6,17 @@
...
 using Hl7.Fhir.Model;
 
 namespace Hl7.Fhir.ElementModel
 {
     public static class TypedElementExtensions
     {
-        public static ITypedElement ToTypedElement(this Base @base, string rootName = null) =>
-            new PocoElementNode(ModelInspector.ForAssembly(@base.GetType().Assembly), @base, rootName: rootName);
+        public static ITypedElement ToTypedElement(this Base @base, string? rootName = null)
+            => new PocoElementNode(ModelInspectorExtensions.ForInstance(@base), @base, rootName: rootName);
```
ewoutkramer, line 19: Have we carefully tracked where we have been using `ForAssembly()` and it should really be `ForInstance()`?

src/Hl7.Fhir.Core/Introspection/ModelInspectorExtensions.cs
```diff
@@ -0,0 +1,86 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+
+namespace Hl7.Fhir.Introspection
+{
+
+    /// <summary>
+    /// Extensions on ModelInspector. This code should be moved to the ModelInspector class itself when Bundle moved
```
ewoutkramer, line 22: What does this have to do with Bundle?

src/Hl7.Fhir.Core/Introspection/ModelInspectorExtensions.cs
```diff
@@ -0,0 +1,86 @@
...
+                                .Concat(bundle.Entry.Select(e => e.Response?.Outcome).OfType<Base>()),
+                Parameters parameters => parameters.Parameter.Select(p => p.Value),
+
+                _ => Enumerable.Empty<Base>()
+            };
+            resources = resources.Where(t => t is not null);
+
+            return resources.Any() == false
+                ? resources
+                : resources.Concat(resources.SelectMany(r => referredResources(r))).ToList();
```
ewoutkramer, line 80: The recursion.....we need it to detect that there might be resources from different satellites, but on the other hand, we could put our head in the sand and just assume that any resources found at this point in time, will provide enough information on which satellite to use, and we do not need to go deeper.

src/Hl7.Fhir.Core/Introspection/ModelInspectorExtensions.cs
```diff
@@ -0,0 +1,86 @@
...
+            }
+
+            return inspector ?? ModelInspector.ForAssembly(callingAssembly);
+        }
+
+        private static IEnumerable<Base> referredResources(Base @base)
+        {
+            var resources = @base switch
+            {
+                DomainResource domainResource => domainResource.Contained.Select(c => c).OfType<Base>(),
```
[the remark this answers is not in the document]
ewoutkramer: `Select(c=>c)` ??

src/Hl7.Fhir.Core/Introspection/ModelInspectorExtensions.cs
```diff
@@ -0,0 +1,86 @@
...
+                Bundle bundle => bundle.Entry.Select(e => e.Resource).OfType<Base>()
+                                .Concat(bundle.Entry.Select(e => e.Response?.Outcome).OfType<Base>()),
+                Parameters parameters => parameters.Parameter.Select(p => p.Value),
+
+                _ => Enumerable.Empty<Base>()
+            };
+            resources = resources.Where(t => t is not null);
+
+            return resources.Any() == false
```
ewoutkramer, line 78: I think you can turn the `resources` into a list here and turn the `Contact` into an `AddRange()` to avoid running the enumerable several times (Any(), SelectMany(), ToList())

src/Hl7.Fhir.Core/Rest/TransactionBuilder.cs
```diff
@@ -268,7 +269,7 @@ private string paramValueToString(Parameters.ParameterComponent parameter)
                     case CodeableConcept codeableConcept:
                         return codeableConcept.ToToken();
                     default:
-                        if (ModelInfoNEW.IsPrimitive(parameter.Value.GetType()))
+                        if (ModelInfoExtensions.IsPrimitive(parameter.Value.GetType()))
```
ewoutkramer, line 272: I could not find the class file for ModelInfoExtensions, but I guess we need to assume here al FHIR *primitives* are known to common, right?

src/Hl7.Fhir.R5.Core/Serialization/FhirJsonParser.cs
```diff
@@ -58,14 +58,14 @@ public async Tasks.Task<Base> ParseAsync(string json, Type dataType = null)
         /// <inheritdoc cref="ParseAsync(JsonReader, Type)" />
         public Base Parse(JsonReader reader, Type dataType = null)
         {
-            var rootName = dataType != null ? ModelInfoNEW.GetFhirTypeNameForType(dataType) : null;
+            var rootName = dataType != null ? ModelInfo.ModelInspector.GetFhirTypeNameForType(dataType) : null;
```
ewoutkramer, line 61: In the satellites, saying `ModelInfo.ModelInspector.X()` is the same as `ModelInfo.X()`, right?

src/Hl7.Fhir.R5.Core/Specification/Validation/PocoValidationExtensions.cs
```diff
@@ -24,22 +21,23 @@ public static class PocoValidationExtensions
     {
         public static OperationOutcome Validate(this Validator me, Base instance)
         {
-            return me.Validate(instance.ToTypedElement());
+            return me.Validate(instance.ToTypedElement(), ModelInfo.ModelInspector);
```
ewoutkramer, line 24: If we'd have an overload for `Validator.Validate(Base b)` (I am pretty sure we have it), we would not need this overload with the ModelInspector?  Because then the validator can derive the modelinspector from the instance passed in with the new `ModelInspector.ForInstance()`?

ewoutkramer: basically, we would "push this code down to the validator" instead of repeating the `ModelInfo.ModelInspector` here everywhere?

src/Hl7.Fhir.Specification.Tests/Source/ArtifactSummaryTests.cs
```diff
@@ -77,7 +77,7 @@ public void TestValueSetXmlSummary()
             // Common properties
             Assert.IsFalse(summary.IsBundleEntry);
             Assert.AreEqual(ResourceType.ValueSet.GetLiteral(), summary.ResourceTypeName);
-            Assert.IsTrue(summary.ResourceType == ResourceType.ValueSet);
+            Assert.IsTrue(summary.ResourceTypeName == ResourceType.ValueSet.GetLiteral());
```
ewoutkramer, line 80: So, we have broken the public interface on this `summary` thing, right?  It's not used much, so this is ok.

src/Hl7.Fhir.Specification.Tests/Validation/ProfileAssertionTests.cs
```diff
@@ -103,18 +102,18 @@ public async T.Task InitializationAndResolution()
         [Fact]
         public void NormalElement()
         {
-            var assertion = new ProfileAssertion("Patient.name[0]", resolve);
-            assertion.SetDeclaredType(FHIRAllTypes.HumanName.GetLiteral());
+            var assertion = new ProfileAssertion("Patient.name[0]", ModelInfo.ModelInspector, resolve);
+            assertion.SetDeclaredType(ModelInfo.CanonicalUriForFhirCoreType(FHIRAllTypes.HumanName));
```
ewoutkramer, line 106: This is a rather big change (I remember you mentioned it). Did you discover how on !@#$%#$ the original tests used to work with the literal name, and now they need the full canonical?  Is it a change in the interface, or were the old test just bad?

src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefinitionExtensions.cs
```diff
@@ -64,16 +64,15 @@ public static IEnumerable<string> PrimaryTypeProfiles(this ElementDefinition ele
             return elem.Type.FirstOrDefault()?.Profile ?? Enumerable.Empty<string>();
         }
 
-        /// <summary>Returns the type code of the primary element type, or <c>null</c>.</summary>
-        public static FHIRAllTypes? PrimaryTypeCode(this ElementDefinition elem)
+        /// <summary>Returns the typeName of the primary element type, or <c>null</c>.</summary>
+        public static string? GetTypeCode(this ElementDefinition elem)
```
ewoutkramer, line 68: Ah, one more change to the public API. Have you been tracking them on a gist or wiki page?

src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefnMerger.cs
```diff
@@ -172,7 +173,7 @@ void merge(ElementDefinition snap, ElementDefinition diff, bool mergeElementId,
                 snap.Binding = mergeBinding(snap.Binding, diff.Binding);
 
                 // [MV 20220803] Remove Binding when the element has no bindable type
-                if (snap.Binding is not null && !snap.Type.Any(t => ModelInfoNEW.IsBindable(t.Code)))
+                if (snap.Binding is not null && !snap.Type.Any(t => ModelInfoExtensions.IsBindable(t.Code)))
```
ewoutkramer, line 176: Idem, we need to assume all bindeable types are in common. Which is probably true, or at least as true as in the old situation. (I can make up exceptions - but the "old" SDK did not support that either)

src/Hl7.Fhir.Specification/Validation/Validator.cs
```diff
@@ -76,66 +77,72 @@ internal Validator NewInstance()
...
+                modelInspector = this.modelInspector,
 
 #if REUSE_SNAPSHOT_GENERATOR
                 _snapshotGenerator = this._snapshotGenerator
 #endif
             };
         }
 
-        public OperationOutcome Validate(ITypedElement instance)
+        public OperationOutcome Validate(ITypedElement instance, ModelInspector inspector)
```
ewoutkramer, line 88: This is probably our biggest change - and I am not sure we can make it. If you operating in a fully ITypedElement world, you didn't need a ModelInspector until now.  And in fact, for FS and probably Simplifier - this is going to be a problem. They do not have a ModelInspector.  They have a StructureDefinitionSummaryProvider.  I guess we didn't need this before because the validator was not common - so it could use "IsSubtypeFor" etc without a modelinspector.

Now, normally, this would be a show stopper - but we are also going to replace the validator, and make this a shim. I am guessing we don't need the ModelInspector anymore (since our new validator is not model-specific anymore), so this will disappear.  Well, not completely of course, since the *compiler* is still FHIR-version bound.  I think we have some challenges ahead here.

ewoutkramer (changes requested):

src/Hl7.Fhir.Core/Introspection/ModelInspectorExtensions.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+                var assemblies = referredResources(@base)
+                    .Select(r => r.GetType().Assembly)
+                    .Distinct()
+                    .Except(new[] { typeof(IModelInfo).Assembly, callingAssembly }); // remove the common assembly, because that's always loaded
+
+
+                if (assemblies.Count() > 1)
+                {
+                    throw new InvalidOperationException("Found multiple assemblies " +
```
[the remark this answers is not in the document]
marcovisserFurore: I changed it.

src/Hl7.Fhir.Core/Introspection/ModelInspectorExtensions.cs
```diff
@@ -0,0 +1,86 @@
...
+                        $"({string.Join(",", assemblies.Select(a => a.FullName))}) which is not supported.");
+                }
+
+                var assembly = assemblies.SingleOrDefault();
+
+                if (assembly is not null) inspector = ModelInspector.ForAssembly(assembly);
```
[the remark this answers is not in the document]
marcovisserFurore: Yeah you are right. I remember I was struggling with this piece of code. I will change it.

src/Hl7.Fhir.Core/Rest/TransactionBuilder.cs
marcovisserFurore: This class lives in the current Common. And all primitives are in Common.

src/Hl7.Fhir.R5.Core/Serialization/FhirJsonParser.cs
marcovisserFurore: Yes, it is. I changed it.

src/Hl7.Fhir.Core/ElementModel/TypedElementExtensions.cs
marcovisserFurore: We decided to remove `ForInstance`.

src/Hl7.Fhir.Specification.Tests/Source/ArtifactSummaryTests.cs
marcovisserFurore: Yes, this breaks the interface. I will create a list of breaking changes.

src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefinitionExtensions.cs
marcovisserFurore: Yes, a list of breaking changes is necessary.

src/Hl7.Fhir.Specification/Specification/Snapshot/ElementDefnMerger.cs
marcovisserFurore: Indeed all bindeable types are in Common

src/Hl7.Fhir.Specification/Validation/Validator.cs
marcovisserFurore: We are aware of this and we know that this lives in the SDK for a very short time. It means that the new validator should be merged in the SDK for version 5 as well.

src/Hl7.Fhir.Specification.Tests/Validation/ProfileAssertionTests.cs
marcovisserFurore: I removed the following method in `ProfileAssertion`, because we don't have `FHIRAllTypes` anymore.

```c#
 public void SetDeclaredType(FHIRAllTypes declaredType)
{
    SetDeclaredType(ModelInfo.CanonicalUriForFhirCoreType(declaredType));
}
```

So the only public method `SetDeclaredType` left is the only one that takes a string as an argument. That string expects a canonical string. That behavior was also in the current version of the SDK. 

Line 107 I changed in the previous PRs, the original line was `assertion.SetDeclaredType(FHIRAllTypes.HumanName);`

So no big change, only that we removed the public method `public void SetDeclaredType(FHIRAllTypes declaredType)`.

src/Hl7.Fhir.R5.Core/Specification/Validation/PocoValidationExtensions.cs
marcovisserFurore: We decided to remove the `ModelInspector.ForInstance()`.

src/Hl7.Fhir.R5.Core/FhirPath/FhirPathExtensions.cs
```diff
@@ -34,7 +34,7 @@ public static class FhirPathExtensions
             ITypedElement? navResolver(string url)
             {
                 var resource = resolver(url);
-                return resource?.ToTypedElement();
+                return resource?.ToTypedElement(ModelInfo.ModelInspector);
```
ewoutkramer, line 37: Since this is code in a satellite, there is now an overload of `ToTypedElement()` *without* the ModelInspector parameter!

src/Hl7.Fhir.R5.Core/Serialization/PocoSerializationExtensions.cs
```diff
@@ -21,7 +21,7 @@ public static class PocoSerializationExtensions
     {
         /// <inheritdoc cref="ToJsonAsync(Base, FhirJsonSerializationSettings)" />
         public static string ToJson(this Base source, FhirJsonSerializationSettings settings = null) =>
-            source.ToTypedElement().ToJson(settings);
+            source.ToTypedElement(ModelInfo.ModelInspector).ToJson(settings);
```
ewoutkramer, line 24: Idem, we can now use the overload without the ModelInspector param.

src/Hl7.Fhir.R5.Core/Specification/Validation/PocoValidationExtensions.cs
```diff
@@ -21,22 +21,22 @@ public static class PocoValidationExtensions
     {
         public static OperationOutcome Validate(this Validator me, Base instance)
         {
-            return me.Validate(instance.ToTypedElement(), ModelInfo.ModelInspector);
+            return me.Validate(instance.ToTypedElement(ModelInfo.ModelInspector), ModelInfo.ModelInspector);
```
ewoutkramer, line 24: Idem ;-)

src/Hl7.Fhir.Core/Rest/BaseFhirClient.cs
```diff
@@ -969,7 +969,7 @@ private void setResourceBase(Resource resource, string baseUri)
             var request = tx.Entry[0];
             // tx (-> ITyped)? -> entryRequest 
             // entry -> ITyped -> tx
-            var entryRequest = await request.ToEntryRequestAsync(Settings, _fhirVersion).ConfigureAwait(false);
+            var entryRequest = await request.ToEntryRequestAsync(Settings, _inspector, _fhirVersion).ConfigureAwait(false);
```
ewoutkramer, line 972: We have discussed that we want to be able to configure the fhir client with a "pluggable" serializer (so you can use the new ones as well).  I think that once we do that, you will be passing this serializer (or a serializer callback) function down, instead of an inspector.

If this is all internal/private, it doesn't matter for now.

ewoutkramer (approved):

src/Hl7.Fhir.Core/Rest/BaseFhirClient.cs
marcovisserFurore: yeah, this is not the final solution for now.

PR #2285, opened 2022-11-04
Split XSD schema stuff into common and satellites

ewoutkramer:
* Made a satellite version of the SchemaCollection and ValidatorExtensions

TWO BREAKING CHANGES:

* ValidationSettings.XsdSchemaCollection is now of type XmlSchemaSet instead of SchemaCollection.
* SchemaCollection(IArtifactSource) is marked as obsolete, and the constructor will now get the schema from the embedded resource, disregarding the IArtifactSource argument completely.


marcovisserFurore (approved):
Typo, but otherwise approved

PR #2305, opened 2022-11-17
Improved split between ZipSource/CommonZipSource

ewoutkramer:
Turned out CommonZipSource had not enough information to know where the cached (unpacked) zip was locatied.  Also fixed nullability errors.


src/Hl7.Fhir.R5.Core/Specification/Source/FhirPackageSource.cs
```diff
@@ -97,14 +97,14 @@ public IEnumerable<string> ListResourceUris(ResourceType? filter = null)
         }
 
         ///<inheritdoc/>
-        public async Task<IEnumerable<ConceptMap>?> FindConceptMapsAsync(string? sourceUri = null, string? targetUri = null)
+        public async Task<IEnumerable<ConceptMap>> FindConceptMapsAsync(string? sourceUri = null, string? targetUri = null)
```
marcovisserFurore, line 100: I think these changes are breaking changes (return type). So we have to document that.

src/Hl7.Fhir.R5.Core/Specification/Source/FhirPackageSource.cs
```diff
@@ -97,14 +97,14 @@ public IEnumerable<string> ListResourceUris(ResourceType? filter = null)
...
         {
             var resources = await _resolver.FindConceptMaps(sourceUri, targetUri).ConfigureAwait(false);
-            return resources == null ? null : resources.Cast<ConceptMap>();
+            return resources.Cast<ConceptMap>();
         }
 
         ///<inheritdoc/>
-        public IEnumerable<ConceptMap>? FindConceptMaps(string? sourceUri = null, string? targetUri = null)
+        public IEnumerable<ConceptMap> FindConceptMaps(string? sourceUri = null, string? targetUri = null)
```
marcovisserFurore, line 107: I think these changes are breaking changes (return type). So we have to document that.

src/Hl7.Fhir.R5.Core/Specification/Source/ZipSource.cs
```diff
@@ -39,18 +38,33 @@ public static ZipSource CreateValidationSource()
...
         }
 
-        private DirectorySource DirectorySource => FileSource as DirectorySource;
-
         #region IConformanceSource
-        /// <inheritdoc/>
-        public IEnumerable<ConceptMap> FindConceptMaps(string sourceUri = null, string targetUri = null)
-            => DirectorySource.FindConceptMaps(sourceUri, targetUri);
+        /// <summary>Returns a reference to an <see cref="IConformanceSource"/> that exposes the contents of the ZIP archive.</summary>
+        public IConformanceSource Source => new DirectorySourceOnCommon(FileSource);
```
marcovisserFurore, line 43: `this` could be return instead of the private class.

src/Hl7.Fhir.R5.Core/Specification/Source/ZipSource.cs
```diff
@@ -39,18 +38,33 @@ public static ZipSource CreateValidationSource()
...
+            public DirectorySourceOnCommon(CommonDirectorySource source)
+            {
+                Source = source;
+            }
+
+            public CommonDirectorySource Source { get; }
 
-        /// <inheritdoc/>
-        public NamingSystem FindNamingSystem(string uniqueId) => DirectorySource.FindNamingSystem(uniqueId);
+            public CodeSystem? FindCodeSystemByValueSet(string valueSetUri) => Source.FindCodeSystemByValueSet(valueSetUri);
```
marcovisserFurore, line 58: These methods can be pushed to `CommonZipSource` as generics (like ` protected internal T? FindNamingSystem<T>(string uniqueId) where T : Resource`)

marcovisserFurore (changes requested):
See my comments

src/Hl7.Fhir.R5.Core/Specification/Source/FhirPackageSource.cs
ewoutkramer: Done.

src/Hl7.Fhir.R5.Core/Specification/Source/FhirPackageSource.cs
ewoutkramer: Done.

src/Hl7.Fhir.R5.Core/Specification/Source/ZipSource.cs
ewoutkramer: done!

marcovisserFurore (approved):

PR #2307, opened 2022-11-21
FhirClient to common

marcovisserFurore:
Moved the following resources from big common to common:
- `Binary`
- `Bundle`
- `Signature`

Move `FhirClient` from the satellites to common.

Make use of Shared projects

Todo:
- Change codegen to reflect these changes.

See also PR: https://github.com/FirelyTeam/firely-net-common/pull/249


src/Hl7.Fhir.Shared.Core/Model/ModelInfo.cs
```diff
@@ -310,7 +310,7 @@ public static ModelInspector ModelInspector
             typeof(Model.Distance),
             typeof(Model.Dosage),
             typeof(Model.Duration),
-            typeof(Model.Expression),
+          //  typeof(Model.Expression),
```
ewoutkramer, line 313: Why is this removed?

ewoutkramer (changes requested):

src/Hl7.Fhir.Shared.Core/Model/ModelInfo.cs
marcovisserFurore: Oops, well spotted.

marcovisserFurore: Solved: `ModelInfo.OpenTypes` has been moved to the satellites.

ewoutkramer (approved):

PR #2308, opened 2022-11-21
#2019 Set ZipCache Location

### LINKED ISSUE #2019: ZipCacher - FhirArtifactCache location
We use the API to host a FHIR Service using ZipSource to have a source for the CachedResolver for the profile definitions.
ZipSource uses ZipCacher to extract these profile definitions to %TEMP% (via GetTempFilePath) - which is hardcoded.

**problem**
On production environments at customer side occurs a periodically cleanup of the %TEMP% directory which destroys the cache and will cause our service to reject all incomming resources due to validation not succeeded

Overall result: FAILURE (1 errors and 0 warnings) [ERROR] [ERROR] Resolution of profile at 'http://hl7.org/fhir/StructureDefinition/Condition'failed:EinTeildesPfades"C:\Windows\TEMP\FhirArtifactCache-3.5.0-Hl7.Fhir.R4.Specification\tmpEFD4\profiles-resources.xml" konnte nicht gefunden werden. (at Condition).

**solution** 

We would like to be able to specify the location of the artifact cache to set it to another directory, e.g. some folder beneath %programdata%. So a solution could be to have a property within DirectorySourceSettings which is passed to ZipCacher to specify the location of the artifact cache.

ewoutkramer:
Fixes #2019.


src/Hl7.Fhir.Specification/Specification/Source/CommonZipSource.cs
```diff
@@ -34,30 +37,35 @@ public class CommonZipSource : ISummarySource, ICommonConformanceSource, IArtifa
...
+            _mask = settings.Mask;
         }
 
+        /// <inheritdoc cref="CommonZipSource.CommonZipSource(ModelInspector, string, string, DirectorySourceSettings)"/>
+        public CommonZipSource(ModelInspector inspector, string zipPath, string cacheDirectory)
+            : this(inspector, zipPath, cacheDirectory, DirectorySourceSettings.CreateDefault()) { }
+
         /// <summary>Gets the location of the ZIP archive, as specified in the constructor.</summary>
         public string ZipPath { get; }
-        public string TargetDirectory { get; }
```
marcovisserFurore, line 60: This is a breaking change that we have to document.

src/Hl7.Fhir.STU3.Specification.Tests/Source/ZipSourceTests.cs
```diff
@@ -16,7 +17,8 @@ public class ZipSourceTests
         public void ListSummariesIncludingSubdirectories()
         {
             var zipfile = Path.Combine("TestData", "ResourcesInSubfolder.zip");
-            var zip = new ZipSource(zipfile, new DirectorySourceSettings() { IncludeSubDirectories = true });
+            var extractDir = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+            var zip = new ZipSource(zipfile, extractDir, new DirectorySourceSettings() { IncludeSubDirectories = true });
```
marcovisserFurore, line 21: Why should we pass here the extract directory? There is still a constructor of `ZipSource` that generates a default extraction directory. Right?

src/Hl7.Fhir.STU3.Specification.Tests/Source/ZipSourceTests.cs
```diff
@@ -27,7 +29,8 @@ public void ListSummariesIncludingSubdirectories()
         public void ListSummariesExcludingSubdirectories()
         {
             var zipfile = Path.Combine("TestData", "ResourcesInSubfolder.zip");
-            var zip = new ZipSource(zipfile, new DirectorySourceSettings() { IncludeSubDirectories = false });
+            var extractDir = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+            var zip = new ZipSource(zipfile, extractDir, new DirectorySourceSettings() { IncludeSubDirectories = false });
```
marcovisserFurore, line 33: Why should we pass here the extract directory? There is still a constructor of ZipSource that generates a default extraction directory. Right?

src/Hl7.Fhir.Specification.Tests/Source/ZipSourceTests.cs
```diff
@@ -19,7 +20,8 @@ public class ZipSourceTests
         public void ListSummariesIncludingSubdirectories()
         {
             var zipfile = Path.Combine("TestData", "ResourcesInSubfolder.zip");
-            var zip = new ZipSource(zipfile, new DirectorySourceSettings() { IncludeSubDirectories = true });
+            var extractDir = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+            var zip = new ZipSource(zipfile, extractDir, new DirectorySourceSettings() { IncludeSubDirectories = true });
```
marcovisserFurore, line 24: Why should we pass here the extract directory? There is still a constructor of `ZipSource` that generates a default extraction directory. Right?

src/Hl7.Fhir.Specification.Tests/Source/ZipSourceTests.cs
```diff
@@ -30,7 +32,8 @@ public void ListSummariesIncludingSubdirectories()
         public void ListSummariesExcludingSubdirectories()
         {
             var zipfile = Path.Combine("TestData", "ResourcesInSubfolder.zip");
-            var zip = new ZipSource(zipfile, new DirectorySourceSettings() { IncludeSubDirectories = false });
+            var extractDir = Path.Combine(Path.GetTempPath(), Guid.NewGuid().ToString());
+            var zip = new ZipSource(zipfile, extractDir, new DirectorySourceSettings() { IncludeSubDirectories = false });
```
marcovisserFurore, line 36: Why should we pass here the extract directory? There is still a constructor of `ZipSource` that generates a default extraction directory. Right?

marcovisserFurore (changes requested):

src/Hl7.Fhir.Specification/Specification/Source/CommonZipSource.cs
ewoutkramer: Wasn't CommonZipSource new? So there is no current public version of this class?  I'll double check...

ewoutkramer: I double checked, TargetDirectory was introduced by me in a previous PR (for 5.x), so it does not exist on the currently public ZIpSource/CommonZipSource yet, so no breaking change to document.

src/Hl7.Fhir.STU3.Specification.Tests/Source/ZipSourceTests.cs
ewoutkramer: I did half of the unit tests against the constructor with one arg, and the other half against the constructor with two args to be able to test both. But it is a bit explicit (and lazy), I'll create a few more tests.

marcovisserFurore (approved):

PR #2309, opened 2022-11-22
Unify terminology interfaces

ewoutkramer:
(description to follow)


marcovisserFurore (approved):
Approved, but only after common has been updated (see my comment on the PR https://github.com/FirelyTeam/firely-net-common/pull/251)

Issue #2219, opened 2022-09-01
Some Enumerations and properties are not generated properly


marcovisserFurore:
**Describe the bug**
Reported by @Rob5045:

Some properties and enumeration types are not generated properly by the code generator. It concerns the following elements:

#### STU3
Enum Types: (starting with lowercase)
- `Sequence.qualityType`
- `Sequence.repositoryType`

Properties: (ending with underscore)
- `Claim.Use_`
- `TestScript.OperationComponent.ContentType_`

#### R4
Enum Types: (starting with lowercase)
- `messageheader_response_request`
- `MolecularSequence.orientationType`
- `MolecularSequence.qualityType`
- `MolecularSequence.repositoryType`
- `MolecularSequence.sequenceType`
- `MolecularSequence.strandType`
	
Properties: (ending with underscore)
- `EffectEvidenceSynthesis.ResultsByExposureComponent.Exposure_`
- `VerificationResult.Status_`

#### R4B
Enum Types: (starting with lowercase)
- `messageheader_response_request`
- `MolecularSequence.orientationType`
- `MolecularSequence.qualityType`
- `MolecularSequence.repositoryType`
- `MolecularSequence.sequenceType`
- `MolecularSequence.strandType`
- `VerificationResult.status`

Properties: (ending with underscore)
- `Composition.Confidentiality_`
- `EvidenceVariable.CharacteristicCombination_`
- `ResearchElementDefinition.VariableType_`
- `VerificationResult.Status_`

#### R5
Enum Types: (starting with lowercase)
- `messageheader_response_request`
- `MolecularSequence.orientationType`
- `MolecularSequence.qualityType`
- `MolecularSequence.repositoryType`
- `MolecularSequence.sequenceType`
- `MolecularSequence.strandType`
- `VerificationResult.status`

Properties: (ending with underscore)
- `Composition.Confidentiality_`
- `EvidenceVariable.CharacteristicCombination.GroupMeasure_`
- `VerificationResult.Status_`

**Version used:**
 - FHIR Version: STU3, R4, R4B and R5
 - Version: 4.2.1

Changing these elements would introduce breaking changes!

marcovisserFurore:
After refinement 2022-09-08:
- we should change codegen so we generate the correct types and properties
- the existing incorrect properties and enum will be retained and flagged as obsolete (move them to a separate partial class).

ewoutkramer:
[...]

PR #2330, opened 2022-12-12
2219 - Correct names for enums

ewoutkramer:
* Renamed enums for which the enum name clashes with a property name in the same class.
* Made sure all enum names are capitalized.


ewoutkramer:
Fixes #2219

marcovisserFurore (approved):
I checked also the breaking changes in https://github.com/FirelyTeam/firely-net-sdk/wiki/Breaking-changes-in-5.0. ✅

PR #2335, opened 2022-12-14
Remove  CheckMinorVersionCompatibility functionality out of ModelInfo

mmsmits:
Also made SemVersion internal instead of public


marcovisserFurore (approved):
Can you update [this page](https://github.com/FirelyTeam/firely-net-sdk/wiki/Breaking-changes-in-5.0) as well: `SemVersion` is internal now.

PR #2344, opened 2022-12-21
Adde nuspec that can be used to create the Specification metpackage.

ewoutkramer:


marcovisserFurore (changes requested):
Maybe add file in the `.\src` directory.

marcovisserFurore (approved):

PR #2358, opened 2023-01-06
Allign CanonicalUri classes

### LINKED ISSUE #2268: Utility class for processing Canonical Urls (to split version)
When using Canonical URLs there's often a need to be able to quickly split the version from the actual canonical.
Implementation I'm using in my facade project would be good to have added to the core support library (isn't version specific)
https://github.com/brianpos/fhir-net-web-api/blob/feature/r4b/src/Hl7.Fhir.StructuredDataCapture/CanonicalUrl.cs
Example where I've been using it is for references from QuestionnaireResponses to Questionnaires, and Questionnaires to ValueSets.
(This could also be used in the validator where version specific references are detected)

Would also consider adding the processing to be able to detect and strip out a bookmark appended to it also.

Move the class referenced into the Firely SDK.

mmsmits:
fixes #2268


ewoutkramer (changes requested):

ewoutkramer (approved):

brianpos:
@ewoutkramer   Wondering why this isn't in https://github.com/FirelyTeam/firely-net-common/blob/develop/src/Hl7.Fhir.Support.Poco/Model/Canonical.cs where it isn't version specific?

ewoutkramer:
> @ewoutkramer Wondering why this isn't in https://github.com/FirelyTeam/firely-net-common/blob/develop/src/Hl7.Fhir.Support.Poco/Model/Canonical.cs where it isn't version specific?

In SDK 5.0, the common repo has been integrated with the main one. So, this is actually the same Canonical you made your PR on!

PR #2359, opened 2023-01-09
Prepare FhirClient for refactoring

ewoutkramer:
After the inital 5.0 release, we will need to refactor the FhirClient. This PR makes the necessary breaking changes to the public interface to be able to do so.

It also makes all public methods on the FhirClient virtual, so they can be mocked.


ewoutkramer:
These are the breaking changes in this PR:

BaseFhirClient:
	•  internal HttpClientRequester Requester { get; init; } from protected to internal.
	• Protected constructor has been replaced by multiple public constructor.  Class is no longer abstract.

UriParamList.FromUri has been removed - it always returned null
BundleToEntryRequest is now internal
EntryRequest is now internal
EntryRequestHeaders is now internal
EntryResponse is now internal
TypedEntryResponse is now internal
EntryToTypedEntryExtensions is now internal
HttpToEntryExtensions is now internal
TypedEntryResponseToBundle is now internal
WebRequestExtensions has been removed (all methods were already internal)

mmsmits (approved):

PR #2360, opened 2023-01-09
Add CancellationToken support

### LINKED ISSUE #2024: Async operations on FhirClient to take a CancellationToken that is can pass to any web requests it makes
When performing async requests the ability to provide a cancellation token will assist in permitting cancelling the requests - such as those to a web api - when a call is cancelled.
Very useful in terminology validations or resolving resources which may be cancelled too.

Include the CancellationToken to any async FhirClient calls, and use that on any web/directory async calls it makes.

**Additional context**
Server validations, client validations, or any other async general requests performed.
I would like this on the ResourceResolver interfaces too, but can do that seperately.

ewoutkramer:
Added a CancellationToken parameter to all FhirClient operations.

Obsoleted non-async calls.

Fixed #2024.

### DIFF
--- a/src/Hl7.Fhir.Base/Rest/HttpClientRequester.cs
+++ b/src/Hl7.Fhir.Base/Rest/HttpClientRequester.cs
@@ -10,6 +10,7 @@ using Hl7.Fhir.Utility;
 using System;
 using System.Net.Http;
 using System.Net.Http.Headers;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace Hl7.Fhir.Rest
@@ -40,7 +41,7 @@ namespace Hl7.Fhir.Rest
             _disposeHttpClient = false;
         }
 
-        public async Task<EntryResponse> ExecuteAsync(EntryRequest interaction)
+        public async Task<EntryResponse> ExecuteAsync(EntryRequest interaction, CancellationToken ct)
         {
             if (interaction == null) throw Error.ArgumentNull(nameof(interaction));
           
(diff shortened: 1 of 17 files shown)


src/Hl7.Fhir.STU3.Tests/Rest/FhirClientTests.cs
```diff
@@ -1324,7 +1327,7 @@ private static void testHandlingHtmlErrorStatus(BaseFhirClient client)
         {
             try
             {
-                var pat = client.Read<Patient>("Patient/1");
```
brianpos, line 1327: Not a fan of this change, if it's to make a compiler warning go away, add in a Diagnostics.Trace.WriteLine($"{pat.Id}"); or similar.
This is too useful just to put a breakpoint on what's come back from the server while diagnosing things.
Having it thrown away makes it harder to inspect things.

src/Hl7.Fhir.Base/Rest/HttpClientRequester.cs
```diff
@@ -40,7 +41,7 @@ public HttpClientRequester(Uri baseUrl, FhirClientSettings settings, HttpClient
             _disposeHttpClient = false;
         }
 
-        public async Task<EntryResponse> ExecuteAsync(EntryRequest interaction)
+        public async Task<EntryResponse> ExecuteAsync(EntryRequest interaction, CancellationToken ct)
         {
             if (interaction == null) throw Error.ArgumentNull(nameof(interaction));
```
brianpos, line 47: Should we add in something here to not bother calling things if the cancellation token has already fired?
(for larger batches would then prevent going into the subsequent calls once the cancel is done)

brianpos:
This PR will address #2024

src/Hl7.Fhir.STU3.Tests/Rest/FhirClientTests.cs
ewoutkramer: Yes, it was just to clean up the warnings. I'll adapt this change to your suggestion.

src/Hl7.Fhir.Base/Rest/HttpClientRequester.cs
ewoutkramer: I have added token.ThrowIfCancellationRequested().

mmsmits (commented):
Should we have a test to check if the CancellationTokens work?

mmsmits (approved):

PR #2364, opened 2023-01-11
Small corrections

marcovisserFurore:
## Description
- ModelInspector.Common => ModelInspector.Base
- FhirModelAssembly for Base is since STU3
- FhirModelAssembly for Conformance is since R4
- Deserializing and Serializing Integer64 with multiple DeclaredTypes
- Update FluentAssertion dependency
- Changed Copyright year
- Bump BenchmarkDotNet from 0.13.3 to 0.13.4
- Implicit Markdown operators can handle null values

## Corresponding PRs:
- https://github.com/FirelyTeam/fhir-codegen/pull/22

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Base/Serialization/FhirJsonPocoDeserializer.cs
+++ b/src/Hl7.Fhir.Base/Serialization/FhirJsonPocoDeserializer.cs
@@ -678,6 +678,7 @@ namespace Hl7.Fhir.Serialization
                 JsonTokenType.String when requiredType == typeof(long) => readLong(ref reader),
                 //JsonTokenType.String when requiredType.IsEnum => readEnum(ref reader, requiredType),
                 JsonTokenType.String => unexpectedToken(ref reader, reader.GetString(), requiredType.Name, "string"),
+                JsonTokenType.Number when requiredType == typeof(long) => new(null, ERR.LONG_CANNOT_BE_PARSED.With(ref reader, reader.GetRawText(), requiredType.Name)),
                 JsonTokenType.Number => tryGetMatchingNumber(ref reader, requiredType),
                 JsonTokenType.True or JsonTokenType.False when requiredType == typeof(bool) => new(reader.GetBoolean(), null),
                 JsonTokenType.True or JsonTokenType.False => unexpectedToken(ref reader, reader.GetRawText(), requiredType.Name, "boolean"),
@@ -721,7 +722,7 @@ namespace Hl7.Fhir.Serialization
 
                 return long.TryParse(contents, out var parsed) ?
                     new(parsed, null) :
-                    new(contents, ERR.STRING_ISNOTAN_INSTANT.With(ref reader, contents));
+                    new(contents, ERR.NUMBER_CANNOT_BE_PARSED.With(ref reader, contents));
             }
 
             // Validation is now done using POCO validation, so have removed it here.
--- a/src/Hl7.Fhir.Base/Serialization/FhirJsonPocoDeserializer.cs
+++ b/src/Hl7.Fhir.Base/Serialization/FhirJsonPocoDeserializer.cs
@@ -728,9 +728,9 @@ namespace Hl7.Fhir.Serialization
                 return long.TryParse(contents, out var parsed) switch
                 {
                     true when isInteger64() => new(parsed, null),
-                    true => new(parsed, ERR.LONG_CANNOT_BE_PARSED.With(ref reader, contents, typeName())),
-                    false when isInteger64() => new(contents, ERR.NUMBER_CANNOT_BE_PARSED.With(ref reader, contents, nameof(Integer64))),
-                    false => new(contents, ERR.LONG_CANNOT_BE_PARSED.With(ref reader, contents, typeName()))
+                    true => new(parsed, ERR.LONG_INCORRECT_FORMAT.With(ref reader, "Json string", contents, typeName(), "Json number")),
+                    false when isInteger64() => new(contents, ERR.LONG_CANNOT_BE_PARSED.With(ref reader, contents, nameof(Integer64))),
+                    false => new(contents, ERR.LONG_INCORRECT_FORMAT.With(ref reader, "Json string", contents, typeName(), "Json number"))
                 };
 
                 string typeName()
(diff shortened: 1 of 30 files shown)


src/Hl7.Fhir.Base/Serialization/FhirJsonPocoDeserializer.cs
```diff
@@ -675,6 +675,7 @@ FhirJsonPocoDeserializerState state
                                 !Settings.DisableBase64Decoding ? readBase64(ref reader) : new(reader.GetString(), null),
                 JsonTokenType.String when requiredType == typeof(DateTimeOffset) => readDateTimeOffset(ref reader),
                 JsonTokenType.String when requiredType.IsEnum => new(reader.GetString(), null),
+                JsonTokenType.String when requiredType == typeof(long) => readLong(ref reader),
```
ewoutkramer, line 678: When writing the parser I assumed longs are also serialized as json numbers, not strings - is that incorrect ? then we must probably adapt tryGetMatchingNumber - since long handling was done there, and we should trigger an error if the model says something is a long, but the json serialization uses a non-string.

ewoutkramer: You'd say based on http://hl7.org/fhir/json.html#primitive that longs are indeed strings. I wonder if this is intentional. Anyway, you probably found an example that does this (thus triggering this change ;-)).

src/Hl7.Fhir.Base/Serialization/FhirJsonPocoDeserializer.cs
```diff
@@ -713,6 +714,16 @@ FhirJsonPocoDeserializerState state
...
             }
 
+            static (object?, FhirJsonException?) readLong(ref Utf8JsonReader reader)
+            {
+                // convert string in json to a long.
+                var contents = reader.GetString()!;
+
+                return long.TryParse(contents, out var parsed) ?
+                    new(parsed, null) :
+                    new(contents, ERR.STRING_ISNOTAN_INSTANT.With(ref reader, contents));
```
ewoutkramer, line 724: That seems to be a copy/paste error (STRING_ISNOTAN_INSTANT), we're talking about longs here.

src/Hl7.Fhir.Serialization.R4.Tests/RoundtripNewSerializers.cs
```diff
@@ -6,6 +6,8 @@ namespace Hl7.Fhir.Serialization.Tests
     [TestClass]
     public partial class RoundTripNewSerializers
     {
+        private readonly string _attachmentJson = "{\"size\":12}";
```
ewoutkramer, line 9: It seems we now have a different serialization for long values for R4/R4B and R5? Scary.  That's the first time this happens in 10 years, truly scary.

ewoutkramer (changes requested):
I can see https://build.fhir.org/json.html#primitive in R5 is very explicit about long being a string, and the rest is not. Are we sure we are doing the right thing by changing the serialization between versions of FHIR?

ewoutkramer:
"small" corrections 😂😂

src/Hl7.Fhir.Base/Serialization/FhirJsonPocoDeserializer.cs
marcovisserFurore: Changed it to `ERR.NUMBER_CANNOT_BE_PARSED`

src/Hl7.Fhir.Serialization.R4.Tests/RoundtripNewSerializers.cs
ewoutkramer: I mistunderstood this - in R4 the datatype was different, so the serialization changed too.

marcovisserFurore:
> I can see https://build.fhir.org/json.html#primitive in R5 is very explicit about long being a string, and the rest is not. Are we sure we are doing the right thing by changing the serialization between versions of FHIR?

The type `Integer64` has been introduced in R5, but we use it in our Base library for all versions. We moved the datatype `Attachment` to Base as well and the type of the property `size` changed from `UnsignedInt` to `Integer64` from 5.0.0-snapshot3. So it means we have prior to R5 resources that use Attachment with an UnsignedInt and from R5 it is Integer64. The only difference is in serialization (string for integer64, number for unsignedint). 
That's why this PR.

src/Hl7.Fhir.Base/Serialization/FhirJsonPocoDeserializer.cs
marcovisserFurore: See my general comment in this PR.

src/Hl7.Fhir.Serialization.R4.Tests/RoundtripNewSerializers.cs
marcovisserFurore: See my general comment in this PR.

src/Hl7.Fhir.Base/Serialization/FhirJsonPocoDeserializer.cs
```diff
@@ -713,6 +720,26 @@ FhirJsonPocoDeserializerState state
...
+            {
+                // convert string in json to a long.
+                var contents = reader.GetString()!;
+
+                return long.TryParse(contents, out var parsed) switch
+                {
+                    true when isInteger64() => new(parsed, null),
+                    true => new(parsed, ERR.LONG_CANNOT_BE_PARSED.With(ref reader, contents, typeName())),
+                    false when isInteger64() => new(contents, ERR.NUMBER_CANNOT_BE_PARSED.With(ref reader, contents, nameof(Integer64))),
+                    false => new(contents, ERR.LONG_CANNOT_BE_PARSED.With(ref reader, contents, typeName()))
```
ewoutkramer, line 733: Since this is a string, but it cannot be parsed as a number, something is wrong. But the error message says: "Json number '{0}' cannot be parsed as a {1}. Json token should be string." - which is confusing since the token IS a string at this point.

src/Hl7.Fhir.Base/Serialization/FhirJsonPocoDeserializer.cs
```diff
@@ -713,6 +720,26 @@ FhirJsonPocoDeserializerState state
...
 
+            static (object?, FhirJsonException?) readLong(ref Utf8JsonReader reader, Type? fhirType)
+            {
+                // convert string in json to a long.
+                var contents = reader.GetString()!;
+
+                return long.TryParse(contents, out var parsed) switch
+                {
+                    true when isInteger64() => new(parsed, null),
+                    true => new(parsed, ERR.LONG_CANNOT_BE_PARSED.With(ref reader, contents, typeName())),
```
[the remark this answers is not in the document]
marcovisserFurore: I've changed the error message. I even added a new errorcode for that.

src/Hl7.Fhir.Base/Serialization/FhirJsonPocoDeserializer.cs
marcovisserFurore: I've changed the error message. I even added a new errorcode for that.

ewoutkramer (approved):

PR #2371, opened 2023-01-24
Make Markdown primitive

ewoutkramer:
Regenerated the POCO's to make all elements of type Markdown follow the "primitive property" pattern.


marcovisserFurore (approved):
For STU3: 
The type of `ElementDefinition.mapping.comment`, `ElementDefinition.constraint.requirements` and `ElementDefinition.binding.description` is still FhirString (which differs from R4+ (Markdown)  ).
We leave it like this, because it is still correct for STU3.

PR #2374, opened 2023-01-26
2373 Added support elements that have changed from choice to non-choice.

### LINKED ISSUE #2373: Datatype Signature does not work for STU3
The datatype `Signature` has been moved since SDK 5.0 to the Base library. So this datatype is used by all the satellites (STU3 to R5). 
But STU3 version of Signature differs quite significantly from R4 and up. 

So users have problems when they serialize and deserialize this datatype in STU3.

We have to change this Poco, so that is working in all of the FHIR versions.

ewoutkramer:
Fixes #2373.

### DIFF
--- a/src/Hl7.Fhir.Serialization.R4.Tests/RoundtripSignature.cs
+++ b/src/Hl7.Fhir.Serialization.R4.Tests/RoundtripSignature.cs
@@ -31,7 +31,7 @@ namespace Hl7.Fhir.Serialization.Tests
         [TestMethod]
         public void SDSHasCorrectInfo()
         {
-            var cls = new PocoStructureDefinitionSummaryProvider().Provide("http://hl7.org/fhir/StructureDefinition/Signature");
+            var cls = new PocoStructureDefinitionSummaryProvider().Provide(ModelInfo.CanonicalUriForFhirCoreType("Signature"));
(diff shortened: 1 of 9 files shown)


src/Hl7.Fhir.Serialization.R4.Tests/RoundtripSignature.cs
```diff
@@ -0,0 +1,83 @@
...
+            prop.Choice.Should().Be(Introspection.ChoiceType.None);
+            prop.ImplementingType.Should().Be(typeof(DataType));
+            prop.PropertyTypeMapping.Name.Should().Be("Reference");
+            prop.FhirType.Should().BeEquivalentTo(new[] { typeof(ResourceReference) });
+        }
+
+        [TestMethod]
+        public void SDSHasCorrectInfo()
+        {
+            var cls = new PocoStructureDefinitionSummaryProvider().Provide("http://hl7.org/fhir/StructureDefinition/Signature");
```
marcovisserFurore, line 34:
```suggestion
            var cls = new PocoStructureDefinitionSummaryProvider().Provide(ModelInfo.CanonicalUriForFhirCoreType("Signature"));
```

marcovisserFurore (approved):
Minor thing

PR #2377, opened 2023-01-27
Signature is now generated by codegen

marcovisserFurore:
## Description
The datatype `Signature` is now generated by codegen.  Changes:
- `Signature.targetFormat` Added Element since R4
- `Signature.sigFormat` Added Element since R4
- `Signature.data` Renamed from blob to data since R4
- `Signature.contentType` deleted since R4
- `Signature.who` can be a choicetype in STU3 and a reference in R4 and up
- `Signature.onBehalfOf` can be a choicetype in STU3 and a reference in R4 and up

See also PR on codegen: https://github.com/FirelyTeam/fhir-codegen/pull/26

### DIFF
--- a/src/Hl7.Fhir.Base/Model/Generated/Signature.cs
+++ b/src/Hl7.Fhir.Base/Model/Generated/Signature.cs
@@ -103,14 +103,14 @@ namespace Hl7.Fhir.Model
     /// <summary>
     /// Who signed. Note: Since R5 the cardinality is expanded to 0..1 (previous it was 1..1)
     /// </summary>
-    [FhirElement("who", InSummary = true, Order = 50, Choice = ChoiceType.DatatypeChoice)]
-    [FhirElement("who", InSummary = true, Order = 50, Since = FhirRelease.R4)]
-    [CLSCompliant(false)]
-    [References("Practitioner", "RelatedPerson", "Patient", "Device", "Organization")]
-    [References("Practitioner", "PractitionerRole", "RelatedPerson", "Patient", "Device", "Organization",Since=FhirRelease.R4)]
+    [FhirElement("who", InSummary=true, Order=50, Choice = ChoiceType.DatatypeChoice)]
+    [FhirElement("who", InSummary=true, Order=50, Since=FhirRelease.R4)]
+    [DeclaredType(Type = typeof(ResourceReference), Since = FhirRelease.R4)]
     [AllowedTypes(typeof(Hl7.Fhir.Model.FhirUri), typeof(Hl7.Fhir.Model.ResourceReference))]
-    [DeclaredType(Type = typeof(ResourceReference), Since=FhirRelease.R4)]
-    [DataMember]      
+    [CLSCompliant(false)]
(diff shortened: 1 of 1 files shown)


src/Hl7.Fhir.Base/Model/Generated/Signature.cs
```diff
@@ -103,14 +103,14 @@ public Hl7.Fhir.Model.Instant WhenElement
     /// <summary>
     /// Who signed. Note: Since R5 the cardinality is expanded to 0..1 (previous it was 1..1)
     /// </summary>
-    [FhirElement("who", InSummary = true, Order = 50, Choice = ChoiceType.DatatypeChoice)]
-    [FhirElement("who", InSummary = true, Order = 50, Since = FhirRelease.R4)]
-    [CLSCompliant(false)]
-    [References("Practitioner", "RelatedPerson", "Patient", "Device", "Organization")]
-    [References("Practitioner", "PractitionerRole", "RelatedPerson", "Patient", "Device", "Organization",Since=FhirRelease.R4)]
+    [FhirElement("who", InSummary=true, Order=50, Choice = ChoiceType.DatatypeChoice)]
```
ewoutkramer, line 106: Should the doccomment mention the change of choice -> fixed type?

src/Hl7.Fhir.Base/Model/Generated/Signature.cs
```diff
@@ -231,6 +263,38 @@ public byte[] Data
       }
     }
 
+    /// <summary>
+    /// The actual signature content (XML DigSig. JWS, picture, etc.). Note: Element is deprecated since R4, do not use with R4 and newer releases.
+    /// </summary>
+    [FhirElement("blob", Order=90)]
```
ewoutkramer, line 269: Actually, blob was renamed to data -> much like in Binary. Shouldn't `data` say that it can be used since R4?

ewoutkramer: Is it an idea to mention the word "rename" here? If so, we should probably do that for Binary too.

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/Model/Generated/Signature.cs
marcovisserFurore: For Binary I changed it now like this:
```
   /// The actual content. Note: Element is replaced by 'Binary.data' since R4. Do not use this element 'content' with R4 and newer releases.
```

src/Hl7.Fhir.Base/Model/Generated/Signature.cs
marcovisserFurore: I changed it now like this:
```
   /// <summary>
    /// Who signed.
    /// Note 1: Since R4 the type of this element should be a fixed type (ResourceReference). For backwards compatibility it remains of type DataType.
    /// Note 2: Since R5 the cardinality is expanded to 0..1 (previous it was 1..1).
    /// </summary>
 ````

ewoutkramer (approved):

PR #2398, opened 2023-03-02
Improved performance of MatchesPrefix

### LINKED ISSUE #2393: Performance of public method StringExtensions.MatchesPrefix could be improved
From the Simplifier team we have heard that the function `StringExtensions.MatchesPrefix` could be improved in performance when we use ReadOnlySpan<char> instead of string. It makes the code sometimes 10x faster. The function is used in the `Children` method of `ITypedElement` implementations.

The proposed new implementation of `MatchesPrefix` would be (thanks @mharthoorn):
```c#
public static bool FastMatchesPrefix(this ReadOnlySpan<char> text, ReadOnlySpan<char> prefix)
    {
        bool wildcard = prefix[^1] == '*';
        if (wildcard) prefix = prefix[..^1];

        bool found = text.StartsWith(prefix);
        if (!found) return false;

        bool full = text.Length == prefix.Length;

        return (full || wildcard);
    }
```

mmsmits:
fixes #2393


src/Hl7.Fhir.Base/Utility/StringExtensions.cs
```diff
@@ -139,14 +138,30 @@ internal static string[] SplitNotEscaped(this string value, char separator)
         /// <param name="text"></param>
         /// <param name="prefix"></param>
         /// <returns></returns>
-        public static bool MatchesPrefix(this string text, string prefix)
+        public static bool MatchesPrefix(this ReadOnlySpan<char> text, ReadOnlySpan<char> prefix)
         {
-            var prefixMatch = prefix?.EndsWith("*") ?? false;
+            bool wildcard = prefix.EndsWith("*".AsSpan()); //used for value[x] elements
```
ewoutkramer, line 143: Since we are optimizing anyway, shall we not make a 

`static readonly Span Asterix = "*".AsSpan()`

That you can use in the `EndsWith` and the `TrimEnd`, so we can re-use that single span instead of creating two of them in every call?

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/Utility/StringExtensions.cs
```diff
@@ -139,14 +138,31 @@ internal static string[] SplitNotEscaped(this string value, char separator)
         /// <param name="text"></param>
         /// <param name="prefix"></param>
         /// <returns></returns>
-        public static bool MatchesPrefix(this string text, string prefix)
+        public static bool MatchesPrefix(this ReadOnlySpan<char> text, ReadOnlySpan<char> prefix)
         {
-            var prefixMatch = prefix?.EndsWith("*") ?? false;
+            ReadOnlySpan<char> asterix = "*".AsSpan();
```
ewoutkramer, line 143:
```suggestion
            static ReadOnlySpan<char> asterix = "*".AsSpan();
```

ewoutkramer (changes requested):
Almost, you forgot the `static` part.....this way you only make the `asterix` ONCE for every subsequent call to the method.

ewoutkramer (approved):

PR #2430, opened 2023-04-11
2410 - Make serialization pluggable - receiving side of FhirClient

ewoutkramer:
* Small fixes (nullability) to the sending side and improvement to setting headers for binary transmissions.
* Removed multiple layers of DTOs (EntryResponse, TypedEntryResponse, etc). All internal classes.
* Simplified parsing code (by removing these layers) and moving them out as extension methods (in HttpContentParsers). These parsing methods are now separately unit-testable.
* Added unit-tests for all phases of parsing and processing.
* Added mocks for parts that could not be unit-tested, plus streamlined the existing mocks.
* Updated to C# 11 to make json and xml literals in tests more readable.
* Added nullability checks to most classes in Rest namespace.

TODO (in subsequent PRs)
* Add support for handling reception of binary data.


ewoutkramer:
I have used C#11 "raw strings", but that meant changing the langVersion, which meant changing the SDK for the build, which means this branch isn't currently building ;-)

mmsmits (approved):

PR #2434, opened 2023-04-17
XML/Json parsing enhancements

### LINKED ISSUE #2414: Update ExceptionNotification to support location as a complete property
When processing exceptions from serialization would like to have the location information seperated from the message to ease the population of the OperationOutcome issues with that information.
Currently we need to parse out of the message to see if it has had the location data appended, then strip it from the message.
https://github.com/FirelyTeam/firely-net-sdk/blob/7d63b1e1b2eafa9ff1d53c98f87cac409bd1f5e6/src/Hl7.Fhir.Base/ElementModel/TypedElementOnSourceNode.cs#L85
https://github.com/FirelyTeam/firely-net-sdk/blob/7d63b1e1b2eafa9ff1d53c98f87cac409bd1f5e6/src/Hl7.Fhir.Base/Utility/ExceptionSourcing.cs#L36

Don't append the location data in the message, add it as a seperate property.

Another alternative would be to have 2 other properties, rawMessage/location and message to be more backward compatible.

I've implemented a workaround in my facade project to handle this, want to have a less delicate approach
https://github.com/brianpos/fhir-net-web-api/blob/102835a4f6ac34e64687599c583b49ff14c989e3/src/Hl7.Fhir.WebApi.AspNetCore/FhirMediaTypeFormatter.cs#L76

brianpos:
## Description
Update the new XML/Json parsers to better handle/report errors and contintue processing when it can.
Specifically:
* Accurately reporting the location of the issue (fhirpath)
  * Correct when resource is contained in other resources (or bundles)
* Reporting the severity and type of the parsing issue (for use in the OperationOutcome)
* Additional parser setting to continue to validation if recoverable parse issues were identified

## Related issues
Resolves #2414, #2435

## Testing
Additional unit tests were added to verify the parser behaviour across the specific edge cases that it covers.
`JsonInvalid*` and `XmlInvalid*`

## FirelyTeam Checklist


src/Hl7.Fhir.Base/Model/CodeOfT.cs
```diff
@@ -104,7 +104,7 @@ public override IEnumerable<ValidationResult> Validate(ValidationContext validat
                 return baseResults;
             else
             {
-                var result = COVE.INVALID_CODED_VALUE.AsResult(validationContext, ObjectValue, EnumUtility.GetName<T>());
+                var result = COVE.INVALID_CODED_VALUE(validationContext, ObjectValue, EnumUtility.GetName<T>()).AsResult();
```
ewoutkramer, line 107: This would have to become  

`INVALID_CODED_VALUE(ObjectValue,EnumUtility.GetName<T>).AsResult(validationContext)`, as it is important for me to keep the domain of the CodedExceptions separate from the ValidationResult class.  In other words: I want to be able to work with CodedValidationExceptions without needing to know about a `ValidationContext`.  E.g. if we ever move away from the .NET attribute validation, I'd still like to be able to use the same CodedValidationException class. We cannot do that if the constructor is using it.

ewoutkramer: In fact, it would make sense to move the AsResult() totally away from the CodedValidationException by turning this into an extension method on CodedValidationException.

src/Hl7.Fhir.Base/Serialization/PathStack.cs
```diff
@@ -43,33 +51,68 @@ public void ExitResource()
...
-        public void EnterElement(string name)
+        public void EnterElement(string name, int? index, bool isPrimitive)
         {
             bool empty = !_tops.Any();
             _tops.Push(_path.Length);
+            _indexer.Push(index ?? 0);
             _path += empty ? name : $".{name}";
+            if (!isPrimitive)
+            {
+                if (index.HasValue)
```
ewoutkramer, line 70: The original stack was pretty conservative in memory usage since we had 1 string + a stack of integer indices. The new stack now has those PLUS a stack of strings. That's a clear step backwards.  Couldn't we still build a single string (with the index markers) and have, say, two stacks (or a stack with valuetuples) with indices so we can construct both the definition and the instance path from the same string?

src/Hl7.Fhir.Base/Serialization/PathStack.cs
```diff
@@ -43,33 +51,68 @@ public void ExitResource()
...
             else
+            {
+                _indexer.Pop();
                 _path = _path.Substring(0, top);
+            }
+
+            _paths.Pop();
+        }
+
+        public void IncrementIndex()
```
ewoutkramer, line 103: It seems to me we now have the knowledge of indexed property names divided over two classes: the parser needs to pass the initial index + call `IncrementIndex`, and the PathStack needs to handle the effects.  Can't we design this so we have the parser call EnterElement, etc. without "knowing" anything about indexed paths? I don't like to introduce this "knowledge" into the parser, just to keep of the instance path.  Maybe the parser needs to pass a little bit more metadata into the calls, but so be it!

src/Hl7.Fhir.Base/Utility/CodedWithLocationException.cs
```diff
@@ -0,0 +1,97 @@
...
+
+namespace Hl7.Fhir.Utility
+{
+    /// <summary>
+    /// This is a class of Exceptions that is raised by the SDK and is coded with a unique code
+    /// to enable callers to identify this exception and react appropriately on the code.
+    /// </summary>
+    /// <remarks>Most modules of the SDK using this Exception will create specific subclasses
+    /// for this subclass, providing a list of codes used by that module.</remarks>
+    public class CodedWithLocationException : CodedException
```
ewoutkramer, line 22: Should we call this: `OperationOutcomeCodedException`, because that's basically what it is: a CodedException with enough information (not just the location!!) to produce an OperationOutcome from?

ewoutkramer: Then the ToIssue member makes sense, since that presents one of the principal purposes of the OperationOutcomeCodedException.

src/Hl7.Fhir.Base/Validation/ValidationContextExtensions.cs
```diff
@@ -85,12 +85,27 @@ public static ValidationContext SetLocation(this ValidationContext ctx, string l
...
+        }
 
+        /// <summary>
+        /// Gets the human-readable location for the validation errors from the ValidationContext.
+        /// </summary>
+        public static string? GetLocation(this ValidationContext ctx)
+        {
+            if (!ctx.Items.TryGetValue(LOCATION_ITEM_KEY, out var result))
+                return null;
+            if (result is Hl7.Fhir.Serialization.PathStack path)
```
ewoutkramer, line 103: Should we not have different keys for different things?  Does it make sense to have both the string and the stack?

src/Hl7.Fhir.Conformance/Model/Generated/CapabilityStatement.cs
```diff
@@ -42,4651 +42,4651 @@
 
 namespace Hl7.Fhir.Model
```
ewoutkramer, line 43: This must be a mistake? Why is CS being regenerated here?

src/Hl7.Fhir.Serialization.R4B.Tests/Hl7.Fhir.Serialization.R4B.Tests.csproj
```diff
@@ -4,7 +4,12 @@
 
 	<Import Project="..\firely-net-sdk-tests.props" />
 
-	<PropertyGroup>
+  <!-- Compiler settings -->
+  <PropertyGroup>
+    <LangVersion>11</LangVersion>
```
ewoutkramer, line 9: You cannot do this without also updating the build scripts. But I also want to use C# 11, so we should.

src/Hl7.Fhir.Conformance/Model/Generated/Template-Bindings.cs
ewoutkramer, line 1: I hope this is a mistake?

src/Hl7.Fhir.Serialization.R4B.Tests/SerializationExcexptionHandlersJsonPoco.cs
```diff
@@ -0,0 +1,875 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text.Json;
+
+namespace Hl7.Fhir.Serialization.Tests
+{
+    [TestClass]
+    public class SerializationExceptionHandlersJsonPoco
```
ewoutkramer, line 15: This is almost complete duplication of the tests in FhirJsonDeserializationTests.cs - you are just testing whether the operation outcomes are constructed correctly, so these tests are now testing both the parsing (which is already tested) + the mapping to operationoutcome (which was not tested). Feels a bit like a waste of your time.

src/Hl7.Fhir.Shared.Tests/TrainingTests/Z-TrainingUnitTests.cs
```diff
@@ -135,7 +135,7 @@ public void ParseElementDefinitionConstraint()
                 {
                     Expression = expression,
                     Key = key,
-                    Severity = severity == "Error" ? ConstraintSeverity.Error : ConstraintSeverity.Warning,
+                    Severity = severity == "Error" ? Hl7.Fhir.Model.ConstraintSeverity.Error : Hl7.Fhir.Model.ConstraintSeverity.Warning,
```
ewoutkramer, line 138: ConstraintSeverity is/was/is/was part of ElementDefinition, and we cannot introduce this breaking change. That was a mistake in the pre-publication version of the generator which has since been fixed by Marco. We need to revert this change.

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonPocoDeserializer.cs
```diff
@@ -766,14 +780,14 @@ bool isInteger64()
             //}
         }
 
-        private static (object?, FhirJsonException) unexpectedToken(ref Utf8JsonReader reader, string? value, string expected, string actual) =>
-            new(value, ERR.UNEXPECTED_JSON_TOKEN.With(ref reader, expected, actual, value));
+        private static (object?, FhirJsonException) unexpectedToken(ref Utf8JsonReader reader, string locationPath, string? value, string expected, string actual) =>
```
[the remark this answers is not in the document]
brianpos: Yup, I'll rename all those now that I know the standard term for it.

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonPocoDeserializer.cs
```diff
@@ -290,7 +299,7 @@ private enum DeserializedObjectKind
 
                     try
                     {
-                        state.Path.EnterElement(propMapping!.Name);
+                        state.Path.EnterElement(propMapping!.Name, !propMapping.IsCollection ? null : 0, propMapping.IsPrimitive);
```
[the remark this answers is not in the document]
brianpos: The impact of passing in null is that it won't do the index tracking/display at all in the instancePath.
I'm ok either way, but seems clean enough to not bring it in.

src/Hl7.Fhir.Base/Serialization/DeserializationContext.cs
```diff
@@ -20,23 +20,27 @@ namespace Hl7.Fhir.Serialization
...
 
+        private readonly PathStack _path;
+
+        internal PathStack PathStack { get { return _path; } }
+
         /// <summary>
         /// The dotted FhirPath path leading to this element from the root.
         /// </summary>
-        public string Path { get; }
+        public string Path { get { return _path.GetInstancePath(); } }
```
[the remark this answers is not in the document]
brianpos: Interestingly, I didn't encounter any use of the definition style path in my investigations.

src/Hl7.Fhir.Base/Serialization/FhirXmlException.cs
```diff
@@ -33,28 +35,62 @@ public class FhirXmlException : CodedException
...
-        internal static readonly FhirXmlException SCHEMALOCATION_DISALLOWED = new(SCHEMALOCATION_DISALLOWED_CODE, "The 'schemaLocation' attribute is disallowed.");
-        internal static readonly FhirXmlException EXPECTED_OPENING_ELEMENT = new(EXPECTED_OPENING_ELEMENT_CODE, "Expected opening element, but found {0}.");
-        internal static readonly FhirXmlException ENCOUNTERED_DTD_REFERENCES = new(ENCOUNTERED_DTP_REFERENCES_CODE, "There SHALL be no DTD references in FHIR resources (because of the XXE security exploit)");
-        internal static readonly FhirXmlException ELEMENT_HAS_NO_VALUE_OR_CHILDREN = new(ELEMENT_HAS_NO_VALUE_OR_CHILDREN_CODE, "Element '{0}' must have child elements and / or a value attribute");
+        internal static readonly FhirXmlException EMPTY_ELEMENT_NAMESPACE = new(EMPTY_ELEMENT_NAMESPACE_CODE, $"The element '{0}' has no namespace, expected the HL7 FHIR namespace ({XmlNs.FHIR})") { IssueSeverity = Model.OperationOutcome.IssueSeverity.Error, IssueType = Model.OperationOutcome.IssueType.Structure };
```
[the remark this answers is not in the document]
brianpos: This was updated in the commit after you've reviewed this...

src/Hl7.Fhir.Conformance/Model/Generated/CapabilityStatement.cs
brianpos: This was whitespace as a bi-product of merging the develop into the branch (I started from an interim branch that Marco was working on R5 so I could cross test things at the time)

src/Hl7.Fhir.Conformance/Model/Generated/Template-Bindings.cs
brianpos: Yup, merge bi-product (if you select to not show whitespace changes, all goes away)

src/Hl7.Fhir.Serialization.R4B.Tests/SerializationExcexptionHandlersJsonPoco.cs
brianpos: You're basically right, I wanted to recheck all the cases in the old/new parsers to verify differences.
(and still deciding if I want to go back and make the old ones handle all these issues or not - and these would be the tests for them - I did start on the old parsers till I found your new one, and didn't throw the work away)

src/Hl7.Fhir.Shared.Tests/TrainingTests/Z-TrainingUnitTests.cs
brianpos: Then this issue is still current in the build - I get compile errors without it.

src/Hl7.Fhir.Base/Serialization/DeserializationContext.cs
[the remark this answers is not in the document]
brianpos: Also, the comment on the property should then be updated as it's not the fhirpath based path...

src/Hl7.Fhir.Base/Model/CodeOfT.cs
brianpos: This was resolved putting the validation context in both functions and not storing it between calls.

src/Hl7.Fhir.Base/Utility/CodedWithLocationException.cs
brianpos: How about a rename to `ExtendedCodedException`?

src/Hl7.Fhir.Base/Serialization/FhirXmlException.cs
```diff
@@ -29,58 +34,79 @@ public class FhirXmlException : CodedException
...
-        internal static readonly FhirXmlException ENCOUNTERED_DTD_REFERENCES = new(ENCOUNTERED_DTP_REFERENCES_CODE, "There SHALL be no DTD references in FHIR resources (because of the XXE security exploit)");
-        internal static readonly FhirXmlException ELEMENT_HAS_NO_VALUE_OR_CHILDREN = new(ELEMENT_HAS_NO_VALUE_OR_CHILDREN_CODE, "Element '{0}' must have child elements and / or a value attribute");
+        // ==========================================
+        // Unrecoverable Errors (or data loss)
+        // ==========================================
+        internal static FhirXmlException UNKNOWN_RESOURCE_TYPE(XmlReader reader, string instancePath, string typeName) => Initialize(reader, instancePath, UNKNOWN_RESOURCE_TYPE_CODE, $"Unknown type '{typeName}' found in root property.", OO_Sev.Fatal, OO_Typ.Structure);
+        internal static FhirXmlException RESOURCE_TYPE_NOT_A_RESOURCE(XmlReader reader, string instancePath, string resourceType) => Initialize(reader, instancePath, RESOURCE_TYPE_NOT_A_RESOURCE_CODE, $"Data type '{resourceType}' in property 'resourceType' is not a type of resource.", OO_Sev.Fatal, OO_Typ.Structure);
```
ewoutkramer, line 46: This list is longer, see https://github.com/FirelyTeam/firely-net-sdk/blob/ae47a158854646bb210b0f135ddbce5d639d5e2d/src/Hl7.Fhir.Base/Serialization/FhirXmlException.cs

src/Hl7.Fhir.Base/Serialization/FhirXmlException.cs
```diff
@@ -29,58 +34,79 @@ public class FhirXmlException : CodedException
...
+            {
+                BaseErrorMessage = message,
+                LineNumber = lineNumber,
+                Position = position,
+                Location = instancePath,
+            };
         }
 
-        public FhirXmlException(string code, string message, Exception? innerException) : base(code, message, innerException)
+        public FhirXmlException(string errorCode, string message, OperationOutcome.IssueSeverity issueSeverity, OperationOutcome.IssueType issueType) : base(errorCode, message, issueSeverity, issueType)
```
ewoutkramer, line 87: This is a breaking change to the constructor signature - we have to find defaults for the severity/issuetype.

src/Hl7.Fhir.Base/Serialization/FhirXmlPocoDeserializerSettings.cs
```diff
@@ -19,6 +19,26 @@ public class FhirXmlPocoDeserializerSettings
         /// and possibly alter the value. Setting this property to <c>null</c> will disable validation completely.
         /// </summary>
         public IDeserializationValidator? Validator { get; init; } = DataAnnotationDeserialzationValidator.Default;
+
+        /// <summary>
+        /// Perform the parse time validation on the deserialized object even if parsing issues occurred
```
ewoutkramer, line 24:
```suggestion
        /// Perform the parse time validation on the deserialized object even if parsing issues occurred.
```

src/Hl7.Fhir.Base/Serialization/FhirXmlPocoDeserializerSettings.cs
```diff
@@ -19,6 +19,26 @@ public class FhirXmlPocoDeserializerSettings
...
+        /// </summary>
+        /// <remarks>
+        /// This is useful for "strict mode" once pass validators and may result in spurious error messages
+        /// from validating incomplete content
```
ewoutkramer, line 28:
```suggestion
        /// from validating incomplete content.
```

src/Hl7.Fhir.Base/Serialization/FhirXmlPocoDeserializerSettings.cs
```diff
@@ -19,6 +19,26 @@ public class FhirXmlPocoDeserializerSettings
...
+        /// </remarks>
+        public bool ValidateOnFailedParse { get; init; } = false;
+
+        /// <summary>
+        /// During parsing any contained resources (such as those in a bundle) that encounter some form of parse/validation exceotion
```
ewoutkramer, line 33:
```suggestion
        /// During parsing any contained resources (such as those in a bundle) that encounter some form of parse/validation exception
```

src/Hl7.Fhir.Base/Serialization/FhirXmlPocoDeserializerSettings.cs
```diff
@@ -19,6 +19,26 @@ public class FhirXmlPocoDeserializerSettings
...
+        /// will have a List&lt;CodedException&gt; of these exceptions added as an annotion to the child resource.
```
ewoutkramer, line 34:
```suggestion
        /// will have a <c>List&lt;CodedException&gt;</c> of these exceptions added as an annotation to the child resource.
```

src/Hl7.Fhir.Base/Serialization/FhirXmlPocoDeserializerSettings.cs
```diff
@@ -19,6 +19,26 @@ public class FhirXmlPocoDeserializerSettings
...
+        /// </summary>
+        /// <remarks>
+        /// This is primarily added to ease the processing of bundles during a batch submission
+        /// (without requiring processing fhirpath expressions in the issues in the parsing operation outcome to determine if a 
+        /// resource was clean and possibly ok to process)
```
ewoutkramer, line 39:
```suggestion
        /// resource was clean and possibly ok to process).
```

src/Hl7.Fhir.Base/Serialization/OperationOutcomeExtensions.cs
```diff
@@ -0,0 +1,140 @@
...
+
+                // case FhirJsonException.PRIMITIVE_ARRAYS_ONLY_NULL_CODE: shortDisplay = "JSON125"; break;
+                // case FhirJsonException.INCOMPATIBLE_SIMPLE_VALUE_CODE: shortDisplay = "JSON126"; break;
+
+
+                case FhirXmlException.EMPTY_ELEMENT_NAMESPACE_CODE: shortDisplay = "Empty Element Namespace"; break;
+                case FhirXmlException.UNKNOWN_RESOURCE_TYPE_CODE: shortDisplay = "Invalid ResourceType"; break;
+                // case FhirXmlException.RESOURCE_TYPE_NOT_A_RESOURCE_CODE: shortDisplay = "XML103"; break;
+                case FhirXmlException.UNKNOWN_ELEMENT_CODE: shortDisplay = "Unknown element"; break;
+                // case FhirXmlException.CHOICE_ELEMENT_HAS_NO_TYPE_CODE: shortDisplay = "XML105"; break;
```
ewoutkramer, line 104: Are these still todo?

src/Hl7.Fhir.Base/Utility/CodedWithLocationException.cs
```diff
@@ -0,0 +1,72 @@
...
+            IssueType = issueType;
+        }
+
+        public CodedWithLocationException(string errorCode, string message, OperationOutcome.IssueSeverity issueSeverity, OperationOutcome.IssueType issueType, Exception? innerException) : base(errorCode, message, innerException)
+        {
+            IssueSeverity = issueSeverity;
+            IssueType = issueType;
+        }
+
+        /// <summary>
```
ewoutkramer, line 36: I know it's a small thing, but can you put a dot (`.`) at the end of the sentences in the doccomments?

src/Hl7.Fhir.Serialization.R4B.Tests/Hl7.Fhir.Serialization.R4B.Tests.csproj
ewoutkramer: I have enabled C# 11 in the mainline develop, so this should no longer be necessary.

src/Hl7.Fhir.Serialization.R4B.Tests/SerializationExcexptionHandlersJson.cs
```diff
@@ -0,0 +1,312 @@
+﻿using Hl7.Fhir.Model;
```
ewoutkramer, line 1: I think these test classes can go into the Shared test project? Why would we only want them from R4B?

src/Hl7.Fhir.Serialization.R4B.Tests/SerializationExcexptionHandlersJsonPoco.cs
```diff
@@ -0,0 +1,847 @@
+﻿using Hl7.Fhir.Model;
```
ewoutkramer, line 1: When we have pulled this PR: https://github.com/FirelyTeam/firely-net-sdk/pull/2437, we can parameterize the tests with an `IFhirSerializationEngine`, which will halve the number of tests (since you can just pass in the engine, instead of having to create a POCO json + "old" json parser.

src/Hl7.Fhir.Base/Utility/CodedWithLocationException.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// Severity of this specific issue
+        /// </summary>
+        /// <remarks>
+        /// Setter is public to permit others to upgrade/downgrade specific issues
+        /// as needed
+        /// </remarks>
+        public OperationOutcome.IssueSeverity IssueSeverity { get; protected init; } = OperationOutcome.IssueSeverity.Error;
```
ewoutkramer, line 43: For all these, `protected init` can become `init` now.

ewoutkramer (changes requested):
Almost there!  We've only the design issue about separating severity/type from the CodedException left to decide on.

src/Hl7.Fhir.Shared.Tests/TrainingTests/Z-TrainingUnitTests.cs
ewoutkramer: By pulling `develop` after the 5.1 release, this should be fixed.

src/Hl7.Fhir.Base/Serialization/PathStack.cs
ewoutkramer: We discussed this during the call - there are no new strings allocated, so it should be fine.

src/Hl7.Fhir.Serialization.R4B.Tests/SerializationExcexptionHandlersJson.cs
brianpos: Was only there as that's the version I was working in (and didn't have to stress over properties across versions giving different results)

src/Hl7.Fhir.Base/Utility/CodedWithLocationException.cs
brianpos: Added in this file (will keep an eye on that from now)

src/Hl7.Fhir.Base/Serialization/FhirJsonPocoDeserializerSettings.cs
```diff
@@ -42,6 +42,26 @@ public record FhirJsonPocoDeserializerSettings
         /// and possibly alter the value. Setting this property to <c>null</c> will disable validation completely.
         /// </summary>
         public IDeserializationValidator? Validator { get; init; } = DataAnnotationDeserialzationValidator.Default;
+
+        /// <summary>
+        /// Perform the parse time validation on the deserialized object even if parsing issues occurred
```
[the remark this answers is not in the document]
brianpos: updated

src/Hl7.Fhir.Base/Serialization/FhirJsonException.cs
```diff
@@ -57,78 +59,95 @@ public class FhirJsonException : CodedException
...
+        //internal static FhirJsonException PRIMITIVE_ARRAYS_INCOMPAT_SIZE => Initialize(PRIMITIVE_ARRAYS_INCOMPAT_SIZE_CODE, "Primitive arrays split in two properties should have the same size.");
 
         // This leaves the incorrect nulls in place, no change in data.
-        internal static readonly FhirJsonException PRIMITIVE_ARRAYS_ONLY_NULL = new(PRIMITIVE_ARRAYS_ONLY_NULL_CODE, "Arrays need to have at least one non-null element.");
+        internal static FhirJsonException PRIMITIVE_ARRAYS_ONLY_NULL(ref Utf8JsonReader reader, string instancePath) => Initialize(ref reader, instancePath, PRIMITIVE_ARRAYS_ONLY_NULL_CODE, "Arrays need to have at least one non-null element.", OO_Sev.Warning, OO_Typ.Structure);
 
-        public FhirJsonException(string code, string message) : base(code, message)
+        public FhirJsonException(string code, string message, OperationOutcome.IssueSeverity issueSeverity, OperationOutcome.IssueType issueType) : base(code, message, issueSeverity, issueType)
```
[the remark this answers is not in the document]
brianpos: @ewoutkramer can I do this as another constructor and put a deprecated warning on the one without, encouraging folks to explicitly choose?

src/Hl7.Fhir.Base/Serialization/FhirJsonException.cs
```diff
@@ -57,78 +59,95 @@ public class FhirJsonException : CodedException
...
+        internal static FhirJsonException UNKNOWN_RESOURCE_TYPE(ref Utf8JsonReader reader, string instancePath, string resourceType) => Initialize(ref reader, instancePath, UNKNOWN_RESOURCE_TYPE_CODE, $"Unknown type '{resourceType}' found in 'resourceType' property.", OO_Sev.Fatal, OO_Typ.Structure);
+        internal static FhirJsonException RESOURCE_TYPE_NOT_A_RESOURCE(ref Utf8JsonReader reader, string instancePath, string name) => Initialize(ref reader, instancePath, RESOURCE_TYPE_NOT_A_RESOURCE_CODE, $"Data type '{name}' in property 'resourceType' is not a type of resource.", OO_Sev.Fatal, OO_Typ.Structure);
+        internal static FhirJsonException UNKNOWN_PROPERTY_FOUND(ref Utf8JsonReader reader, string instancePath, string propName) => Initialize(ref reader, instancePath, UNKNOWN_PROPERTY_FOUND_CODE, $"Encountered unrecognized element '{propName}'.", OO_Sev.Error, OO_Typ.Structure); // this could be ignored, so isn't fatal?
```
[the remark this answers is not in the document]
brianpos: This was the case where if you're parsing normative content and other parts are added in, that shouldn't be fatal, or else this is something that most places will choose to have a "mode" where this doesn't break things.
(I'm not attached to this if you want it changed to fatal - we probably need to document somewhere what the defaults are - I think you came down on is fatal if data is not able to be extracted into the object model)

src/Hl7.Fhir.Base/Serialization/FhirJsonException.cs
```diff
@@ -57,78 +59,95 @@ public class FhirJsonException : CodedException
...
+        internal static FhirJsonException UNEXPECTED_JSON_TOKEN(ref Utf8JsonReader reader, string instancePath, string expected, string actual, string? value) => Initialize(ref reader, instancePath, UNEXPECTED_JSON_TOKEN_CODE, $"Expecting a {expected}, but found a json {actual} with value '{value}'.", OO_Sev.Warning, OO_Typ.Value);
 
         // In R5 Integer64 (long) are serialized as string. So we would expect a string during parsing.
-        internal static readonly FhirJsonException LONG_CANNOT_BE_PARSED = new(LONG_CANNOT_BE_PARSED_CODE, "Json string '{0}' cannot be parsed as a {1}.");
-        internal static readonly FhirJsonException LONG_INCORRECT_FORMAT = new(LONG_INCORRECT_FORMAT_CODE, "{0} '{1}' cannot be parsed as a {2}, because it should be a {3}.");
+        internal static FhirJsonException LONG_CANNOT_BE_PARSED(ref Utf8JsonReader reader, string instancePath, string? s0, string typeName) => Initialize(ref reader, instancePath, LONG_CANNOT_BE_PARSED_CODE, $"Json string '{s0}' cannot be parsed as a {typeName}.", OO_Sev.Warning, OO_Typ.Value);
```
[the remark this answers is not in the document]
brianpos: This is the one that has the edge case on it, and if the string parses to a long then still want the warning - as strict modes will want to know about it (incorrect format) however there was no data loss as it got into the object model.

src/Hl7.Fhir.Base/Serialization/FhirJsonException.cs
```diff
@@ -57,78 +59,95 @@ public class FhirJsonException : CodedException
...
+                LineNumber = lineNumber,
+                Position = position,
+                Location = instancePath,
+            };
+        }
 
-            return new FhirJsonException(ErrorCode, message, inner);
+        public FhirJsonException? CloneWith(string baseMessage, OO_Sev issueSeverity, OO_Typ issueType)
+        {
+            string location = $"At line {this.LineNumber}, position {this.Position}";
```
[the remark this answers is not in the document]
brianpos: This wasn't so subtle, I refactored this a lot across both the XML and JSON parsers as it was duplicated there too!
And while there I renamed the property in the ExtendedCodedException from `Location` to `InstancePath` inline with everything else.

src/Hl7.Fhir.Base/Serialization/OperationOutcomeExtensions.cs
brianpos: I'm happy to leave these at that, the others I'm not so fussed on as unlikely to have been a result of user input, but from bad software.

src/Hl7.Fhir.Base/Serialization/FhirXmlException.cs
```diff
@@ -104,31 +124,27 @@ UNKNOWN_ELEMENT_CODE or
...
+            {
+                BaseErrorMessage = message,
+                LineNumber = lineNumber,
+                Position = position,
+                InstancePath = instancePath,
+            };
         }
-
     }
 }
```
ewoutkramer, line 149: INVALID_DUPLICATE_PROPERTY_CODE is missing in the IsRecoverableIssue

src/Hl7.Fhir.Base/Utility/ExtendedCodedException.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        /// <summary>
+        /// Type of issue to report in a FHIR OperationOutcome.
+        /// </summary>
+        public OperationOutcome.IssueType IssueType { get; init; } = OperationOutcome.IssueType.Invalid;
+
+        /// <summary>
+        /// The error message without any location information appended to it (which is in Exception.Message property).
+        /// </summary>
+        public string? BaseErrorMessage { get; init; }
```
ewoutkramer, line 62: The design of this property is incorrect. You can set this property on construction, but it will not change the exception at all. In fact, this message is supposed to be the base of the message you pass to the constructor (which includes location) etc, but that is not expressed anywhere structurally (it is in the comment).  This will for sure confuse the users.

I think we can improve on this: have the original constructors with just code  + message do the "old" behaviour (just for backwards compat, we'll not be using them anymore).  Then we add new constructors that explicitly take the code + *basemessage* + location + anything that influences the formatted message - that constructor will set the basemessage (if still necessary), construct the formatted message and pass that off to the base constructor.

This way you are either explicitly in backwards compat mode (and we can obsolete these constructors) OR in the new mode where you pass in the base message + the rest.

ewoutkramer: I'll try to push a version that expresses what I mean.

ewoutkramer (changes requested):

marcovisserFurore (approved):
The code has been reviewed and approved by @ewoutkramer.  And granted on his behalf.

src/Hl7.Fhir.Base/Utility/ExtendedCodedException.cs
ewoutkramer: Done.

Issue #2404, opened 2023-03-08
FhirPath function memberOf implementation


marcovisserFurore:
The FP function `memberOf` is not implemented yet. . Is there a way to implement this FP function?

brianpos:
I did a POC of this here
https://github.com/brianpos/fhirpath-lab-dotnet/blob/master/fhirpath-lab-dotnet/Helpers/FhirPathTerminologies.cs

ewoutkramer:
I don't really understand the way this is solved, @brianpos - I think we should just add a terminology service to the `FhirEvaluationContext` and invoke that service from the function. Am I missing something?

brianpos:
Provided that is able to reach out and call an external service too I'm good with that.
Admittedly, the code I've referenced here is for the other terminology related functions in the fhirpath spec, not this specific one.

PR #2450, opened 2023-05-03
#2404: Implementation of FhirPath function memberOf()

marcovisserFurore:
## Description
Implemented the FhirPath function `memberOf(valueset)`. See also the definition of the function [here](https://hl7.org/fhir/fhirpath.html#functions).

## Related issues
Resolves issue #2404.

## Testing
The following unittest are made:
- `Hl7.FhirPath.R4.Tests.FhirPathTest.MemberOfTests`
- `Hl7.FhirPath.R4.Tests.FhirPathTest.MemberOfTestsWithoutTerminologyService`
- `Hl7.FhirPath.R4.Tests.FhirPathTest.MemberOfTestWithExampleFromSpecification`

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Base/FhirPath/ElementNavFhirExtensions.cs
+++ b/src/Hl7.Fhir.Base/FhirPath/ElementNavFhirExtensions.cs
@@ -1,5 +1,16 @@
+/* 
+ * Copyright (c) 2023, Firely (info@fire.ly) and contributors
+ * See the file CONTRIBUTORS for details.
+ * 
+ * This file is licensed under the BSD 3-Clause license
+ * available at https://raw.githubusercontent.com/FirelyTeam/firely-net-sdk/master/LICENSE
+ */
+
+#nullable enable
+
 using Hl7.Fhir.ElementModel;
 using Hl7.Fhir.Model;
+using Hl7.Fhir.Rest;
 using Hl7.Fhir.Specification.Terminology;
 using Hl7.Fhir.Utility;
 using Hl7.FhirPath;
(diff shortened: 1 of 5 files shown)


src/Hl7.Fhir.Base/FhirPath/ElementNavFhirExtensions.cs
```diff
@@ -243,5 +255,54 @@ internal static P.Time BoundaryTime(P.Time time, long? precision, int minutes, i
...
+        {
+            var service = (ctx is FhirEvaluationContext fctx ? fctx.TerminologyService : null)
+                ?? throw new ArgumentNullException("The 'memberOf' function cannot be executed because the FhirEvaluationContext does not include a TerminologyService.");
+
+            ValidateCodeParameters? inParams = new ValidateCodeParameters()
+                        .WithValueSet(valueset);
+
+            inParams = input.InstanceType switch
+            {
+                "code" when input is ScopedNode sn => inParams.WithCode(code: sn.Value as string, context: sn.LocalLocation),
```
ewoutkramer, line 282: The `LocalLocation` will look something like `Patient.name[1].family[2]` etc. while according to http://hl7.org/fhir/valueset-operation-validate-code.html, the "context" that we need to pass in is defined as 

> The recommended format for this URI is [Structure Definition URL]#[name or path into structure definition] e.g. http://hl7.org/fhir/StructureDefinition/observation-hspc-height-hspcheight#Observation.interpretation

This is obviously written from the viewpoint of a profile validator, since that would call the service with full information about the profile (Note: even our profile validator does not pass in the canonical as far as I know).
In a "random" fhirpath evaluation on an instance, we would never have a profile uri (at most the canonical to the core definition), so we cannot really align with this requirement.   

Finally, the `LocalPath` has indices for repeating elements, which the context path is not supposed to have. Do we put our head in the sand here for now?

ewoutkramer: Yes, let's just leave this as is and go for it.

ewoutkramer (approved):

PR #2470, opened 2023-05-23
Added MimeType Terminology Service

### LINKED ISSUE #1055: Terminology service raises warningi for mimetypes
When adding a mimetype (EG: DocumentReference.content.attachment.contentType) the terminology service raises a warning. For example, when using the "text/plain" mimetype, the service raises this warning:

`Terminology service failed while validating code 'text/plain' (system ''): Cannot resolve canonical reference 'urn:ietf:bcp:13' to CodeSystem`

Is this expected behavior? It is just a warning so we can ignore it, but it would be nice if there was a way to remove this warning from being raised.

mmsmits:
Fixes #1055

Todo: refactor some duplicate code with local terminology service

### DIFF
--- a/src/Hl7.Fhir.Base/Model/ParametersExtensions.cs
+++ b/src/Hl7.Fhir.Base/Model/ParametersExtensions.cs
@@ -1,23 +1,44 @@
-﻿using Hl7.Fhir.Model;
+﻿using Hl7.Fhir.Rest;
 using Hl7.Fhir.Utility;
-using System;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
+using System.Net;
 
 namespace Hl7.Fhir.Model
 {
     public static class ParametersExtensions
     {
-        public static bool TryGetDuplicates(this Parameters parameters , out IEnumerable<string> duplicates)
+        private const string CODEATTRIBUTE = "code";
+        private const string SYSTEMATTRIBUTE = "system";
+        private const string CONTEXTATTRIBUTE = "context";
+
+        public static bool TryGetDuplicates(this Parameters parameters, out IEnumerable<string> duplicates)
         {
-            duplicates =  parameters.Parameter?.Select(p => p.Name)?
+            duplicates = parameters.Parameter?.Select(p => p.Name)?
                           .GroupBy(x => x)
                           .Where(g => g.Count() > 1)
                           .Select(y => y.Key)
                           .ToList();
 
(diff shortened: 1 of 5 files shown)


src/Hl7.Fhir.Base/Model/ParametersExtensions.cs
```diff
@@ -1,23 +1,44 @@
...
+            {
+                //422 Unproccesable Entity
+                throw new FhirOperationException($"List of input parameters contains the following duplicates: {string.Join(", ", duplicates)}", (HttpStatusCode)422);
+            }
+            //If a code is provided, a system or a context must be provided (http://hl7.org/fhir/valueset-operation-validate-code.html)
+            if (parameters.Parameter.Any(p => p.Name == CODEATTRIBUTE) && !(parameters.Parameter.Any(p => p.Name == SYSTEMATTRIBUTE) ||
+                                                                                    parameters.Parameter.Any(p => p.Name == CONTEXTATTRIBUTE)))
+            {
+                //422 Unproccesable Entity
+                throw new FhirOperationException($"If a code is provided, a system or a context must be provided", (HttpStatusCode)422);
```
ewoutkramer, line 40: Isn't there a `HttpStatusCode.UnprocessableEntity` ?

src/Hl7.Fhir.Base/Specification/Terminology/MimeTypeTerminologyService.cs
```diff
@@ -0,0 +1,156 @@
...
+                var success = validateMimeType(code);
+
+                if (success)
+                {
+                    result.Add("result", new FhirBoolean(true));
+                }
+                else
+                {
+                    result.Add("result", new FhirBoolean(false))
+                          .Add("message", new FhirString("Invalid mimetype specified"));
```
ewoutkramer, line 135: I'd change this to `XXXXX is not a valid mimetype.`  Always end errors with periods, it's a message for humans.

ewoutkramer (changes requested):
Just very small things. In general: make the error messages end with a period ;-)

src/Hl7.Fhir.Base/Model/ParametersExtensions.cs
mmsmits: Only in .NET 6.0 and up.

ewoutkramer (approved):

PR #2474, opened 2023-05-25
1340 Improve support for Binary endpoint

ewoutkramer:
Add streaming binary support to the FhirClient.


ewoutkramer:
I am archiving this PR - we know its there ;-)  Now do we?

mmsmits:
Should we put this somewhere on the backlog then?

PR #2492, opened 2023-06-05
Re-generate model with default codesystems

mmsmits:


ewoutkramer (changes requested):
I think we should keep the definition of the metadata of the enum in EnumMapping unchanged - we can keep it separate from the way we read it from the attributes. Not introducing a default system in the EnumMapping keeps the enummapping simpler to work with (less to worry about), and if we later find something smarter to put in the attributes, we don't have to change the EnumMapping again.

Put otherwise, the shortcut notation in the attributes should be "expanded" into the EnumMapping when we read the metadata.  To make sure we don't waste memory on repeating the same string internally, we can use `String.Internalize()`.

ewoutkramer (approved):

PR #2506, opened 2023-06-19
Support STU3 mime type vs url in MimeTypeTerminologyService

### LINKED ISSUE #2502: MimeTypeTerminologyService cannot handle ValueSet canonicals with version
**Describe the bug**
Currently, the terminology service `MimeTypeTerminologyService` is unable to handle ValueSet canonicals that include versions. Consequently, when the `Url` parameter is set to `http://hl7.org/fhir/ValueSet/mimetypes|4.0.1`, it throws a `FhirOperationException` with the following error message: "Cannot find valueset 'http://hl7.org/fhir/ValueSet/mimetypes|4.0.1'".

**Expected behavior**
Ideally, the `MimeTypeTerminologyService` should be modified to accommodate ValueSet canonicals with versions. Therefore, it should perform a check on the incoming Url, specifically `http://hl7.org/fhir/ValueSet/mimetype`, excluding the version.

**Version used:**
- FHIR Version: all
- Version: 5.2.0

### LINKED ISSUE #2504: MimeTypeTerminologyService should handle STU3 canonical for MimeType ValueSet
In STU3 the canonical for the MimeType ValueSet, e.g. in Attachment.contentType is "http://www.rfc-editor.org/bcp/bcp13.txt" not "http://hl7.org/fhir/ValueSet/mimetypes". This should be handled by the the terminology service as an additional canonical. Note that this is a bit more tricky as there is not a ValueSet resource defined by the spec itself. It's just a binding on the datatype itself as far as I can see. This might cause some issues in the VS resolver.

mmsmits:
Also support versioned valueset canonicals

fixes #2502
fixes #2504

### DIFF
--- a/src/Hl7.Fhir.Base/Specification/Terminology/MimeTypeTerminologyService.cs
+++ b/src/Hl7.Fhir.Base/Specification/Terminology/MimeTypeTerminologyService.cs
@@ -17,7 +17,7 @@ namespace Hl7.Fhir.Specification.Terminology
     public class MimeTypeTerminologyService : ITerminologyService
     {
         private const string MIMETYPE_SYSTEM = "urn:ietf:bcp:13";
-        private const string MIMETYPE_VALUESET = "http://hl7.org/fhir/ValueSet/mimetypes";
+        private const string MIMETYPE_VALUESET_R4_AND_UP = "http://hl7.org/fhir/ValueSet/mimetypes";
 
 
         ///<inheritdoc />
@@ -41,8 +41,9 @@ namespace Hl7.Fhir.Specification.Terminology
             parameters.CheckForValidityOfValidateCodeParams();
 
             var validCodeParams = new ValidateCodeParameters(parameters);
+            var valueSetUri = validCodeParams?.Url?.Value != null ? new Canonical(validCodeParams?.Url?.Value)?.Uri : null;
 
-            if (validCodeParams?.Url?.Value is not null && validCodeParams?.Url?.Value != MIMETYPE_VALUESET)
+            if (valueSetUri is not null && valueSetUri != MIMETYPE_VALUESET_R4_AND_UP && valueSetUri != MIMETYPE_VALUESET_STU3)
             {   // 404 not found
                 throw new FhirOperationException($"Cannot find valueset '{validCodeParams!.Url.Value}'", HttpStatusCode.NotFound);
             }
@@ -114,6 +115,7 @@ namespace Hl7.Fhir.Specification.Terminology
         private static Task<Parameters> validateCodeVS(string? code, string? system)
         {
             var result = new Parameters();
+            var systemUri = system != null ? new Canonical(system)?.Uri : null;
 
             if (code is null)
             {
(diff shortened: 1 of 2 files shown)


src/Hl7.Fhir.Base/Specification/Terminology/MimeTypeTerminologyService.cs
```diff
@@ -41,8 +41,9 @@ public async T.Task<Parameters> ValueSetValidateCode(Parameters parameters, stri
             parameters.CheckForValidityOfValidateCodeParams();
 
             var validCodeParams = new ValidateCodeParameters(parameters);
+            var valueSetUri = validCodeParams?.Url?.Value != null ? new Canonical(validCodeParams?.Url?.Value)?.Uri : null;
 
-            if (validCodeParams?.Url?.Value is not null && validCodeParams?.Url?.Value != MIMETYPE_VALUESET)
+            if (valueSetUri is not null && valueSetUri != MIMETYPE_VALUESET_R4_AND_UP && valueSetUri != MIMETYPE_VALUESET_STU3)
```
marcovisserFurore, line 46: Actually the condition `valueSetUri is not null` is not really needed here: `null !=  "http://hl7.org/fhir/ValueSet/mimetypes"` will also do the trick

src/Hl7.Fhir.Base/Specification/Terminology/MimeTypeTerminologyService.cs
```diff
@@ -114,14 +115,15 @@ private async Task<Parameters> validateCodeVS(CodeableConcept cc)
         private static Task<Parameters> validateCodeVS(string? code, string? system)
         {
             var result = new Parameters();
+            var systemUri = system != null ? new Canonical(system)?.Uri : null;
```
marcovisserFurore, line 118: I think the `?` in `new Canonical(system)?.Uri` is not needed here: after construction it cannot be null...

src/Hl7.Fhir.Support.Tests/Specification/MimeTypeTerminologyServiceTests.cs
```diff
@@ -52,5 +54,51 @@ public async Task MimeTypeValidationTest()
...
+                   .WithCode(code: "application/json", context: "context")
+                   .Build();
+
+            result = await _service.ValueSetValidateCode(parameters);
+            result.Parameter.Should().Contain(p => p.Name == "result")
+                .Subject.Value.Should().BeEquivalentTo(new FhirBoolean(true));
+        }
+
+        [TestMethod]
+        public async Task MimeTypeValidationWithStu3VS()
```
marcovisserFurore, line 82: Almost exactly the same code as `MimeTypeValidationWithVersionedVSTest`. Can you make this a unit test with `DataRow`s? The variable is here the ValueSet.

marcovisserFurore (changes requested):
Some small changes required.

marcovisserFurore (approved):
Perfect!

PR #2507, opened 2023-06-20
Improve FhirPath expression cache unit test to not use timers anymore

mmsmits:


marcovisserFurore (changes requested):
Nicely done. It makes the unit test much more robust. 

Only 1 thing: the reflection is done multiple times. Maybe better to do it only once in the constructor or Initialize().

marcovisserFurore (approved):

PR #2535, opened 2023-07-13
Add routing mechanism to the MultiTerminologyService

### LINKED ISSUE #2518: Harmonize MultiTerminologyService
Currently, the `MultiTerminologyService` in Firely Server faces usability issues. It lacks a routing system for valuesets, which determines the preferred ValuesetsUrl for each TerminologyService. To address this, an implementation example can be found in the new validator, specifically the [CompositeValidateCodeService](https://github.com/FirelyTeam/firely-car/blob/35122cbd95fd54ec06f0ce2bacb32f042dd42e89/src/Firely.Validator.WebAPI/Terminology/CompositeValidateCodeService.cs#L1). Incorporating a similar routing system would greatly enhance the functionality and usability of the `MultiTerminologyService` in Firely Server.

mmsmits:
fixes #2518

### DIFF
--- a/src/Hl7.Fhir.Base/Specification/Terminology/MultiTerminologyService.cs
+++ b/src/Hl7.Fhir.Base/Specification/Terminology/MultiTerminologyService.cs
@@ -21,7 +21,7 @@ namespace Hl7.Fhir.Specification.Terminology
         /// A collection of multiple terminology services to allow for one or multiple fallback services when validating codes for example.
         /// </summary>
         /// <param name="services">Terminology services to be used. Note that the first service in the list will be used first, the others will be used for fallback based on their order.</param>
-        public MultiTerminologyService(IEnumerable<ITerminologyService> services) : this(services.Select(s => TerminologyServiceRoutingSettings.CreateDefault(s))) { }
+        public MultiTerminologyService(IEnumerable<ITerminologyService> services) : this(services.Select(s => new TerminologyServiceRoutingSettings(s))) { }
 
         /// <summary>
         /// A collection of multiple terminology services to allow for one or multiple fallback services when validating codes for example.
@@ -130,7 +130,9 @@ namespace Hl7.Fhir.Specification.Terminology
             List<FhirOperationException> exceptions = new();
 
             // check if any preferedServices
-            var preferred = preferredService(parameters);
+            var inputVsUrl = new ValidateCodeParameters(parameters).Url?.Value;
+            var preferred = (inputVsUrl != null) ? preferredService(inputVsUrl) : Enumerable.Empty<ITerminologyService>();
+
             var services = _termServices.Select(s => s.Service);
             var orderedTermServices = preferred.Any() ?
                                       reorderList(services, preferred)
@@ -164,12 +166,10 @@ namespace Hl7.Fhir.Specification.Terminology
             throw new InvalidOperationException("We should never have come here");
         }
 
-        private IEnumerable<ITerminologyService> preferredService(Parameters parameters)
+        private IEnumerable<ITerminologyService> preferredService(string inputVsUrl)
         {
-            var valueSetUrl = new ValidateCodeParameters(parameters).Url?.Value;
-            return valueSetUrl != null
-                ? _termServices.Where(t => matchSystem(t.PreferredValueSets, valueSetUrl)).Select(t => t.Service)
-                : Enumerable.Empty<ITerminologyService>();
+            return _termServices.Where(t => matchSystem(t.PreferredValueSets, inputVsUrl)).Select(t => t.Service);
+
 
             static bool matchSystem(IEnumerable<string>? preferredSystems, string valueSetUrl) =>
                 preferredSystems?.Any() == true && preferredSystems.Any(valueSetUrl.StartsWith);


src/Hl7.Fhir.Support.Tests/Specification/MultiTerminologyServiceTests.cs
```diff
@@ -72,11 +72,63 @@ public async Task MultipleTerminologyServicesErrorHandlingTest()
...
+            inputParams = new ValidateCodeParameters().WithValueSet("http://example.org/fhir/ValueSet/second-preference").Build();
+            firstFailingTS = createOrderableTerminologyService(setupMockTermService(inputParams, firstFailingParameters), 0, "http://example.org/fhir/ValueSet/first-preference");
+            secondFailingTS = createOrderableTerminologyService(setupMockTermService(inputParams, secondFailingParameters), 1, "http://example.org/fhir/ValueSet/second-preference");
+
+            //run tests
+            multits = new MultiTerminologyService(firstFailingTS, secondFailingTS);
+            result = await multits.ValueSetValidateCode(inputParams);
+
+            //check results
+            result.Parameter.Where(p => p.Name == "result").FirstOrDefault().Value.As<FhirBoolean>().Value.Should().Be(false);
```
marcovisserFurore, line 111: Does this not raise a nullability warning (`.FirstOrDefault().Value`) ?

src/Hl7.Fhir.Support.Tests/Specification/MultiTerminologyServiceTests.cs
```diff
@@ -72,11 +72,63 @@ public async Task MultipleTerminologyServicesErrorHandlingTest()
             await call.Should().ThrowAsync<FhirOperationException>().Where(e => ((AggregateException)e.InnerException).InnerExceptions.Count == 2);
         }
 
-        private ITerminologyService setupMockTermService(Parameters output)
+        [TestMethod]
+        public async Task OrderableTerminologyServiceTest()
```
marcovisserFurore, line 76: This will test the happy flow. But what if 
- there are no TermServs registered at all?
- all the TermServs are not matching?

src/Hl7.Fhir.Base/Specification/Terminology/MultiTerminologyService.cs
```diff
@@ -15,19 +15,85 @@ namespace Hl7.Fhir.Specification.Terminology
...
+            Add(settings);
+        }
+
+        /// <summary>
+        /// Add an <see cref="ITerminologyService"/> including their routing settings at the back of the order of terminologyServices to be used.
+        /// </summary>
+        /// <param name="settings"><see cref="TerminologyServiceRoutingSettings"/> to be added.</param>
+        public void Add(TerminologyServiceRoutingSettings settings)
+        {
+            var position = _termServices.Keys.Last() + 1;
```
ewoutkramer, line 72: Since the order is now based on a automatically assigned sequential number and it is not possible to specifity a prio via the settings, the Keys in the OrderedLst will always just be 1...N.  This means you could also use a normal List now, and Insert() at the beginning or Add at the end!

src/Hl7.Fhir.Base/Specification/Terminology/MultiTerminologyService.cs
```diff
@@ -15,19 +15,96 @@ namespace Hl7.Fhir.Specification.Terminology
...
+        public MultiTerminologyService(params TerminologyServiceRoutingSettings[] services) : this(services.ToList()) { }
+
+
+        /// <summary>
+        /// Add an <see cref="ITerminologyService"/> at the back of the order of terminologyServices to be used.
+        /// </summary>
+        /// <param name="service"><see cref="ITerminologyService"/> to be added.</param>
+        public void AddLast(ITerminologyService service)
+        {
+            var settings = new TerminologyServiceRoutingSettings(service);
```
ewoutkramer, line 56: Is there a difference between `TerminologyServiceRoutingSettings.Default(service)` and `new TerminologyServiceRoutingSettings(service)`?  You are using both forms.

src/Hl7.Fhir.Base/Specification/Terminology/MultiTerminologyService.cs
```diff
@@ -15,19 +15,96 @@ namespace Hl7.Fhir.Specification.Terminology
...
+        /// </summary>
+        /// <param name="index"></param>
+        /// <param name="settings"><see cref="TerminologyServiceRoutingSettings"/> to be added.</param>
+        public void Insert(int index, TerminologyServiceRoutingSettings settings)
+        {
+            _termServices.Insert(index, settings);
+        }
 
         ///<inheritdoc/>
         public Task<Resource> Closure(Parameters parameters, bool useGet = false) => throw new NotImplementedException();
```
ewoutkramer, line 110: Why is this the `Closure()` function, not the `ValidateCode()` function?

src/Hl7.Fhir.Base/Specification/Terminology/MultiTerminologyService.cs
```diff
@@ -78,6 +163,22 @@ public async Task<Parameters> ValueSetValidateCode(Parameters parameters, string
 
             throw new InvalidOperationException("We should never have come here");
         }
+
+        private IEnumerable<ITerminologyService> preferredService(Parameters parameters)
+        {
+            var valueSetUrl = new ValidateCodeParameters(parameters).Url?.Value;
```
ewoutkramer, line 169: This function validates the parameters and then gets the right ITerminologyService - this sounds like a function with "low cohesion", since these are two completely different tasks.  E.g. what if the calling function also needs to get other parameters from the Parameters?  It would need to validate the parameters before it uses them, right?  So, I think you should separate the concerns of validating the parameters and getting the right iterminologyservi ce.

ewoutkramer (changes requested):
I also added two separate comments.

ewoutkramer (approved):

marcovisserFurore (approved):

PR #2536, opened 2023-07-14
Refactored TransactionBuilder to include fullUrl in Bundle.Entry for every operation

### LINKED ISSUE #2115: Add fullUrl to Bundle entry with TransactionBuilder
When adding entries to the TransactionBuilder, there is no option to set a fullUrl, and the TransactionBuilder logic does not add the fullUrl itself. This means that you need to convert to a Bundle and then run through all the entries to add the fullUrls by hand after the transaction has been setup.

I would like an overload to the TransactionBuilder methods to be able to set the fullUrl when needed.

marcovisserFurore:
## Description
Each operation of the `TransactionBuilder`, like `Read` for example, has now an extra parameter to include a `fullUrl` for the `Bundle.Entry`.

## Related issues
Resolves #2115 .

## Testing
See unit test `Hl7.Fhir.Test.TransactionBuilderTests.TransactionBuilderWithFullUrlTest()`

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Base/Rest/BaseFhirClient.cs
+++ b/src/Hl7.Fhir.Base/Rest/BaseFhirClient.cs
@@ -45,8 +45,8 @@ namespace Hl7.Fhir.Rest
         /// <param name="inspector"></param>
         /// <param name="settings"></param>
         public BaseFhirClient(Uri endpoint, HttpMessageHandler? messageHandler, ModelInspector inspector, FhirClientSettings? settings = null)
-            : this(endpoint, inspector, 
-                  settings ?? new(), 
+            : this(endpoint, inspector,
+                  settings ?? new(),
                   new(endpoint, (settings ?? new()).Timeout, messageHandler ?? makeDefaultHandler(), messageHandler == null))
         {
             // Nothing
@@ -603,7 +603,7 @@ namespace Hl7.Fhir.Rest
 
 
             var tx = new TransactionBuilder(Endpoint);
-            var resourceType = Inspector.GetFhirTypeNameForType(typeof(TResource));
+            var resourceType = Inspector.GetFhirTypeNameForType(typeof(TResource)) ?? throw new ArgumentException($"{typeof(TResource).Name} is not recognized as a valid FHIR resource.");
 
             if (!string.IsNullOrEmpty(versionId))
                 tx.Patch(resourceType, id, patchParameters, versionId);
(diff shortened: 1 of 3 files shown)


src/Hl7.Fhir.Base/Rest/BaseFhirClient.cs
```diff
@@ -603,7 +603,7 @@ public virtual void Delete(string resourceType, SearchParams condition)
 
 
             var tx = new TransactionBuilder(Endpoint);
-            var resourceType = Inspector.GetFhirTypeNameForType(typeof(TResource));
+            var resourceType = Inspector.GetFhirTypeNameForType(typeof(TResource)) ?? throw new ArgumentException($"{typeof(TResource).Name} is not recognized as a valid FHIR resource.");
```
ewoutkramer, line 606: there is a private method `typeNameOrDie` that does this - other methods use this private method to provide a consistent exception.

src/Hl7.Fhir.Base/Rest/BaseFhirClient.cs
```diff
@@ -632,7 +632,7 @@ public virtual void Delete(string resourceType, SearchParams condition)
         public Task<TResource?> PatchAsync<TResource>(SearchParams condition, Parameters patchParameters, CancellationToken? ct = null) where TResource : Resource
         {
             var tx = new TransactionBuilder(Endpoint);
-            var resourceType = Inspector.GetFhirTypeNameForType(typeof(TResource));
+            var resourceType = Inspector.GetFhirTypeNameForType(typeof(TResource)) ?? throw new ArgumentException($"{typeof(TResource).Name} is not recognized as a valid FHIR resource.");
```
ewoutkramer, line 635: use `typeNameOrDie()`

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/Rest/BaseFhirClient.cs
marcovisserFurore: Ah ok. I was not aware of this. It was a proposal of Visual Studio to do this, and I agreed...

src/Hl7.Fhir.Base/Rest/BaseFhirClient.cs
marcovisserFurore: Ah ok. I was not aware of this. It was a proposal of Visual Studio to do this, and I agreed...

ewoutkramer (approved):

Issue #2205, opened 2022-08-17
Let resources having an Identifier extend a common interface


taconaut:
I have to access the `Identifier` property of different resources (`MedicationStatement`, `MedicationDispense`, `MedicationRequest`, `Observation`). This property is not on a base class or available on an interface, so I have to handle every case explicitely, whereas I'd like to make it generic.

Could you add an interface `IIdentifiable` (like the already existing `IExtendable`) and make all classes implementing `Identifier` extend it?

Here's the code I've written, which led me to opening this feture request:
```c#
    private static bool IsResourceContainedInRetrievals(DomainResource resource, IEnumerable<MedicamentRetrieval> retrievals)
    {
        if (resource is MedicationStatement medicationStatement)
            return IsMedicationStatementContainedInRetrievals(medicationStatement, retrievals);

        if (resource is MedicationDispense medicationDispense)
            return IsMedicationDispenseContainedInRetrievals(medicationDispense, retrievals);

        if (resource is MedicationRequest medicationRequest)
            return IsMedicationRequestContainedInRetrievals(medicationRequest, retrievals);

        if (resource is Observation observation)
            return IsObservationContainedInRetrievals(observation, retrievals);

        throw new NotSupportedException($"Resource of type='{(resource == null ? "NULL" : resource.GetType().Name)}' is not supported");
    }

    private static bool IsMedicationStatementContainedInRetrievals(MedicationStatement medicationStatement, IEnumerable<MedicamentRetrieval> retrievals) =>
        retrievals.Select(retrieval => retrieval.Resource)
                  .OfType<MedicationStatement>()
                  .Any(ms => medicationStatement.Identifier.Single(id => id.System == FhirConstants.SystemIdentifier.ResourceIdentifier).Value
                          == ms.Identifier.Single(id => id.System == FhirConstants.SystemIdentifier.ResourceIdentifier).Value);

    private static bool IsMedicationDispenseContainedInRetrievals(MedicationDispense medicationDispense, IEnumerable<MedicamentRetrieval> retrievals) =>
        retrievals.Select(retrieval => retrieval.Resource)
                  .OfType<MedicationDispense>()
                  .Any(md => medicationDispense.Identifier.Single(id => id.System == FhirConstants.SystemIdentifier.ResourceIdentifier).Value
                          == md.Identifier.Single(id => id.System == FhirConstants.SystemIdentifier.ResourceIdentifier).Value);

[...]

PR #2496, opened 2023-06-08
Add tool to apply interfaces on generated classes

taconaut:
## Description
This PR adds the possibility to dynamically add interfaces to generated classes.
See [InterfaceApplier README](https://github.com/taconaut/firely-net-sdk/blob/feature/apply-interface-on-code-generate/src/Hl7.Fhir.InterfaceApplier.CLI/README.md) for a brief description of the tool
See [InterfacesAppliedToGeneratedClasses README](https://github.com/taconaut/firely-net-sdk/blob/feature/apply-interface-on-code-generate/src/Hl7.Fhir.Base/Model/InterfacesAppliedToGeneratedClasses/README.md) for a brief description of how to add support for additional interfaces.

### Execution example
here's an example output of the performed operations when running the new powershell script:
```
PS D:\Dev\firely-net-sdk\build> .\apply-interfaces-on-generated-classes.ps1
MSBuild version 17.5.1+f6fdcf537 for .NET
  Determining projects to restore...
  Restored D:\Dev\firely-net-sdk\src\Hl7.Fhir.InterfaceApplier.CLI\Hl7.Fhir.InterfaceApplier.CLI.csproj (in 218 ms).
  Hl7.Fhir.InterfaceApplier.CLI -> D:\Dev\firely-net-sdk\src\Hl7.Fhir.InterfaceApplier.CLI\bin\Debug\net6.0\win-x64\Int
  erfaceApplier.dll
  Hl7.Fhir.InterfaceApplier.CLI -> D:\Dev\firely-net-sdk\build\tools\
[14:55:19 INF] *** Executing InterfaceApplier ***
[14:55:19 INF] Start applying interfaces on classes for project D:\Dev\firely-net-sdk\src\Hl7.Fhir.R4\Hl7.Fhir.R4.csproj[14:55:32 INF] Published project to D:\Dev\firely-net-sdk\src\Hl7.Fhir.R4\bin\Debug\InterfaceApplier
[14:55:32 INF] Loaded assemblies from published project:
   - Fhir.Metrics, Version=1.2.2.0, Culture=neutral, PublicKeyToken=43e377ac229b6873
   - Hl7.Fhir.Base, Version=5.1.1.0, Culture=neutral, PublicKeyToken=null
   - Hl7.Fhir.Conformance, Version=5.1.1.0, Culture=neutral, PublicKeyToken=null
   - Hl7.Fhir.R4, Version=5.1.1.0, Culture=neutral, PublicKeyToken=null
[...]

PR #2539, opened 2023-07-17
2205 Add IIdentifiable<T> and ICoded<T> to POCOs

ewoutkramer:
This provides a consistent interface to retrieve (and set) identifiers and codes on identifiable and coded poco's .

* Identifiable POCOs are poco's that have an element called "Identifier". 
* Coded POCO's are poco's that have a "primary code path" according to the FHIR binding for CQL. Many times, these are POCO's with a "Code" element, but it could also be called differently.  Because the CQL binding is only specified for R4, these interfaces are only used on R4 POCO's (and POCO's in common).


ewoutkramer:
This PR contains the same functionality as #2496, but uses codegen directly instead of post-processing of the source files after codegen.

ewoutkramer:
Fixes #2205

src/Hl7.Fhir.Base/Model/Generated/Coding.cs
```diff
@@ -121,7 +121,6 @@ public string Version
     /// Symbol in syntax defined by the system
     /// </summary>
     [FhirElement("code", InSummary=true, Order=50)]
-    [CqlElement(IsPrimaryCodePath = true)]
```
marcovisserFurore, line 124: Should this class not implementing `ICoded` as well?

src/Hl7.Fhir.Base/Model/Generated/Quantity.cs
```diff
@@ -225,7 +225,6 @@ public string System
     /// Coded form of the unit
     /// </summary>
     [FhirElement("code", InSummary=true, Order=70)]
-    [CqlElement(IsPrimaryCodePath = true)]
```
marcovisserFurore, line 228: Should this class not implementing `ICoded` as well?

src/Hl7.Fhir.Base/Model/Generated/UsageContext.cs
```diff
@@ -59,7 +59,7 @@ public partial class UsageContext : Hl7.Fhir.Model.DataType
     /// Type of context being specified
     /// </summary>
     [FhirElement("code", InSummary=true, Order=30)]
-    [CqlElement(IsPrimaryCodePath=true)]
```
marcovisserFurore, line 62: Should this class not implementing `ICoded` as well?

marcovisserFurore (changes requested):
Some remarks about missing interfaces....

src/Hl7.Fhir.Base/Model/Generated/Coding.cs
ewoutkramer: No, ICoded<T> should only be used on resources - this is a mistake in the CQL specification and I have corrected it here.

marcovisserFurore (approved):

taconaut:
Thanks 👍 , I'll check you solved this.

PR #2541, opened 2023-07-18
2534 Improve performance of Date handling

ewoutkramer:
Replicated changes from DateTime to Date.

### DIFF
--- a/src/Hl7.Fhir.Base/Model/Date.cs
+++ b/src/Hl7.Fhir.Base/Model/Date.cs
@@ -28,8 +28,8 @@
 
 */
 
-using Hl7.Fhir.Serialization;
 using System;
+using P = Hl7.Fhir.ElementModel.Types;
 
 #nullable enable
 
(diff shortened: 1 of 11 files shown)


src/Hl7.Fhir.Support.Poco.Tests/Model/DateTests.cs
```diff
@@ -0,0 +1,152 @@
...
+            for (var i = 0; i < 1000; i++)
+            {
+                _ = dt.TryToDate(out var _);
+            }
+            sw2.Stop();
+
+            Console.WriteLine(sw2.Elapsed.ToString());
+
+            // It's actually about 20x faster on my machine
+            (sw2.Elapsed).Should().BeLessThan(sw.Elapsed);
```
marcovisserFurore, line 149: These kind of tests can make the build server unstable.  When there is a drop of machine capacity in the second part of this unit test, the test could fail.

src/Hl7.FhirPath.Tests/Tests/BasicFunctionTests.cs
```diff
@@ -186,6 +185,7 @@ public void TestSubstring()
         public void TestExpressionTodayFunction()
         {
             // Check that date comes in
+            var s = scalar("today()");
```
marcovisserFurore, line 188: I think this was for debugging only.

marcovisserFurore (changes requested):
See my remarks..

src/Hl7.Fhir.Base/Model/Date.cs
```diff
@@ -38,41 +38,114 @@ namespace Hl7.Fhir.Model
...
+        {
+            _parsedValue = null;
+            base.OnObjectValueChanged();
+        }
+
+        /// <summary>
+        /// Converts this Fhir Fhir Date to a <see cref="DateTimeOffset"/>.
+        /// </summary>
+        /// <returns>A DateTimeOffset filled out to midnight, january 1 (UTC) in case of a partial date.</returns>
+        public DateTimeOffset ToDateTimeOffset()
```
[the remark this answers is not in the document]
ewoutkramer: You are right!  I (inadvertedly) copied this behaviour from FhirDateTime:

```
  public DateTimeOffset ToDateTimeOffset(TimeSpan zone)
        {
            if (Value == null) throw new InvalidOperationException("FhirDateTime's value is null");
```

Ouch, we're very inconsistent here - and this is impossible to fix without breaking the one or the other, so for now we'll have to keep this inconsistent behaviour.

src/Hl7.Fhir.Support.Poco.Tests/Model/DateTests.cs
ewoutkramer: That's why I kept a huge margin. I'll see if I can turn this into benchmarks instead.

src/Hl7.FhirPath.Tests/Tests/BasicFunctionTests.cs
ewoutkramer: Correct.

src/Hl7.Fhir.Base/Model/Date.cs
```diff
@@ -38,41 +38,113 @@ namespace Hl7.Fhir.Model
...
+        /// Gets the current date in UTC.
         /// </summary>
-        /// <returns>Gets the current date in the timezone UTC</returns>
-        public static Date UtcToday() => new(DateTimeOffset.UtcNow.ToString("yyyy-MM-dd"));
+        public static Date UtcToday() => FromDateTimeOffset(DateTimeOffset.UtcNow);
+
+        [NonSerialized]  // To prevent binary serialization from serializing this field
+        private P.Date? _parsedValue = null;
+
+        private static readonly P.Date INVALID_VALUE = P.Date.Today();
```
brianpos, line 69: This would resolve to the current value on initialization of the app, hence long running services would have the boot time in this value, shouldn't it be a function so it is evaluated each time it is called?

ewoutkramer: No. `hasInvalidParsedValue()` uses `ReferenceEquals()` to compare the cached time to INVALID_VALUE, so the actual value does not matter: it could have been mine or your birthdate as well. It's used here as a sentinel value to just indicate the value is invalid (as opposed to not cached (null)). It avoids keeping a separate bool `hasInvalidValue` that I need to keep synchronized.

GinoCanessa: In case someone is looking at the data, would it be better to use something like DateTime.MinValue to make it obvious?

Issue #2555, opened 2023-08-01
Add support for multiple valuesets and system filters to the ValueSetExpander


ewoutkramer:
The CQL engine uses valuesets, and some of the valuesets in its test set have multiple compose.include.valueset.  We do not support that yet. In order to get the SDK's terminology stuff into the CQL engine (and not break the unit tests), we need to add support for this feature.

Specifically, from https://www.hl7.org/fhir/valueset.html#compositions, do this:

* valueSet(s) only: Codes are 'selected' for inclusion if they are in all the referenced value sets
* valueSet and System: Codes are 'selected' for inclusion if they are selected by the code system selection (after checking for concept and filter) and if they are in all the referenced value sets

We currently only accept a single referenced value set (while the first bullet requires value set*s*), and we don't use the system to limit the codes coming in from the valueset isn the second bullet.

mmsmits:
Fixed by: https://github.com/FirelyTeam/firely-net-sdk/pull/2558

PR #2559, opened 2023-08-02
Move some termservice code to shared projects

ewoutkramer:
This will make sure both the new functionality and the new tests for #2555 are included in all versions.


ewoutkramer:
Review this after #2555, then pull together with #2555.

PR #2582, opened 2023-09-06
VONK-5093: introducing IScopedNode

marcovisserFurore:
## Description
In this pull request (PR), we are introducing the `IScopedNode` interface, a derivative of `IBaseElementNavigator`. Essentially, it mirrors `ITypedElement` but excludes the location and definition members.

The purpose of this new `IScopedNode` interface is its integration into the upcoming validator and potentially into FhirPath at a later stage.

## Issue
This PR resolves VONK-5093. 

## Remarks
- This is the time to think about the naming of the interface `IScopedNode` and `IBaseElementNavigator`


src/Hl7.Fhir.Base/ElementModel/IScopedNodeExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        /// <summary>
+        /// Converts a <see cref="IScopedNode"/> to a <see cref="ITypedElement"/>.
+        /// </summary>
+        /// <param name="node">An <see cref="IScopedNode"/> node</param>
+        /// <returns>An implementation of <see cref="ITypedElement"/></returns>
+        /// <remarks>Be careful when using this method, the returned <see cref="ITypedElement"/> does not implement
+        /// the methods <see cref="ITypedElement.Location"/> and <see cref="ITypedElement.Definition"/>.    
+        /// </remarks>
+        public static ITypedElement AsTypedElement(this IScopedNode node) =>
```
ewoutkramer, line 27: I did something comparable in `SourceNodeExtensions.cs`, but there, I used an `Obsolete` attribute to make this crystal clear to the user:

```csharp
        [Obsolete("WARNING! For internal API use only. Turning an untyped SourceNode into an ITypedElement without providing" +
"type information (see other overload) will cause side-effects with components in the API that are not prepared to deal with" +
"missing type information. Please don't use this overload unless you know what you are doing.")]
        public static ITypedElement ToTypedElement(this ISourceNode node) =>
        new SourceNodeToTypedElementAdapter(node);
```

src/Hl7.Fhir.Base/ElementModel/TypedElementParseExtensions.cs
```diff
@@ -79,11 +79,11 @@ public static Model.Quantity ParseQuantity(this ITypedElement instance)
             return newQuantity;
         }
 
-        public static T ParsePrimitive<T>(this ITypedElement instance) where T : PrimitiveType, new()
+        public static T ParsePrimitive<T, U>(this U instance) where T : PrimitiveType, new() where U : IBaseElementNavigator<U>
```
ewoutkramer, line 82: I think we need to provide an overload with the old signature + this one, since I think this is a breaking change (you had to change your code above, after all!).

src/Hl7.Fhir.Base/ElementModel/ScopedNodeExtensions.cs
```diff
@@ -136,7 +136,7 @@ public static ResourceIdentity MakeAbsolute(this ScopedNode node, ResourceIdenti
             string? url = element switch
             {
                 { Value: string s } => s,
-                { InstanceType: FhirTypeConstants.REFERENCE } => element.ParseResourceReference()?.Reference,
+                { InstanceType: FhirTypeConstants.REFERENCE } => element.ParseResourceReference<ITypedElement>()?.Reference,
```
ewoutkramer, line 139: Was this necessary? It couldn't be derived from the type of `element` here? If so, then this is a breaking change to the public interface, right?

ewoutkramer (changes requested):

ewoutkramer:
Writing down some ideas....

IBaseNavigator = IElementNavigator
IScopedNode = IElementNode

IBaseNavigator = IElementNode
IScopedNode =IElementTreeNode

src/Hl7.Fhir.Base/ElementModel/TypedElementParseExtensions.cs
marcovisserFurore: Every parse extension has now an extra method for IScopedNode, so we don't break the old signature.

ewoutkramer (approved):

PR #2594, opened 2023-09-18
VONK-5300 Accessing the Definition when visiting all descendants of a ITypedElement 

louisledev:
## Description
Add an access to the Definition property when visiting all descendants of a ITypedElement in order to detect element with an invalid definition .
Background for this PR: In Vonk, for the patch operation, we create a PartToElementAdapter (see https://github.com/FirelyTeam/Vonk/blob/0b352a37b8d717b4c590622dc5d89347670e5c54/src/Vonk.Core/ElementModel/FhirPatchedElement.cs#L176) with the field to patch. In case of a non-existing field for the resource type to be patched, when getting the definition of the element, we get an exception. So, as part of the prevalidation for the patch operation, we would like to check that both the value but also the definition of all the PartToElementAdapter are valid.
see PR fixing the issue locally in VONK: https://github.com/FirelyTeam/Vonk/pull/1300

## Related issues
VONK-5300

## Testing
Tested using Firely Server. See https://github.com/FirelyTeam/Vonk/pull/1300

## FirelyTeam Checklist


marcovisserFurore:
We think this change will have a big impact on the performance for our existing and future users. We understand that this issue is a problem for the Firely Server, but maybe we can come up with a different solution with less impact. See the added issue [here](https://github.com/FirelyTeam/firely-net-sdk/issues/2597).

mmsmits (approved):

PR #2595, opened 2023-09-19
#1442: Added Quantity arithmetic

### LINKED ISSUE #1442: [FhirPath] Add support for Quantity arithmetic

marcovisserFurore:
## Description
Added the `Quantity` arithmetic operations, like Add, Substract, Multiply and Divide. These operations can now also be used in FhirPath.

I changed the signature of the not implemented static operators of `Quantity`. Technically is that a breaking change, but here we can get away with it or not?

## Related issues
Resolves #1442.

## Testing
- Hl7.Fhir.ElementModel.Tests.QuantityTests.ArithmeticOperationsTests
- HL7.FhirPath.Tests.OperationsTests.AssertTestcases

## FirelyTeam Checklist


ewoutkramer (approved):

ewoutkramer:
> I changed the signature of the not implemented static operators of Quantity. Technically is that a breaking change, but here we can get away with it or not?

Technically, yes, but we would only bother people who had used the addition operators to cause a `NotImplementedException`. The `bool` return type also was clearly a copy/paste error!

I think we can easily make this change.

Issue #1698, opened 2021-04-28
Make SnapshotResolver accept an IAsyncResourceResolver in the constructor


ewoutkramer:
**Is your feature request related to a problem? Please describe.**
Currently, the constructors for `SnapshotResolver` only accept an `IResourceResolver` as a dependency. However, we do need to use it with async resolvers as well.

**Describe the solution you'd like**
I'd like the SnapshotResolver to work async, and thus implement IAsyncResourceResolver (which I think it does, but it must be on top of a sync IResourceResolver) natively.

PR #2632, opened 2023-12-01
Upgrade SnapshotSource to support async resolvers

ewoutkramer:
Internally, SnapshotSource was already using an async resolver, but the constructor still demanded a non-async resolver. This needed to be solved to have the SnapshotSource working in concert with the validator and terminology service.


ewoutkramer:
Fixes #1698

ewoutkramer:
Closing, as I created a new PR (#2635) based on develop.

PR #2634, opened 2023-12-05
Add Process message operation

### LINKED ISSUE #2605: $process-message operation on FhirClient
We're working with [FHIR messaging](http://hl7.org/fhir/R4/messaging.html). For this we opted to use the [$process-message operation](http://hl7.org/fhir/R4/messageheader-operation-process-message.html). This line in the specifications is giving me headaches:
```
This operation does not use the parameters resource; the parameters "async" and "response-url" always go in the URL, if they are used, and the "content" parameter is always the body of the HTTP message.
```
I don't think it is possible to do this with the FhirClient. All the `operation`-based functions take a `Parameter` object and post it to the    server.
cf #2601

```
public virtual Task<Resource?> ProcessMessageAsync(Bundle bundle, bool async = false, string responseUrl = null, CancellationToken? ct = null)
```

Not using the FhirClient and do everything using the vanilla HttpClient :(

LodewijkSioen:
## Description
Adds support for the ($process-message)[http://hl7.org/fhir/messageheader-operation-process-message.html] operation used in fhir (messaging)[http://hl7.org/fhir/messaging.html#process].

## Related issues
Resolves #2605

## Testing
Added unit test.

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Base/Rest/BaseFhirClient.cs
+++ b/src/Hl7.Fhir.Base/Rest/BaseFhirClient.cs
@@ -938,6 +938,20 @@ namespace Hl7.Fhir.Rest
             return OperationAsync(operation, parameters, useGet).WaitResult();
         }
 
+        public virtual Task<Bundle?> ProcessMessageAsync(Bundle bundle, bool async = false, string? responseUrl = null, CancellationToken? ct = null)
+        {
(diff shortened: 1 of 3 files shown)


mmsmits (approved):
Thanks for the PR!

Issue #2647, opened 2024-01-02
SourceNodeComparator is missing returns.


mmsmits:
`SourceNodeComparator ` sometimes calls `TreeComparisonResult.Fail(actual.Location, $"number of children was different");`, but then just continues, instead of returning the results. 

This results in false positive outcomes.

ewoutkramer:
oops ;-)

PR #2648, opened 2024-01-02
Add missing returns to SourceNodeComparator

mmsmits:
fixes: #2647


marcovisserFurore (approved):

p-liubchenko (changes requested):
Line 20 still missing return

marcovisserFurore:
> Line 20 still missing return

Well spotted Paul!

p-liubchenko:
also please consider adding tests for each fail case so we will never get back to this again

Issue #2583, opened 2023-09-06
ModelBinding fails with Error ... Patient is abstract .... could not be instantiated and populated


Brilbroeder:
project-file
`       <PackageReference Include="Hl7.Fhir.R5" Version="5.3.0" />
`

startup.cs
```
            services
                .AddControllers()
                .AddJsonOptions(options =>
                {
                    options.JsonSerializerOptions.ForFhir(ModelInfo.ModelInspector);
                });
```

controller-action
```
        [HttpPost]
        public ActionResult Create(Patient patient)
        {
            return Ok(patient);
        }
```

Error : 
`System.NotSupportedException: The collection type 'Hl7.Fhir.Model.Patient' is abstract, an interface, or is read only, and could not be instantiated and populated. Path: $ | LineNumber: 0 | BytePositionInLine: 1.
`

Any suggestions ?

ewoutkramer:
The text "The collection type" is suspicous. It's not a collection.,,,,,wait a second....

Yes, every resource is an `IEnumerable<KeyValuePair<>>`.  

I guess this means that in the list of converters installed in the service, there's a (default) converter installed that serializes collections BEFORE our own special FHIR serializer is called.  Which is entirely possible.  `ForFhir()` just adds our converter at the end of the list.

That's probably it. Not yet a solution, but we could start to experiment by wiping the `options.Converters` array in the lambda before the call to `ForFhir()`?

ewoutkramer:
I do regret letting Base (and subclasses) implement IReadOnlyDictionary<> directly. We've seen a few frameworks thrown off by this, much like the serializer here.  Also, for example, the Visual Studio debugger shows resources as key value pairs in its enumerable viewer.

Could we explore the idea of having an indirect implementation, i.e.  having an `AsReadOnlyDictionary()` on Base, instead of it implementing it directly?  That's obviously a breaking change, but would it be worth it?

Brilbroeder:
@ewoutkramer I tried the following. (see below) This should put your converter first. I put a breakpoint to see the number of savedConverters. It is 0. Hence this did not change anything.

```
                .AddJsonOptions(options =>
                {
                    var savedConverters = options.JsonSerializerOptions.Converters.ToList();
                    options.JsonSerializerOptions.Converters.Clear();
                    options.JsonSerializerOptions.ForFhir(ModelInfo.ModelInspector);

                    foreach (var savedConverter in savedConverters)
                    {
[...]

PR #2651, opened 2024-01-03
Modify JsonSerializationOptions instead of creating a new instance

mmsmits:
fixes: #2583

### DIFF
--- a/src/Hl7.Fhir.Support.Poco.Tests/NewPocoSerializers/FhirJsonSerializationTests.cs
+++ b/src/Hl7.Fhir.Support.Poco.Tests/NewPocoSerializers/FhirJsonSerializationTests.cs
@@ -41,7 +41,8 @@ namespace Hl7.Fhir.Support.Poco.Tests
         {
             var (poco, expected) = getEdgecases();
 
-            var options = BaseOptions.Pretty();
+            var options = new JsonSerializerOptions().ForFhir(typeof(TestPatient).Assembly).Pretty();
+
             string actual = JsonSerializer.Serialize(poco, options);
 
             var errors = new List<string>();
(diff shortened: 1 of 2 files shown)


ewoutkramer (changes requested):
There is a `ForFhir<F>` just below that has the same problem.

More in general, we create & return new options, which does not work well with the ASP.NET service configuration pattern, e.g. see here: https://www.meziantou.net/configuring-json-options-in-asp-net-core.htm

`Pretty()` and `Compact()` will have the same problems.  Thinking about this a bit more, changing this could be considered a breaking change, since we're suddenly changing the input parameter, which is probably unexpected.

What to do about that?

ewoutkramer (dismissed):

src/Hl7.Fhir.Support.Poco.Tests/NewPocoSerializers/FhirJsonSerializationTests.cs
```diff
@@ -41,7 +41,8 @@ public void RoundtripEdgeCases()
         {
             var (poco, expected) = getEdgecases();
 
-            var options = BaseOptions.Pretty();
+            var options = new JsonSerializerOptions().ForFhir(typeof(TestPatient).Assembly).Pretty();
```
ewoutkramer, line 44: This is exactly why this could be considered a breaking change, as we discussed on our call. Still, the benefits for the use in service initialization outweigh the risk I think. And I guess many people even assume we return the same instance (but then modified). But we cannot be sure.

ewoutkramer (approved):

src/Hl7.Fhir.Support.Poco.Tests/NewPocoSerializers/FhirJsonSerializationTests.cs
mmsmits: I'll add something to the release notes

Issue #2674, opened 2024-01-18
Upgrade to .NET8.0


mmsmits:
Upgrade to .NET8.0, ditch .NET6.0, keep netstandard2.0 for backwards compatibility

baseTwo:
This may require some work on the build pipelines as well if we upgrade to latest C# version while upgrading to .net8

PR #2677, opened 2024-01-22
Update solution to .NET 8

mmsmits:
Dropping .NET 6 keeping netstandard2.0, for backwards compatibility

### DIFF
--- a/build/build-variables.yml
+++ b/build/build-variables.yml
@@ -3,7 +3,7 @@
 # Variables used during builds.
 
 variables:
-  NET_CORE_SDK: 7.0.x
+  NET_CORE_SDK: 8.0.x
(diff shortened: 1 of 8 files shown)


ewoutkramer (approved):

mmsmits:
forgot to update the framework for running the unit tests. Fixed this so this doesn't happen again next time.

mmsmits:
As of .NET 8 Microsoft not only not recommends  the use of BinaryFormatter, they are throwing an error if you have not explicitly enabled it.
I think we can safely do that, since we only use this in the test classes, but please review.
More info: https://learn.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-security-guide

ewoutkramer (approved):

mmsmits:
fixes #2674

PR #2680, opened 2024-01-23
2610 assert unique json ids

### LINKED ISSUE #2610: Behaviour of json validation wrt duplicate keys
How to deal with json resources with duplicate keys?

``` json
{
    "a" : "x",
    "a" : "y"
}
```

The Json ECMA spec is silent about it, the RFC "recommends" unique keys, and the FHIR spec, luckily, is explicit: these are not allowed.

The fact that the JSON spec itself is not explicit is probably the reason that Json.net will not complain, the default behaviour is to let subsequent occurrences of a key overwrite the previous ones:

https://www.newtonsoft.com/json/help/html/P_Newtonsoft_Json_Linq_JsonLoadSettings_DuplicatePropertyNameHandling.htm

* We should hardwire this setting to "Error" for Json.net
* In the new parsers, we are using a forward-only json reader, so the task of detecting this rests on our own shoulders.

Kasdejong:
## Description
Added exceptions for duplicate json IDs in an object fed to the JSON deserializer

## Related issues
Resolves #2610.

## Testing
Unit test added for duplicate IDs

### DIFF
--- a/src/Hl7.Fhir.Base/Serialization/BaseFhirJsonPocoDeserializer.cs
+++ b/src/Hl7.Fhir.Base/Serialization/BaseFhirJsonPocoDeserializer.cs
@@ -251,12 +251,10 @@ namespace Hl7.Fhir.Serialization
             var oldErrorCount = state.Errors.Count;
             var (line, pos) = reader.GetLocation();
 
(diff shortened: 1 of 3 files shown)


ewoutkramer:
@Kasdejong started out with a real PR (#2708), so I am closing this one.

PR #2708, opened 2024-02-13
#2610 signal duplicate properties while parsing

Kasdejong:
## Description
JSON deserializer should now correctly identify and report duplicate property keys

## Related issues
Resolves issue #2610.

## Testing
Unit tests added to assert (fatal) duplicate property keys errors and to assert their differentiation from (recoverable) duplicate array key errors.


src/Hl7.Fhir.Base/Serialization/BaseFhirJsonPocoDeserializer.cs
```diff
@@ -367,11 +367,28 @@ FhirJsonPocoDeserializerState state
...
                     state.Errors.Add(ERR.USE_OF_UNDERSCORE_ILLEGAL(ref reader, state.Path.GetInstancePath(), propertyMapping.Name, propertyName));
-
+                
                 // Note that repeating simple elements (like Extension.url) do not currently exist in the FHIR serialization
-                result = propertyMapping.IsCollection
-                    ? deserializeNormalList((IList)existingValue!, propertyValueMapping, ref reader, propertyMapping, state)
-                    : deserializeSingleValue(ref reader, propertyValueMapping, propertyMapping, state);
+                if (propertyMapping.IsCollection)
+                {
+                    IList l = (existingValue as IList)!;
```
ewoutkramer, line 374:
```suggestion
                    var l = (IList)existingValue;
```

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonPocoDeserializer.cs
```diff
@@ -367,11 +367,28 @@ FhirJsonPocoDeserializerState state
...
+                    if (l.Count > 0)
+                    {
+                        state.Path.IncrementIndex(l.Count);
+                        state.Errors.Add(ERR.DUPLICATE_ARRAY(ref reader, state.Path.GetInstancePath()));
+                    }
+                    result = deserializeNormalList(l, propertyValueMapping, ref reader, propertyMapping, state);
+                } 
+                else 
+                {
+                    // if the property already has a mapping, its key must have been encountered before
```
ewoutkramer, line 385:
```suggestion
                    // if the property already has a value, its key must have been encountered before
```

ewoutkramer:
We broke it with our last change! It did work before. So maybe the cast can result in null after all?

ewoutkramer:
> Looks like this should be non-null asserted after all

No, that does not make sense, as the `!` is just a compiler hint for static analysis and has no effect on the generated code.

Kasdejong:
With the compiler hint, the code works though... Could it be a runtime type error (nullable/non-nullable)?

ewoutkramer (approved):

PR #2711, opened 2024-02-14
Refactored the round-trip tests so they don't use intermediate files anymore.

### LINKED ISSUE #2707: Improve roundtrip tests performance
Question from @GinoCanessa 

> A lot of the 'long' tests are reading and writing contents to disk for all intermediate stages - are those artifacts necessary somewhere, or could we remove them?  
> 
> For example, skipping the intermediate file IO on my machine takes the execution of `FullRoundtripOfAllExamplesJsonNewSerializer` from 4.1 minutes down to 25.3 seconds.

It would be very nice to make this faster, that would improve the duration of our build pipeline greatly.

ewoutkramer:
* Combined all parsers into one test class
* Removed tests for async stuff, since that only tests Newtonsoft/Microsoft XML library (we don't have any async parsers)
* Moved specific attachment tests out to another class.

Fixes #2707.


mmsmits (approved):
nice!

Issue #2264, opened 2022-10-10
Add terminology support for some non-supplied codesystems


ewoutkramer:
There are a few popular codesystems that are not supplied as a FHIR codesystem resource, but for which basic support should be built into our terminology services. E.g. by supplying an `ITerminologyService` implementation that is specifically written to handle these codesystems.

The codesystems that we should support are:
* Currencies
* Country codes
* Languages
* Mime types

For some of these, we can write some basic validations (e.g. the allowed first parts of mime types, syntax of language codes). In fact, since these lists are (somewhat) volatile, we probably should not try to be complete.

brianpos:
I have some of these already...

mmsmits:
We have a MimeTypeTerminologyService, we could do the rest too now easily.

PR #2715, opened 2024-02-15
#2264 add language terminology support

Kasdejong:
## Description
Added base class for Code/Coding-based Terminology services. Added support for language terminology (Regular expression matching)

## Related issues
Resolves issue #2264.

## Testing
Added new unit tests to test both the new implementation and regex (found in LanguageTerminologyServiceTests.cs)


src/Hl7.Fhir.Base/Specification/Terminology/CodeSystemTerminologyService.cs
```diff
@@ -0,0 +1,175 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using T = System.Threading.Tasks;
+
+namespace Hl7.Fhir.Specification.Terminology
+{
+    /// <summary>
+    /// Base class for checking Code terminology
+    /// </summary>
+    abstract public class CodeSystemTerminologyService : ITerminologyService
```
mmsmits, line 18: This is really minor, but this is usually `public abstract` instead of the other way around.

src/Hl7.Fhir.Base/Specification/Terminology/LanguageTerminologyService.cs
```diff
@@ -0,0 +1,28 @@
+#nullable enable
+
+using System.Text.RegularExpressions;
+
+namespace Hl7.Fhir.Specification.Terminology
+{
+    /// <summary>
+    /// Checks if codes are valid currency types
```
mmsmits, line 8: Not currencies but languages

src/Hl7.Fhir.Base/Specification/Terminology/MimeTypeTerminologyService.cs
```diff
@@ -1,154 +1,24 @@
...
-using System.Threading.Tasks;
-using T = System.Threading.Tasks;
 
 namespace Hl7.Fhir.Specification.Terminology
 {
     /// <summary>
     /// Checks if codes are valid Mime-Types (urn:ietf:bcp:13)
     /// </summary>
-    public class MimeTypeTerminologyService : ITerminologyService
+    public sealed class MimeTypeTerminologyService : CodeSystemTerminologyService
```
mmsmits, line 10: This is a breaking change

src/Hl7.Fhir.Base/Specification/Terminology/CodeSystemTerminologyService.cs
```diff
@@ -0,0 +1,175 @@
...
+    {
+        private readonly string _terminologyType;
+        private readonly string _codeSystem;
+        private readonly string[] _codeValueSets;
+
+        protected CodeSystemTerminologyService(string terminologyType, string codeSystem, string[] codeValueSets)
```
mmsmits, line 24: This should be documented, what the parameters mean especially.

src/Hl7.Fhir.Base/Specification/Terminology/CodeSystemTerminologyService.cs
mmsmits: And does this class have to be public? Isn't internal good enough? Or do we want our users also to use this?

mmsmits (changes requested):

src/Hl7.Fhir.Base/Specification/Terminology/CodeSystemTerminologyService.cs
Kasdejong: It has to be public or the code will not compile, derived classes cannot be more accessible than their base class

mmsmits (approved):
Good job!

PR #2720, opened 2024-02-19
FHIRPath optimizations

cknaap:
## Description
Optimizations to FHIRPath, child navigation and chid caching

## Related issues
none

## Testing
Regression test on the existing unit tests

## FirelyTeam Checklist
- [x] Mark the PR with the label **breaking change** when this PR introduces breaking changes (not applicable)

### DIFF
--- a/src/Hl7.Fhir.Base/ElementModel/PocoElementNode.cs
+++ b/src/Hl7.Fhir.Base/ElementModel/PocoElementNode.cs
@@ -80,14 +80,15 @@ namespace Hl7.Fhir.ElementModel
 
         public string ShortPath { get; private set; }
 
-        private static (ElementValue val, int ix)[] returnElements(string name, object element)
+        private (ElementValue val, int ix)[] returnElements(string name, object element)
         {
-            return element switch
+            return (element, Current, name) switch
             {
-                Base @base => new[] { (new ElementValue(name, @base), 0) },
-                IEnumerable<Base> bases => bases.Select((e, i) => (new ElementValue(name, e), i)).ToArray(),
+                (Base @base, _, _)  => new[] { (new ElementValue(name, @base), 0) },
+                (IEnumerable<Base> bases, _, _)  => bases.Select((e, i) => (new ElementValue(name, e), i)).ToArray(),
+                (string s, Extension, "url")  => new[]{(new ElementValue(name, new FhirUri(s)), 0)},
+                (string s, Element, "id")  => new[]{(new ElementValue(name, new FhirString(s)), 0)},
                 _ => Array.Empty<(ElementValue, int)>(),
-                // _ => throw new ArgumentException($"element '{location}.{name}' is of type '{element.GetType().FullName}' not a Base or IEnumerable<Base>", nameof(element))
             };
         }
 
--- a/src/Hl7.Fhir.Base/ElementModel/PocoElementNode.cs
+++ b/src/Hl7.Fhir.Base/ElementModel/PocoElementNode.cs
@@ -1,7 +1,7 @@
-/* 
+/*
  * Copyright (c) 2016, Firely (info@fire.ly) and contributors
  * See the file CONTRIBUTORS for details.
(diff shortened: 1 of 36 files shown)


src/Hl7.Fhir.Base/FhirPath/Expressions/SymbolTableInit.cs
```diff
@@ -223,6 +224,23 @@ public static SymbolTable AddStandardFP(this SymbolTable t)
         }
 
 
+        internal static void AddBuiltinChildren(this SymbolTable table)
```
ewoutkramer, line 227: What does this do?

src/Hl7.Fhir.Base/FhirPath/Functions/UtilityOperators.cs
```diff
@@ -10,6 +10,10 @@ internal static class UtilityOperators
     {
         public static IEnumerable<ITypedElement> Extension(this IEnumerable<ITypedElement> focus, string url)
         {
+            var exts = focus.Navigate("extension").ToArray();
```
ewoutkramer, line 13: This seems to double the work?

ewoutkramer: I think you meant to return `matches` here, splitting it up was maybe for debug purposes?

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/ElementModel/PocoElementNode.cs
```diff
@@ -67,50 +71,89 @@ private Type determineInstanceType(PropertyMapping definition)
...
+                (string s, Extension, "url")  => new[]{(new ElementValue(name, new FhirUri(s)), 0)},
+                (string s, Element, "id")  => new[]{(new ElementValue(name, new FhirString(s)), 0)},
+                _ => Array.Empty<(ElementValue, int)>(),
+            };
+        }
 
-            var children = parentBase.NamedChildren;
-            string oldElementName = null;
-            int arrayIndex = 0;
+        public IEnumerable<ITypedElement> Children(string name)
```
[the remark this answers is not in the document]
cknaap: The crucial difference is that previously:
- it got *all* children with `NamedChildren`
- then visited that whole list until all the items with the right name were seen
- and maintained the array index per child name along the way
And currently:
- it utilizes IReadOnlyDictionary to quickly get *only* the children with the correct name
- appends the array index

I see your point that ElementValue may not be necessary anymore, since `ElementValue.name` is not even used. I will check that.

src/Hl7.Fhir.Base/FhirPath/Expressions/SymbolTableInit.cs
cknaap: It avoids the `Wrap` for the `builtin.children` function (which is added by the `Add` method used before), since input and output are known to be `IEnumerable<ITypedElement>`.

src/Hl7.Fhir.Base/FhirPath/Functions/UtilityOperators.cs
cknaap: Indeed. I removed it again.

src/Hl7.Fhir.Base/ElementModel/PocoElementNode.cs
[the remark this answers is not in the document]
cknaap: ElementValue was indeed not needed anymore, I removed it from this method.
Note that NamedChildren and ElementValue have no use at all anymore now in the SDK itself. But they are public...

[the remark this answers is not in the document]
ewoutkramer: Yes, and I think e.g. Forge is using it, since it was added by Michel initially!

ewoutkramer (approved):

Issue #2550, opened 2023-07-27
Make it easier to ignore specific errors when using the new parsers


ewoutkramer:
When parsing finishes with issues, a `DeserializationFailedException` is thrown. This will signal to most people that deserialization has failed. However, this is not actually the intent: some of the issues found may be acceptable to the caller. Also, it is unclear that parsing continues until all issues are found, and that a poco is returned in the `PartialResult` property of the exception. Even when users would realize this, you'd still have to filter the issues for acceptable issues or real errors.

To make this easier to do, I propose to extend the current configuration interface of the System.Text.Json parsers like so:

```
var serializerOptions = new JsonSerializerOptions()
     .ForFhir()
     .Ignoring("JSON004", "JSON005");
```

In addition we can re-use the "standard set" of ignorable errors from `FhirSerializationEngineFactory`:

```
var serializerOptions = new JsonSerializerOptions()
     .ForFhir()
     .WithLeniency(BackwardsCompatible/Ostrich/etc);
```

Beneath this small layer of System.Text.Json, we could also adapt the new JsonPocoDeserializer() class with comparable methods:

* `var resource = Deserialize(jsonText, out var issues);`
* `var deserializer = new JsonPocoDeserializer().WithLeniency(....)`.
* `var deserializer = new JsonPocoDeserializer().Ignoring(....)`.

I like the `WithLeniency`, although we have a slightly different syntax for configuring the `FhirClient`.  Maybe we should add `WithLeniency` there, and adopt the other concrete methods (`WithBackwardsCompatibleSerializer`). In any case, keep this aligned.

PR #2730, opened 2024-03-06
2550 parser error ignore customization

Kasdejong:
## Description
- Added support for fluent configuration (ignore error/enforce errors/select error ignore mask from preset deserializer mode) of the System.Text.Json deserializer by adding extension methods to the Json options. 
- Changed the internal behaviour of the json converter to maintain its own serialization engine as opposed to directly maintaining a low-level deserializer. 
- Added an interface for distinction between element model serialization engines and poco serialization engines.

## Related issues
Closes [#2550].

## Testing
Added tests to assert correct behaviour of the error-filtering predicates and their prevalence according to where in the chain they are called


src/Hl7.Fhir.Base/Serialization/engine/PocoSerializationEngine_Json.cs
```diff
@@ -0,0 +1,68 @@
...
+
+    private BaseFhirJsonPocoSerializer getJsonSerializer() =>
+        _jsonSerializer ??= new BaseFhirJsonPocoSerializer(_inspector.FhirRelease, _jsonSerializerSettings!);
+    
+    /// <inheritdoc />
+    public Resource DeserializeFromJson(string data)
+    {
+        return (Resource)deserializeAndFilterErrors(() =>
+        { 
+            _ = getJsonDeserializer().TryDeserializeResource(data, out var instance, out var issues);
```
mmsmits, line 31: What happens when this fails, we still return the resource, which could be null?
No, it can't be, because Rider/VS would throw an error.

src/Hl7.Fhir.Base/Serialization/engine/PocoSerializationEngine_Json.cs
```diff
@@ -0,0 +1,68 @@
...
+    public Resource DeserializeFromJson(ref Utf8JsonReader reader) => deserializeAndFilterErrors(getJsonDeserializer(), ref reader);
+    
+    /// <inheritdoc />
+    public Base DeserializeObjectFromJson(Type targetType, ref Utf8JsonReader reader) => 
+        deserializeObjectAndFilterErrors(targetType, getJsonDeserializer(), ref reader);
+    
+    // overload necessary since ref structs cannot be captured in the lambda
+    private Resource deserializeAndFilterErrors(BaseFhirJsonPocoDeserializer deserializer, ref Utf8JsonReader reader)
+    {
+        _ = deserializer.TryDeserializeResource(ref reader, out var instance, out var issues);
```
mmsmits, line 46: Same thing here. Using a try, but no matter the outcome do the same thing triggers me. Can you explain?

mmsmits (dismissed):

src/Hl7.Fhir.Base/Serialization/JsonSerializerOptionsExtensions.cs
```diff
@@ -96,6 +98,54 @@ public static JsonSerializerOptions ForFhir(this JsonSerializerOptions options,
...
+        /// <remarks>
+        /// Modifying the options is always left-associative. This means that defining custom constraints should probably be done AFTER setting the mode.
+        /// </remarks>
+        public static JsonSerializerOptions UsingMode(this JsonSerializerOptions options, DeserializerModes mode)
+        {
+            var factory = getCustomFactoryFromList(options.Converters);
+            
+            factory.IgnoreFilter = mode switch
+            {
+                DeserializerModes.Recoverable => FilterPredicateExtensions.IsRecoverableIssue,
```
ewoutkramer, line 114: This is somewhat of a duplicate of the factory methods in the `FhirSerializationEngineFactory`. In fact, there is a risk that the mode switch is different from creating a specific engine. In fact, setting the mode to Ostrich would only take into account the list of errors, while it is also important that the validators are set to null (for performance reasons). I know you deleted that code, but it is relevant.  So in future updates we run the risk of updating the factory methods and not this or vice versa. Can we find a smart solution to this?

ewoutkramer: Maybe, when setting deserializer modes, we shouldn't swap filters, but we should swap engines -> create the engine with the factory and then set it in the CustomFactory?

src/Hl7.Fhir.Base/Serialization/engine/FhirSerializationEngineFactory.cs
```diff
@@ -111,50 +114,67 @@ private static FhirJsonParsingSettings buildJsonParsingSettings(Mode mode) => ne
 
 
         /// <summary>
-        /// Create an implementation of <see cref="IFhirSerializationEngine"/> configured to flag all parsing errors, 
+        /// Create an implementation of <see cref="IFhirExtendedSerializationEngine"/> configured to flag all parsing errors,
```
ewoutkramer, line 117: I don't think I want to make this interface public. So maybe it is better for these to return IFhirSerializationEngine, and then in the FhirJsonConverter, we can assume/enforce that the engine implements `IFhirExtendedSerializationEngine` or that it is actually a `PocoSerializationEngine`.

src/Hl7.Fhir.Base/Serialization/engine/PocoSerializationEngine_Json.cs
```diff
@@ -0,0 +1,68 @@
...
+internal partial class PocoSerializationEngine
+{
+    private readonly FhirJsonPocoDeserializerSettings? _jsonDeserializerSettings;
+    private readonly FhirJsonPocoSerializerSettings? _jsonSerializerSettings;
+    
+    private BaseFhirJsonPocoDeserializer? _jsonDeserializer;
+    private BaseFhirJsonPocoSerializer? _jsonSerializer;
+    
+    private BaseFhirJsonPocoDeserializer getJsonDeserializer() => 
+        _jsonDeserializer ??= new BaseFhirJsonPocoDeserializer(_inspector, _jsonDeserializerSettings!);
```
ewoutkramer, line 21: Ha! A Null coalescing assignment combined with the fact that assignment can be used as an expression, so that is can be returned. You're sure you did not use C# before?

src/Hl7.Fhir.Base/Serialization/engine/PocoSerializationEngine.cs
```diff
@@ -10,67 +10,58 @@
...
+        private Base deserializeAndFilterErrors(TryDeserializer deserializer)
         {
-            var deserializer = new BaseFhirJsonPocoDeserializer(_inspector, _jsonSettings);
-            return deserializeAndIgnoreErrors(deserializer.TryDeserializeResource, data);
-        }
-
-        private Resource deserializeAndIgnoreErrors(TryDeserializer deserializer, string data)
-        {
-            bool success = deserializer(data, out var instance, out var issues);
+            (Base? instance, IEnumerable<CodedException> issues) = deserializer();
```
ewoutkramer, line 60: You don't have to mention the types (but you can if you think it's clearer), `var (instance, issues) = deserializer()` would work too.

src/Hl7.Fhir.Support.Poco.Tests/NewPocoSerializers/FhirJsonDeserializationTests.cs
```diff
@@ -1049,6 +1050,144 @@ public void TestBackboneElementEmptyStack()
...
+            string testJson = File.ReadAllText(Path.Combine("TestData", "fp-test-patient-errors.json"));
+            
+            try
+            {
+                _ = JsonSerializer.Deserialize<TestPatient>(testJson, options);
+                throw new DeserializationFailedException(null, []);
+            }
+            catch (DeserializationFailedException dfe)
+            {
+                Assert.IsTrue(shouldHold(dfe.Exceptions));
```
ewoutkramer, line 1106: We should be using FluentAssertions for all new unit tests, so  `shouldHold(dfe.Exceptions)`.Should().BeTrue();`

ewoutkramer (changes requested):
No fundamental issues, just some remarks ;-)

src/Hl7.Fhir.Base/Serialization/engine/PocoSerializationEngine_Json.cs
Kasdejong: I more or less copied this from the previous implementation, deserializeAndFilterErrors either returns an instance or throws a DFE (where the instance is nullable).

We should check if the ostrich parser breaks this behaviour with completely malformed json by returning null and not reporting any issues

src/Hl7.Fhir.Base/Serialization/engine/PocoSerializationEngine_Json.cs
Kasdejong: Copied from previous implementation, we could change this of course

src/Hl7.Fhir.Base/Serialization/engine/FhirSerializationEngineFactory.cs
```diff
@@ -111,50 +114,67 @@ private static FhirJsonParsingSettings buildJsonParsingSettings(Mode mode) => ne
...
 
         /// <summary>
-        /// Create an implementation of <see cref="IFhirSerializationEngine"/> configured to allow errors
+        /// Create an implementation of <see cref="IFhirExtendedSerializationEngine"/> configured to allow errors
         /// and just continue parsing. Note that this may mean data loss.
         /// </summary>
-        public static IFhirSerializationEngine Ostrich(ModelInspector inspector) =>
-            new PocoSerializationEngine(inspector, _ => true, new FhirJsonPocoDeserializerSettings(){Validator = null}, new FhirXmlPocoDeserializerSettings(){Validator = null});
+        public static IFhirExtendedSerializationEngine Ostrich(ModelInspector inspector) =>
+            new PocoSerializationEngine(inspector, _ => true);
```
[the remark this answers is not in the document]
Kasdejong: This was a mistake on my part. The branch which implemented that behaviour was not merged yet when I started this one, and I probably looked over this on an intermediate merge conflict resolution

src/Hl7.Fhir.Base/Serialization/JsonSerializerOptionsExtensions.cs
Kasdejong: We want the predicates to be called on the (known) base predicate. I had the implementation like this at first, passing on the engine mode, but this did not work, since the intended behaviour for .Enforcing() has to be able to bind to something (some base mode) to keep track of the .ignoring and .enforcing combined predicate

src/Hl7.Fhir.Base/Serialization/FhirJsonConverter.cs
```diff
@@ -15,97 +15,95 @@
...
         }
 
-        public FhirJsonConverterFactory(
-            ModelInspector inspector,
-            FhirJsonPocoSerializerSettings? serializerSettings = null,
-            FhirJsonPocoDeserializerSettings? deserializerSettings = null,
-            Predicate<CodedException>? ignoreFilter = null)
+        internal void SetEnforcedErrors(IEnumerable<string> toEnforce)
+        {
+            Engine ??= (FhirSerializationEngineFactory.Strict(inspector, serializerSettings, deserializerSettings) as PocoSerializationEngine)!;
```
ewoutkramer: `(x as Y)!` suggests you really want `((Y)x)` here, which will never return null (and throw if the cast does not work).

src/Hl7.Fhir.Base/Serialization/FhirJsonConverter.cs
```diff
@@ -15,97 +15,95 @@
...
+            Engine.IgnoreFilter = Engine.IgnoreFilter.And(toEnforce.ToPredicate().Negate());            
+        }
+
+        internal void SetIgnoredErrors(IEnumerable<string> toIgnore)
+        {
+            Engine ??= (FhirSerializationEngineFactory.Strict(inspector, serializerSettings, deserializerSettings) as PocoSerializationEngine)!;
```
ewoutkramer: Maybe this factory code should be in a small private function, instead of repeating it.

src/Hl7.Fhir.Base/Serialization/engine/FhirSerializationEngineFactory.cs
```diff
@@ -41,23 +42,20 @@ private static PocoBuilderSettings buildPocoBuilderSettings(Mode mode) => new()
...
+                    : null
             };
 
 
             private static FhirXmlParsingSettings buildXmlParsingSettings(Mode mode) => new()
             {
-                DisallowSchemaLocation = mode is Mode.Strict,
-                PermissiveParsing = mode is Mode.Permissive or Mode.Ostrich,
-                ValidateFhirXhtml = mode is Mode.Strict
+                DisallowSchemaLocation = mode is Mode.Strict, PermissiveParsing = mode is Mode.Permissive or Mode.Ostrich, ValidateFhirXhtml = mode is Mode.Strict
```
ewoutkramer: Weird reformat?

src/Hl7.Fhir.Base/Serialization/engine/IFhirExtendedSerializationEngine.cs
```diff
@@ -3,24 +3,26 @@
 using System.Text.Json;
 using System.Xml;
 
+#nullable enable
+
 namespace Hl7.Fhir.Serialization;
 
-public interface IFhirExtendedSerializationEngine : IFhirSerializationEngine
+internal interface IFhirExtendedSerializationEngine : IFhirSerializationEngine
```
ewoutkramer: I don't think you need this interface anymore?

ewoutkramer (changes requested):

ewoutkramer (approved):

Issue #2653, opened 2024-01-06
Nullability Checks On Try ..like method ie: TryGetIdentifier


dibiancoj:
First off want to thank the team for all the hard work with this code. It's been great for us. 

One new feature / change which would be nice is for c# nullability checks with your try-get methods.

In - https://github.com/FirelyTeam/firely-net-sdk/blob/develop/src/Hl7.Fhir.Base/Model/IIdentifiable.cs#L57C28-L57C44 (as an example)

Even if the method returns true - the c# compile will think it can be null so we have to add a guard regardless or ignore the compiler warning.

The solution which will correct these warnings would be to add the **[NotNullWhen(true)]**
`public static bool TryGetIdentifier(this IIdentifiable<List<Identifier>> identifiable, string system, [NotNullWhen(true) out Identifier? identifier)`

Which basically says if the method returns true, the identifier out parameter will not be null. Then the calling code won't need to add an additional guard.

If your open to it - I can create a PR but wanted to add the request first. 

Thank you again for this great nuget package. It's a really robust codebase.

ewoutkramer:
I have been thinking about doing this for a while, but since we also need to get this to compile on netstandard 2.0, we need to think of a trick to make this compile under both without ifdefs.  Maybe add our own definition of the attribute under netstandard that simply does nothing?

dibiancoj:
Targeting 2.0 makes sense since you need to support .net framework. The nullability attributes sit in net standard 2.1 but that kills support for .net framework.

Those attributes are all sealed so I think the only way to truly support this is multi targeting which gets messy quickly. I can't find anything to switch the attribute types so the compiler looks for your attributes.

I think you can close this and if you ever decide to multi target bring it back in. This was helpful to understand the complexities with trying to support .net core and .net framework.

mmsmits:
@dibiancoj Thanks for getting in touch, and offering to create a PR. I'll close this now, and will reopen when we move away from 2.0 .

dibiancoj:
I did find an example where you can polyfill the nullability stuff in .net standard 2.0. Towards the bottom you can add System.Diagnostics.CodeAnalysis...class essentially. I tried it out and and works well.

Not sure if the effort is worth it though.

https://www.meziantou.net/how-to-use-nullable-reference-types-in-dotnet-standard-2-0-and-dotnet-.htm

ewoutkramer:
This guy has so many useful things on his blog, I should read it more often ;-)

[...]

PR #2741, opened 2024-03-20
Nullability and Try-style function annotations

Kasdejong:
## Description
Added the [NotNullWhen(true)] annotation along with nullability to all non-generated files containing Try-style methods (returning a bool and an out parameter for the result)

## Related issues
#2653 

## Testing
Our own tests


src/Hl7.Fhir.Base/ElementModel/DomNode.cs
```diff
@@ -16,39 +18,39 @@ namespace Hl7.Fhir.ElementModel
...
         public DomNodeList<T> this[string name] => new DomNodeList<T>(ChildrenInternal(name));
 
-        public T this[int index] => ChildList[index];
+        public T? this[int index] => ChildList?[index];
 
         #region << Annotations >>
-        private AnnotationList _annotations = null;
-        protected AnnotationList AnnotationsInternal => LazyInitializer.EnsureInitialized(ref _annotations, () => new());
+        private AnnotationList? _annotations = null;
+        protected AnnotationList? AnnotationsInternal => LazyInitializer.EnsureInitialized(ref _annotations, () => new());
```
ewoutkramer: This can never be null, right? Since, when you access it, it will return a `new()` instance?

src/Hl7.Fhir.Base/ElementModel/DomNode.cs
```diff
@@ -16,39 +18,39 @@ namespace Hl7.Fhir.ElementModel
...
 
         protected bool HasAnnotations => _annotations is not null && !_annotations.IsEmpty;
 
         public void AddAnnotation(object annotation)
         {
-            AnnotationsInternal.AddAnnotation(annotation);
+            AnnotationsInternal!.AddAnnotation(annotation);
```
ewoutkramer: And then you don't need these anymore.

src/Hl7.Fhir.Base/ElementModel/DomNode.cs
```diff
@@ -16,39 +18,39 @@ namespace Hl7.Fhir.ElementModel
...
+        internal IEnumerable<T> ChildrenInternal(string? name = null) =>
             name == null ? ChildList : ChildList.Where(c => c.Name.MatchesPrefix(name));
 
-        public T Parent { get; protected set; }
+        public T? Parent { get; protected set; }
 
         public DomNodeList<T> this[string name] => new DomNodeList<T>(ChildrenInternal(name));
 
-        public T this[int index] => ChildList[index];
+        public T? this[int index] => ChildList?[index];
```
ewoutkramer: ChildList is never null -- and `T` (the element type) is not null, so we can never return null here.

src/Hl7.Fhir.Base/ElementModel/ElementNode.cs
```diff
@@ -104,13 +106,13 @@ values switch
...
-        public static IEnumerable<ITypedElement> CreateList(IEnumerable<object> values) => values switch
+        public static IEnumerable<ITypedElement?> CreateList(IEnumerable<object> values) => values switch
         {
             null => EmptyList,
-            _ => values.Select(toTT).ToList()
+            _ => values.Select(toTT).ToList()!
         };
 
-        private static ITypedElement toTT(object value) => value switch
+        private static ITypedElement? toTT(object value) => value switch
```
ewoutkramer: The parameter should be object? right? We're checking it for null in the next line.

src/Hl7.Fhir.Base/ElementModel/PrimitiveElement.cs
```diff
@@ -39,7 +39,7 @@ private PrimitiveElement(object value, string instanceType, string name)
             Name = name ?? throw new ArgumentNullException(nameof(name));
         }
 
-        public PrimitiveElement(object value, string? name = null, bool useFullTypeName = false)
+        public PrimitiveElement(object? value, string? name = null, bool useFullTypeName = false)
```
ewoutkramer: Value should not be null, so remove the `?` here - the exception on the next line makes that clear!

src/Hl7.Fhir.Base/ElementModel/ElementNode.cs
```diff
@@ -27,7 +29,7 @@ public class ElementNode : DomNode<ElementNode>, ITypedElement, IAnnotated, IAnn
         // HACK: For now, allow a Quantity (which is NOT a primitive) in the .Value property
         // of ITypedElement. This is a temporary situation to make a quick & dirty upgrade of
         // FP to Normative (with Quantity support) possible.
-        public static ITypedElement ForPrimitive(object value)
+        public static ITypedElement ForPrimitive(object? value)
```
ewoutkramer: Since `new PrimitiveElement()` does not take a null as parameter, we should not allow it here either.

ewoutkramer (changes requested):
There are too many things I don't agree with - we'll discuss per namespace, or per file or .....I don't know ;-)

src/Hl7.Fhir.Base/Introspection/PropertyMapping.cs
```diff
@@ -23,7 +24,7 @@ namespace Hl7.Fhir.Introspection
     /// <summary>
     /// A container for the metadata of an element of a FHIR datatype as present on a property of a (generated) .NET POCO class.
     /// </summary>
-    [System.Diagnostics.DebuggerDisplay(@"\{Name={Name} ElementType={ElementType.Name}}")]
+    [System.Diagnostics.DebuggerDisplay(@"\{Name={Name} ElementType={ImplementingType.Name}}")]
```
ewoutkramer, line 27: That's an unexpected change!

src/Hl7.Fhir.Base/Rest/CompressedContent.cs
```diff
@@ -40,6 +40,8 @@ protected override bool TryComputeLength(out long length)
             length = -1;
             return false;
         }
+
```
ewoutkramer: What? The whitespace?

src/Hl7.Fhir.Base/Utility/AnnotationList.cs
```diff
@@ -6,30 +6,42 @@
...
     /// This list is thread safe
     /// </summary>
-    public class AnnotationList : IAnnotatable, IAnnotated
+    [CollectionBuilder(typeof(AnnotationList), nameof(AnnotationList.create) )]
+    public class AnnotationList : IAnnotatable, IAnnotated, IEnumerable<object>
     {
         private Lazy<ConcurrentDictionary<Type, List<object>>> _annotations = new Lazy<ConcurrentDictionary<Type, List<object>>>(() => new ConcurrentDictionary<Type, List<object>>());
         private ConcurrentDictionary<Type, List<object>> annotations { get { return _annotations.Value; } }
 
+        public static AnnotationList create(ReadOnlySpan<object> annotations)
```
ewoutkramer: We should make this `Create`, not `create`

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #2752, opened 2024-03-27
element types tryparse nullability

### LINKED ISSUE #2749: Element type parsers should return null, not default instances
E.g. DateTime.TryParse in the system types (namespace Element) currently has a non-nullable out parameter, which is set to default values (e.g. a DateTime with a minimal date) when the parse function returns false.

We should change this to nullables with a `[NotNullWhen]` attribute added to them.

This might be breaking, although the caller is not supposed to actually do something with the return value when the function returns `false`, so the consequences would be minor.

Kasdejong:
⚠️ this branch depends on #2741. Merge that branch before reviewing this one

## Description
Made all out variables of ElementModel/Types classes' TryParse methods nullable. Added annotations indicating they are not null when the method returns true.

## Related issues
Closes #2749 

## Testing
Added unit tests to check if an invalid string indeed leads to a null value in the out variable


src/Hl7.Fhir.Base/ElementModel/IBaseElementNavigator.cs
```diff
@@ -26,7 +26,7 @@ public interface IBaseElementNavigator<TDerived> where TDerived : IBaseElementNa
         /// </summary>
         /// <param name="name">Return only the children with the given name.</param>
         /// <returns></returns>
-        IEnumerable<TDerived> Children(string? name = null);
+        IEnumerable<ITypedElement> Children(string? name = null);
```
ewoutkramer: Should be `TDerived`.

src/Hl7.Fhir.Base/ElementModel/ScopedNode.cs
```diff
@@ -81,10 +81,10 @@ private ScopedNode(ScopedNode parentNode, ScopedNode? parentResource, ITypedElem
         public string Name => Current.Name;
 
         /// <inheritdoc/>
-        public string InstanceType => Current.InstanceType;
+        public string InstanceType => Current.InstanceType!;
```
ewoutkramer: This should really be `string?`, right?  There is no real promise that `Current.InstanceType` is not null (although we say it is here).

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #2758, opened 2024-04-04
fhirpath substring with negative length

### LINKED ISSUE #2150: [FhirPath] Change behaviour of `substring()` with negative arguments
Different FP implementations treat having a negative 1st/2nd parameter in `substring()` differently.

According to https://jira.hl7.org/browse/FHIR-37705, negative parameters should be treated as '0'.

Kasdejong:
## Description
The FhirPath substring method should now correctly return a string of length 0 when the length parameter is set to a negative value

## Related issues
Fixes #2150 

## Testing
Added new unit tests for substring method


src/Hl7.FhirPath.Tests/Functions/StringFunctionsTests.cs
```diff
@@ -25,6 +25,16 @@ public void StringSplit()
             Assert.AreEqual(5, StringOperators.FpSplit("Peter,James,Jim,Peter,James", ",").Count());
         }
 
+        [TestMethod]
```
ewoutkramer: Data test method is probably better here.

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #2760, opened 2024-04-04
Added fhirpath environment vars

### LINKED ISSUE #2570: Add environment variables to FhirPath
At this moment, it is possible to add constants to the SymbolTable so you can supply your own external values (just like functions) and refer to them from FhirPath.

However, it is also really useful to supply extrernal variables that can be different from invocation to invocation of a compiled fhirpath statement, variables that are specific for a single run of an expression.

We already pass in an `EvaluationContext` when we run an expression, so a natural extension point would be to make it possible to add context variables here, that can then be referenced from the FhirPath, e.g. with the normal %<variablename> syntax.

Alternatively, we could introduce our own syntax in FhirPath, but that requires changes to the parser and brings little added value.

Kasdejong:
## Description
FhirPath EvaluationContext now takes an IDictionary with KV-pairs denoting environment variables which are evaluated at runtime. These are defined using a '%%' syntax followed by an identifier. This identifier should match a key in the EvaluationContext Dictionary.

## Related issues
Closes #2570 

## Testing
Added unit tests to both the grammar tests and a new test file for environment variables


src/Hl7.Fhir.Base/FhirPath/EvaluationContext.cs
```diff
@@ -29,6 +29,11 @@ public EvaluationContext(ITypedElement resource, ITypedElement rootResource)
             Resource = resource;
             RootResource = rootResource ?? resource;
         }
+        
+        public EvaluationContext(ITypedElement resource, ITypedElement rootResource, IDictionary<string, IEnumerable<ITypedElement>> environment) : this(resource, rootResource)
```
ewoutkramer, line 33: I think we should also add this new parameter (optionally) to the extension methods (like POCO.Predicate, ITypedElement.Select etcetera).

ewoutkramer, line 33: Also update the documentation at https://docs.fire.ly/projects/Firely-NET-SDK/en/latest/fhirpath/evaluation-contexts.html about the new environment stuff.

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #2768, opened 2024-04-10
Added defineVariable to FhirPath

### LINKED ISSUE #2759: Add "defineVariable" to the FhirPath engine
See https://chat.fhir.org/#narrow/stream/179266-fhirpath/topic/.22grouping.22.20elements

Kasdejong:
## Description
Added the defineVariable function to FhirPath as specified in the continuous build of FhirPath

## Related issues
Closes #2759 

## Testing
A bunch of tests, special kudos to @brianpos

### DIFF
--- a/src/Hl7.Fhir.Base/FhirPath/Expressions/ExpressionNode.cs
+++ b/src/Hl7.Fhir.Base/FhirPath/Expressions/ExpressionNode.cs
@@ -305,6 +305,11 @@ namespace Hl7.FhirPath.Expressions
         {
             Name = name ?? throw Error.ArgumentNull("name");
         }
+        
+        internal IEnumerable<ITypedElement> ResolveVariable(Closure context, IEnumerable<Invokee> _)
+        {
+            return context.ResolveValue(Name) ?? throw Error.InvalidOperation($"Variable {Name} not found in environment");
+        }
 
         public string Name { get; private set; }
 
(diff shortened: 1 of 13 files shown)


src/Hl7.Fhir.Base/FhirPath/Expressions/ExpressionNode.cs
```diff
@@ -305,6 +305,11 @@ public VariableRefExpression(string name, ISourcePositionInfo location = null) :
         {
             Name = name ?? throw Error.ArgumentNull("name");
         }
+        
+        internal IEnumerable<ITypedElement> ResolveVariable(Closure context, IEnumerable<Invokee> _)
+        {
+            return context.ResolveValue(Name) ?? throw Error.InvalidOperation($"Variable {Name} not found in environment");
```
ewoutkramer: This should return null now, otherwise the chaining to the resolve in the symbol table will never be called....  or, to avoid nulls, make it TryResolveVariable....

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/FhirPath/Expressions/ExpressionNode.cs
Kasdejong: It seems like this was dead code. The function call was pretty useless anyways, so I guess we unintentionally got rid of it, but it is cleaner without this method anyways :)

ewoutkramer (approved):

Issue #1340, opened 2020-04-21
Binary resource handling


brianpos:
When creating a binary resource it currently switches to the binary stream rather than just pushing the XML/JSON.
There is nothing in the spec that describes this behaviour, and specifically it misses the `X-Security-Context` header value also.
http://hl7.org/fhir/binary.html
What this page does describe is the returning of the content, and this isn't handled by the FHIR Client.

ewoutkramer:
When reviewing the FhirClient's code for SDK 5.0 I have noticed that this codepath has disappeared after the refactorings for the new HttpMessageHandler based implementation. The code is still there, but is logically unreachable.

ewoutkramer:
I am fixing this tomorrow!

PR #2771, opened 2024-04-11
Improved binary support

### LINKED ISSUE #2695: Finish Binary support for the FhirClient
We have a PR in progress here:

https://github.com/FirelyTeam/firely-net-sdk/pull/2474, which is closed until we restart this task again. 

Our FhirClient does not support Binaries correctly, however, the spec is very vague about how to properly support binaries, we should discuss with HAPI and allign.

Kasdejong:
## Description
Added improved binary support for the FhirClient. 

- The FhirClient will attempt to request/recieve binary data instead of "Binary" resources where possible (these will still be parsed into resources upon receiving them, but it puts less stress on the network). If the server sends a resource nonetheless, the old implementation will handle this. This can be switched off in the settings (is on by default). 

- It is now also possible to send binary data, but far from all servers support this, so this will have to be turned on manually.

## Related issues
Closes #2695 

## Testing
Integration test against firely and HAPI


src/Hl7.Fhir.Base/Rest/EntryToHttpExtensions.cs
```diff
@@ -12,38 +12,44 @@
...
-    {     
+    {
         public static HttpRequestMessage ToHttpRequestMessage(
             this Bundle.EntryComponent entry,
             Uri baseUrl,
             IFhirSerializationEngine ser,
             string? fhirVersion,
-            FhirClientSettings settings)
+            FhirClientSettings settings,
+            bool binaryEndpoint = true)
```
ewoutkramer: Since the binary endpoint is the exception, shoud the default not be `false`?

src/Hl7.Fhir.Base/Rest/FhirClientSettings.cs
```diff
@@ -128,6 +129,17 @@ public bool CompressRequestBody
         /// </summary>
         public ParserSettings? ParserSettings = ParserSettings.CreateDefault();
 
+        /// <summary>
+        /// How to transfer binary data when sending data to a Binary endpoint.
+        /// </summary>
+        public BinaryTransferBehaviour BinarySendBehaviour = BinaryTransferBehaviour.UseResource;
+
+        /// <summary>
+        /// How to transfer binary data when receiving data from a Binary endpoint.
```
ewoutkramer: Maybe change the comment to: "Whether we ask the server to return us binary data or a Binary resource."

ewoutkramer: Maybe it should be called "BinaryReceivePreference"

src/Hl7.Fhir.Base/Rest/HttpContentParsers.cs
```diff
@@ -131,84 +147,144 @@ internal record ResponseData(Bundle.ResponseComponent Response, byte[]? BodyData
...
 
-            // Depending on whether this is a failure, the server could have set the content-type incorrectly as well,
-            // do double check whether we need to take this payload as FHIR data
-            Exception? issue = null;
-            Resource? resource = null;
+            // If this is not binary data, try to capture the body as text
+            if (!expectBinaryProtocol)
+                result = result with { BodyText = await message.Content.GetBodyAsString().ConfigureAwait(false) };
+            
+            var isResource = ContentType.GetResourceFormatFromContentType(message.Content.GetContentType()) != ResourceFormat.Unknown;
```
ewoutkramer: Maybe `isResource` should be `usesFhirFormat`.

src/Hl7.Fhir.Shared.Tests/Rest/FhirClientTests.cs
```diff
@@ -895,6 +896,21 @@ public async T.Task TestMultipleMessageHandlersInFhirClient()
             Assert.IsNotNull(testDegatingHandler.LastRequest);
             Assert.IsNotNull(testMessageHandler.LastResponse);
         }
+        
+        
+        [TestMethod]
+        [TestCategory("IntegrationTest"), TestCategory("FhirClient")]
```
ewoutkramer, line 902: Did you test the normal behaviour - have you run some of the other Fhirclient tests to see whether we had regressions?

ewoutkramer (changes requested):

ewoutkramer (approved):

mmsmits:
also closes #1340

Issue #2517, opened 2023-07-05
Support position information in FHIRPath expressions


brianpos:
**Is your feature request related to a problem? Please describe.**
The FhirPath `Expression` and derived classes have support for location information however this is never populated by the sprache parser.

**Describe the solution you'd like**
During the parsing of a fhirpath expression optionally support the inclusion of the position information for each sub-expression in the AST.

**Additional context**
I plan to use this information in the fhirpath-lab project through my open source FhirPath Expression validator nuget package.
This will help significantly with error messages and also highlighting expressions in related tooling.

ewoutkramer:
I did make it work in the MappingLanguage parser (which was based on Superpower). Too bad we cannot use that code anymore. Are we sure we can make this work in sprache? maybe by patching the original code, which we have included anyway?

brianpos:
After spending some time in there, I do believe that we can. Just wanted to check before I really started into it to ensure I hadn't missed something already.

brianpos:
I've progressed with this and have something to demonstrate, but not quite to the level of a branch to show.
@ewoutkramer Any chance we have a catchup to show what I have and see if I'm not doing something too drastic that would prevent it merging in if completed...

brianpos:
As noted in the PR this is now fully functional and combined with the comments/whitepsace branch code will permit full round tripping with ObjectModel, which can permit some manual tweaks to the Model and then back out to a string again without losing the whitespace/comments

PR #2779, opened 2024-05-03
fhirpath parsing now includes comments and whitespace into the Expression object model.

brianpos:
## Description
This extends on the work in the fhirpath parsing changes to be able to capture the whitespace including comments to permit exact round trips with the Expression tree.

## Related issues
#2517 

## Testing
There are many unit tests for this, and also tested with my fhirpath validator project locally to verify that it sucessfully roundtripped all the expressions in the R4B/R5 invariants, searchparameters and about 10 large implementation guides.

## FirelyTeam Checklist


brianpos:
I expect this will be just reviewed in the diffs, then pushed direct into the other branch before (or after) going to develop.

src/Hl7.Fhir.Base/FhirPath/Parser/Grammar.cs
```diff
@@ -165,36 +200,55 @@ public static Parser<Expression> Invocation(Expression focus)
...
         public static Parser<Expression> IndexerInvocation(Expression focus)
         {
-            return Parse.Contained(Expression, Parse.Char('[').TokenIgnoreComments(), Parse.Char(']').TokenIgnoreComments())
-                .Select(ix => new IndexerExpression(focus, ix)).Positioned();
+            return
+                // The whitespace before the '[' is captured in the ws1 token
+                // so that the position information for the indexer isn't confused with the comment
+                from wsLeading in WhitespaceOrComments()
+                from indexer in (
+                    from lparen in Parse.Char('[').Select(v => new SubToken(v)).Positioned()
```
ewoutkramer, line 215: Where is the equivalent of the `TokenIgnoreComments` on the left hand side? So, where are ww capturing the whitespace between the `[` and the following expression?  Or is Expression capturing those?

src/Hl7.Fhir.Base/FhirPath/Parser/Lexer.cs
```diff
@@ -68,7 +68,7 @@ internal partial class Lexer
 
         public static readonly Parser<string> DelimitedIdentifier =
 
-            DelimitedContents('"')
+            DelimitedContents('"') // this isn't to standard
```
ewoutkramer: It was there for backwards compat. Should we axe it?

ewoutkramer (dismissed):

src/Hl7.Fhir.Base/FhirPath/Parser/Lexer.cs
brianpos: I'm OK with it remaining for that purpose. Maybe I'll just update the code comment to say that's why it's there?

src/Hl7.Fhir.Base/FhirPath/Parser/Grammar.cs
brianpos: Expression captures those - just as it does in function arguments between the commas.
There is a unit test that covers this specific case too.
https://github.com/FirelyTeam/firely-net-sdk/blob/6046024baf75120465423899e9c343ede45c4e97/src/Hl7.FhirPath.Tests/Tests/FhirPathGrammarTest.cs#L372

Kasdejong (approved):

ewoutkramer (approved):

PR #2780, opened 2024-05-03
Remove the use of the TestXXXX classes.

ewoutkramer:
The TestPatient/TestXXXX classes are becoming a burden since they have not been updated (it's quite a bit of work) and they start to diverge from the actually generated poco's. Since we're going to make more changes in 6.0, I have taken the step to remove them from our source code.  Our parsing frameworks are now flexible enough that we can just pick STU3 to test against (the TestXXX classes were derived from STU3 anyway).


ewoutkramer:
Note that I had to jump through a hoop because it turns out we have two classes in STU3 and Base with exactly the same name (JsonSerializerOptionExtensions). This led to ambiguous references, unless I turned one (internal) function into an extension function. But in 6.0 we might want to rename this anyway to avoid such confusions.

src/Benchmarks/DeserializationBenchmark.cs
```diff
@@ -28,18 +28,18 @@ public void BenchmarkSetup()
             var xmlFileName = Path.Combine("TestData", "fp-test-patient.xml");
             XmlData = File.ReadAllText(xmlFileName);
 
-            XmlDeserializer = new BaseFhirXmlPocoDeserializer(typeof(TestPatient).Assembly);
-            JsonDeserializer = new BaseFhirJsonPocoDeserializer(typeof(TestPatient).Assembly);
+            XmlDeserializer = new FhirXmlPocoDeserializer();
+            JsonDeserializer = new FhirJsonPocoDeserializer();
```
Kasdejong, line 32: If we're changing this anyways, we may as well use an IFhirSerializationEngine here

src/Hl7.Fhir.Support.Poco.Tests/Rest/FhirClientMockTests.cs
```diff
@@ -19,14 +18,14 @@
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-using T = System.Threading.Tasks;
+using Task = System.Threading.Tasks.Task;
```
Kasdejong, line 21: Let's use T = System.Threading.Tasks with T.Task everywhere

(also this reference looks duplicate/obsolete now)

Kasdejong (changes requested):

src/Hl7.Fhir.Support.Poco.Tests/Rest/FhirClientMockTests.cs
ewoutkramer: I had to choose "Tasks.Task" and "using Tasks =", since you'll get compiler errors if a generic `T` is in scope otherwise.

Kasdejong (approved):

PR #2803, opened 2024-06-25
fix: #2802 - let SnapshotSource reuse self-generated snapshot

### LINKED ISSUE #2802: SnapshotSource does not reuse self-generated snapshot
**Describe the bug**
SnapshotSource should return the same snapshot upon multiple retrievals. Instead, it regenerates the snapshot every time.

**To Reproduce**
1. Create a SnapshotSource, with `regenerate: true`, based on a CachedResolver
2. Get a StructureDefinition by canonical, e.g. from Element
3. Save a reference to the resulting StructureDefinition.Snapshot
4. Repeat step 2
5. Save a reference to the second resulting StructureDefinition.Snapshot
6. Observe that both references are not the same.

**Expected behavior**
When the SnapshotSource has generated (by the Generator) a snapshot, it is marked as `IsCreatedBySnapshotGenerator()`, and it should not be re-generated again upon subsequent retrievals.

**Screenshots**
n/a

**Version used:**
 - FHIR Version: All
 - Version: 5.8.1

**Additional context**
-

cknaap:
## Description
If SnapshotSource detects a snapshot generated by our own SnapshotGenerator, it does not re-generate it.

## Related issues
Fixes #2802.

## Testing
See 2 new unit tests in the PR.

## FirelyTeam Checklist


ewoutkramer:
I do not understand why #2802 is considered a bug. When you set the setting "ForceRegenerateSnapshots" to true, it will re-generate the snapshots, no matter what. This is the documented behaviour of this option:

>  Force expansion of all external profiles, disregarding any existing snapshot components.
     If enabled, the snapshot generator will re-generate the snapshot components of all the core resource and datatype profiles
     as well as of all other referenced external profiles.
     Re-generated snapshots are annotated to prevent duplicate re-generation (assuming the provided resource resolver uses caching).
     If disabled (default), then the snapshot generator relies on existing snapshot components, if they exist.

This PR does not look like a fix, but a behavioural change. The "ForceRegenerateSnapshot" used to *always* regenerate the snapshot, now it will only renegerate it if it was not created by us. That's a breaking change that we should not make in a non-major version.  

Why don't you just set ForceRegenerateSnapshot to `false` instead? This will create a new snapshot if there is none, or if the snapshot is not ours.

ewoutkramer:
Design discussion will be continued on issue #2802.

Kasdejong (dismissed):
I think this PR is overall an improvement now. From what I've read this is what we're going with, so I'll pull it

Kasdejong:
I will also deprecate the setting for good measure, because it seems as though right now it does not do anything. In fact, setting it to false while there is no snapshot will just produce no snapshot now :)

src/Hl7.Fhir.Shims.Base/Specification/Source/SnapshotSource.cs
```diff
@@ -96,7 +98,9 @@ private async Tasks.Task<Resource> ensureSnapshot(Resource res)
         {
             if (res is StructureDefinition sd)
             {
-                if (!sd.HasSnapshot || Generator.Settings.ForceRegenerateSnapshots || !sd.Snapshot.IsCreatedBySnapshotGenerator())
+#pragma warning disable CS0618 // Type or member is obsolete
+                if (!sd.HasSnapshot || (Generator.Settings.ForceRegenerateSnapshots && !sd.Snapshot.IsCreatedBySnapshotGenerator()))
```
mmsmits: As Ewout says in https://github.com/FirelyTeam/firely-net-sdk/issues/2802 this will surely result in regressions. We shouldn't change this.

mmsmits (changes requested):

mmsmits:
This still needs to be done:

> We should obsolete it and **introduce a more explicit setting. Basically there are only two cases:**
    - Add a snapshot if there is none or use whatever is there (new behaviour)
    - Regenerate snapshot once to ensure it is "ours" (and/or at least updated to the lastest version) (current default behaviour)

mmsmits (approved):

PR #2810, opened 2024-07-08
Code changes because of new generator

ewoutkramer:
Used Microsoft's new code generator to regenerate Base and Conformance.

Some points to note:

* This includes work on generating Narrative and Element (they were hand-written)
* Narrative.div has become a true "primitive element", so it has a DivElement + Div. This should be compile-compatible, though the metadata attributes have moved from the "Div" to the "DivElement".
* In ElementDefinition, the order of constraint.xpath and constraint.expression have been swapped, this is not good for XML, asked Gino to double check.
* Generally speaking the changes are due to having more documentation text available.

❗ This PR can only be pulled and reviewed after we have pulled https://github.com/FirelyTeam/fhir-codegen/pull/39.


ewoutkramer:
Unit tests now failing because of 

>In ElementDefinition, the order of constraint.xpath and constraint.expression have been swapped, this is not good for XML, asked Gino to double check.

ewoutkramer:
This PR can be reviewed after we have pulled https://github.com/FirelyTeam/fhir-codegen/pull/39.

mmsmits (approved):

PR #2841, opened 2024-08-20
%resource and %rootResource in default context

### LINKED ISSUE #2177: The default context for FhirPath evaluation is leaving out `%resource`
When you execute one of the extension methods for FhirPath on POCO or ITypedElement - and you don't supply the optional parameter for the context, a default, empty context is provided.  This means that `%resource` and `%rootResource` will return null. Although these environment variables are not used often, they are not expected to be null.  In fact, the constructor for `FhirEvaluationContext` that takes a `ScopedNode` as a parameter, will try to locate the nearest resource in the input data and then set the `%resource` accordingly. It makes total sense that this is the default. In fact, when we ourselves invoke the FP engine in the validator, we *are* always supplying the correct parameters.

Pick a sensible default, instead of null. Preferably the same approach as the constructor taking `ScopedNode`.

Kasdejong:
## Description
Added %resource and %rootresource to default evaluation context on FP extensions

## Related issues
Closes #2177

### DIFF
--- a/src/Hl7.Fhir.Base/FhirPath/Expressions/Closure.cs
+++ b/src/Hl7.Fhir.Base/FhirPath/Expressions/Closure.cs
@@ -28,8 +28,8 @@ namespace Hl7.FhirPath.Expressions
             var newContext = ctx ?? new EvaluationContext();
             
             var node = root as ScopedNode;
-            newContext.Resource ??= node != null ? getResourceFromNode(node) : root;
-            newContext.RootResource ??= node != null ? getRootResourceFromNode(node) : root;
+            newContext.Resource ??= node != null ? getResourceFromNode(node) : root; // if the root is a scoped node, use the resource from the node
+            newContext.RootResource ??= node != null ? getRootResourceFromNode(node) : root; // if the root is a scoped node, use the root resource from the node
             
             var newClosure = new Closure() { EvaluationContext = ctx ?? new EvaluationContext() };
 
--- a/src/Hl7.Fhir.Base/FhirPath/Expressions/Closure.cs
+++ b/src/Hl7.Fhir.Base/FhirPath/Expressions/Closure.cs
@@ -39,8 +39,8 @@ namespace Hl7.FhirPath.Expressions
             newContext.SetIndex(ElementNode.CreateList(0));
             newContext.SetOriginalContext(input);
             
-            if (ctx.Resource != null) newContext.SetResource(new[] { ctx.Resource });
-            if (ctx.RootResource != null) newContext.SetRootResource(new[] { ctx.RootResource });
+            if (ctx?.Resource != null) newContext.SetResource(new[] { ctx.Resource });
+            if (ctx?.RootResource != null) newContext.SetRootResource(new[] { ctx.RootResource });
 
             return newContext;
         }
--- a/src/Hl7.Fhir.Base/FhirPath/Expressions/Closure.cs
+++ b/src/Hl7.Fhir.Base/FhirPath/Expressions/Closure.cs
@@ -25,24 +25,30 @@ namespace Hl7.FhirPath.Expressions
 
         public static Closure Root(ITypedElement root, EvaluationContext ctx = null)
         {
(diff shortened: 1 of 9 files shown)


src/Hl7.Fhir.Shims.STU3AndUp/FhirPath/FhirPathExtensions.cs
```diff
@@ -48,23 +48,23 @@ public static class FhirPathExtensions
 
         /// <inheritdoc cref="FhirPathCompilerCache.Select(ITypedElement, string, EvaluationContext?)"/>
         public static IEnumerable<Base?> Select(this Base input, string expression, FhirEvaluationContext? ctx = null)
-            => CACHE.Select(input.ToTypedElement(), expression, ctx ?? FhirEvaluationContext.CreateDefault()).ToFhirValues();
+            => CACHE.Select(input.ToTypedElement(), expression, ctx ?? new FhirEvaluationContext(input.ToTypedElement())).ToFhirValues();
```
ewoutkramer: I think we should obsolete CreateDefault(), and/or create a new version of it that takes the current focus as a parameter.

ewoutkramer:
> fixed nullability. How did this even work before?

Good question.

src/Hl7.Fhir.Shims.STU3AndUp/FhirPath/FhirPathExtensions.cs
Kasdejong, line 1: When do you think we have the highest rate of success? .ToScopedNode here or just set RootResource to our Poco.ToTE?

src/Hl7.Fhir.Base/FhirPath/Expressions/Closure.cs
```diff
@@ -82,5 +98,13 @@ public virtual IEnumerable<ITypedElement> ResolveValue(string name)
 
             return null;
         }
+
+        private static ITypedElement getResourceFromNode(ScopedNode node) => node.AtResource ? node : node.ParentResource;
```
ewoutkramer: I think we should return ScopedNode here, since in this function we have first hand knowledge that if you pass in a ScopedNode, node and node.ParentResource are ScopedNodes too. This way you can avoid the cast when you call this function, which looks far more suspicious.

src/Hl7.Fhir.Base/FhirPath/Expressions/Closure.cs
```diff
@@ -82,5 +98,13 @@ public virtual IEnumerable<ITypedElement> ResolveValue(string name)
...
+        
+        private static ITypedElement getRootResourceFromNode(ScopedNode node)
+        {
+            var resource = (ScopedNode)getResourceFromNode(node);
+            return resource?.ParentResource?.ContainedResources() is {} c && c.Any() ? resource.ParentResource : resource;
```
ewoutkramer: Doesn't this draw the wrong conclusion if we have a DomainResource with child resources (not contained, but just an element of typeResource), and if we are such a child resource?  Then our boundary shifts up, if though it should not?  I think we should check that the parent node of us (as a resource) is called "contained" instead.

src/Hl7.Fhir.Shims.STU3AndUp/FhirPath/FhirPathExtensions.cs
ewoutkramer: Let's discuss that irl. I don't know. It's too warm here 🥵

src/Hl7.FhirPath.Tests/Tests/EnviromentTests.cs
```diff
@@ -1,7 +1,11 @@
 using Hl7.Fhir.ElementModel;
+using Hl7.Fhir.Introspection;
```
ewoutkramer: Left over extra usings?

ewoutkramer (changes requested):

ewoutkramer (changes requested):
(see my comment)

ewoutkramer (approved):

PR #2842, opened 2024-08-22
Changed iif to work with the correct context

### LINKED ISSUE #2625: Fhirpath `iif` function context...
**Describe the bug**
The fhirpath function `iif` does not consider contexts correctly, and appears to always want to be executed at the root.

**To Reproduce**
1. Using the below test fhirpath expression (and test engine)
```fhirpath
name.iif((trace('iif').use = 'official').trace('tr'), family, given.join(''))
```
:test_tube: [Test with FHIRPath-Lab](https://fhirpath-lab.com/FhirPath?parameters=N4IgpgHgDgTmDO8CWB7AdiAXCNBDAtmAHRJIBmAFBQC4y4DGYFA5KWcwJRECu8YABAF5+zFGTJJ6SXABtORWgybNanADT8yBJDICeGgOZIAbmDREAViiRoWnDiDXg0RtGCwgiAOQCiAFX4KCTg9BydqMBh8GxQZFANdAGVI0xgPAAtqaih4TAB6PPgARzi0MnSkGABaGAAWIlwAL244AHcwACNkCPgiN2o88srHEGNcGGkOmQQsAG0AXQBfIA)

2. When selecting the Firely Engine see that there is no result returned, as the context (as shown in the trace window) is the resource still, the focus wasn't moved to the name collection.

**Expected behavior**
As demonstrated in the tool, the focus should be set to the name before evaluating the arguments to the function.

**Version used:**
 - FHIR Version: R4B (but really version agnostic)
 - Version: 5.3.0

**Additional context**
http://hl7.org/fhirpath/N1/#iifcriterion-expression-true-result-collection-otherwise-result-collection-collection
[Discussion on Zulip with Grahame](https://chat.fhir.org/#narrow/stream/179266-fhirpath/topic/context.20of.20the.20.60iif.20.60)

Kasdejong:
## Description
the FP iif function should now work with the correct focus.

## Related issues
Resolves #2625 

## Testing
Extra iif tests.

### DIFF
--- a/src/Hl7.Fhir.Base/FhirPath/Expressions/SymbolTableInit.cs
+++ b/src/Hl7.Fhir.Base/FhirPath/Expressions/SymbolTableInit.cs
@@ -328,6 +328,12 @@ namespace Hl7.FhirPath.Expressions
         {
             // iif(criterion: expression, true-result: collection [, otherwise-result: collection]) : collection
             // note: short-circuit behavior is expected in this function
+            
+            var focus = arguments.First()(ctx, InvokeeFactory.EmptyArgs);
+            if (focus.Count() > 1)
+            {
+                return runSelect(ctx, [arguments.First(), (closure, _) => runIif(closure, new List<Invokee>{(_, _) => closure.GetThis(), arguments.Skip(1).First(), arguments.Skip(2).First(), arguments.Skip(3).FirstOrDefault()})]);
+            }
 
             var expression = arguments.Skip(1).First()(ctx, InvokeeFactory.EmptyArgs);
             var trueResult = arguments.Skip(2).First();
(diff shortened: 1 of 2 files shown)


Kasdejong:
The expected (in hapi and fplab) behaviour seems to be strange here. It is also not well documented and it relies on reflection. For example, on fp-test-patient, where Patient.contained[0].name is an array of 2 names:
- "Patient.contained[0].name.iif(exists(), 'named', 'unnamed')" is expected to run exists() on the name array and return a single string.
- "Patient.contained[0].name.iif(use = 'official', given, 'not official')" is expected to run use = 'official' on every individual name element (as if there is a hidden select there), and so it returns an array of strings.

We cannot determine at compile time if name is enumerable, so instead, we would have to mark a method as a method which works on collections on or single elements, and add an overload to iif. I feel like this was not the intention, and I would like to discuss with eg @brianpos about what truly is expected of us.

It is worth noting that we pass the official tests by accident! "Patient.contained[0].name.iif(exists(), 'named', 'unnamed')" throws an error (iif not supported on lists), but still manages to return the correct result, since it only runs on the first element on the list :)

ewoutkramer (approved):

PR #2845, opened 2024-08-29
Scoped node cache overhaul

Kasdejong:
## Description
Refactored scoped node caches to internally use dictionaries. This should help with performance.

Note that this might work inversely due to the higher base allocation of dictionaries. We do not have proper control over the children of our scoped nodes right now (they seem to be created with new all the time when they are enumerated). I will make an issue for that, which will need more research

## Related issues
Should be the first step to alleviating https://github.com/FirelyTeam/firely-validator-api/issues/334

### DIFF
--- a/src/Hl7.Fhir.Base/ElementModel/ReferencedResourceCache.cs
+++ b/src/Hl7.Fhir.Base/ElementModel/ReferencedResourceCache.cs
@@ -19,9 +19,9 @@ internal class ReferencedResourceCache : IEnumerable<ScopedNode.BundledResource>
         }
     }
     
-    public IEnumerable<ScopedNode> Resources => _items.Values.OfType<ScopedNode>();
+    internal IEnumerable<ScopedNode> Resources => _items.Values.OfType<ScopedNode>();
 
-    public ScopedNode? resolveReference(string reference)
+    internal ScopedNode? ResolveReference(string reference)
     {
         return _items.TryGetValue(reference, out var node) ? node : null;
     }
--- a/src/Hl7.Fhir.Base/ElementModel/ReferencedResourceCache.cs
+++ b/src/Hl7.Fhir.Base/ElementModel/ReferencedResourceCache.cs
@@ -1,0 +1,33 @@
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Hl7.Fhir.ElementModel;
+
+#nullable enable
+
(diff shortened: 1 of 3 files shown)


src/Hl7.Fhir.Base/ElementModel/ScopedNode.cs
```diff
@@ -183,20 +215,25 @@ public class BundledResource
...
+                _cache.BundledResources = new ReferencedResourceCache(referenceEntryPairs);
             }
+                    
+            else
+                _cache.BundledResources = new ReferencedResourceCache([]);
 
             return _cache.BundledResources;
         }
 
+        internal ReferencedResourceCache? getContainedCache() => _cache.ContainedResources;
```
ewoutkramer: `internal` names should be capitalized too!

ewoutkramer: And these functions are not used ;-)

src/Hl7.Fhir.Base/ElementModel/ScopedNodeExtensions.cs
```diff
@@ -110,14 +111,15 @@ public static string MakeAbsolute(this ScopedNode node, string reference) =>
                 {
                     if (parent.InstanceType == FhirTypeConstants.BUNDLE)
                     {
-                        var result = parent.BundledResources().FirstOrDefault(br => br.FullUrl == url)?.Resource;
-                        if (result != null) return result;
+                        if (parent.BundledResources() is ReferencedResourceCache result) return result.resolveReference(url);
```
ewoutkramer: I rather have a hard cast to indicate to the reader that we KNOW this is always a ReferencedResourceCache.

src/Hl7.Fhir.Base/ElementModel/ReferencedResourceCache.cs
```diff
@@ -0,0 +1,33 @@
...
+        _items = new Dictionary<string, ScopedNode?>();
+        foreach (var item in items)
+        {
+            _items.Add(item.Key, item.Value);
+        }
+    }
+    
+    public IEnumerable<ScopedNode> Resources => _items.Values.OfType<ScopedNode>();
+
+    public ScopedNode? resolveReference(string reference)
```
ewoutkramer: Public function names (public + private) should be capitalized.

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #2847, opened 2024-09-02
Fix isNotA filter code in ValueSet expander

### LINKED ISSUE #2801: Expanded codes of Value set with filter  "op" : "is-not-a" includes the code that should not be included
**Describe the bug**

Value set http://hl7.org/fhir/ValueSet/patient-contactrelationship(326) expanded by car tool for us core package has a filter **"op" : "is-not-a"**. But O is included in the expanded value sets.

```
 "compose" : {
    "include" : [{
      "system" : "http://terminology.hl7.org/CodeSystem/v2-0131",
      "filter" : [{
        "property" : "concept",
        "op" : "is-not-a",
        "value" : "O"
      }]
    }]
  }
```

**Expanded value set**
BP
C
CP
E
EP
F
I
N
O
PR
S
U

**To Reproduce**
1. Verify the Code system and the filter in Value set : https://hl7.org/fhir/R4/valueset-patient-contactrelationship.json.html
3. Verify the Codes from Code system : https://terminology.hl7.org/5.5.0/CodeSystem-v2-0131.json.html
4. Use latest car file of Us-core-test-5.0.1 generated by car tool 2.1.2
5. Go to Codes Table and navigate to Value set Id 326
6. Verify the Codes

**Expected behavior**
Code O should be omitted and not displayed in Codes Table.

**Version used:**
FHIR Version: [R4]
Car tool Version: [2.1.2]
Us-core-test-5.0.1 package

mmsmits:
fixes #2801

### DIFF
--- a/src/Hl7.Fhir.Shims.Base/Model/ValueSetExpansionExtensions.cs
+++ b/src/Hl7.Fhir.Shims.Base/Model/ValueSetExpansionExtensions.cs
@@ -62,26 +62,19 @@ namespace Hl7.Fhir.Model
 
         internal static List<CSDC> RemoveCode(this ICollection<CSDC> concepts, string code)
         {
-            return concepts.removeCodeByPredicate(c => c.Code == code).ToList();
+            return concepts.getNonMatchingCodes(c => c.Code == code);
         }
 
-        private static ICollection<CSDC> removeCodeByPredicate(this ICollection<CSDC> concepts, Predicate<CSDC> predicate)
+        private static List<CSDC> getNonMatchingCodes(this ICollection<CSDC> concepts, Predicate<CSDC> predicate)
         {
-            var itemsToBeRemoved = concepts.Where(c => predicate(c)).ToList();
-
-            //first remove the items that match the predicate (including their children)
-            foreach (var item in itemsToBeRemoved)
+            var filtered = concepts.Where(c => !predicate(c)).ToList();
+            
+            foreach (var concept in filtered.Where(concept => concept.Concept?.Count is > 0))
             {
-                concepts.Remove(item);
+                concept.Concept = concept.Concept.getNonMatchingCodes(predicate).ToList();
             }
-
-            //then check the children of the items that were not removed to see if they match the predicate
-            foreach (var concept in concepts)
-            {
-                removeCodeByPredicate(concept.Concept, predicate);
-            }
-
-            return concepts;
+            
+            return filtered;
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Hl7.Fhir.Shims.Base/Model/ValueSetExpansionExtensions.cs
```diff
@@ -67,27 +67,23 @@ internal static List<CSDC> RemoveCode(this ICollection<CSDC> concepts, string co
 
         private static ICollection<CSDC> removeCodeByPredicate(this ICollection<CSDC> concepts, Predicate<CSDC> predicate)
```
Kasdejong: Right now we have side effects: we remove items, but we return the collection because we want to keep the fluent syntax. We should really be making this pure, that way the signature makes sense.

src/Hl7.Fhir.Shims.Base/Model/ValueSetExpansionExtensions.cs
```diff
@@ -67,27 +67,23 @@ internal static List<CSDC> RemoveCode(this ICollection<CSDC> concepts, string co
...
         {
-            foreach (var concept in concepts)
+            var itemsToBeRemoved = concepts.Where(c => predicate(c)).ToList();
```
Kasdejong: We iterate 3 times now. We can get away with only doing this once, by just putting all the calls in a foreach loop, or better yet: a select (if we are going pure)

Kasdejong (changes requested):

src/Hl7.Fhir.Shims.Base/Model/ValueSetExpansionExtensions.cs
mmsmits: You are suggesting to make this a void you mean?

src/Hl7.Fhir.Shims.Base/Model/ValueSetExpansionExtensions.cs
mmsmits: I was hoping to do this in a single forloop (or linq statement), but I couldn't figure it out without duplicating collections, and making it a mess. 
You have a suggestion?

Kasdejong: In this case duplicating subsets of collections is a fact of life I think. The most optimized version is a big chain call, since the compiler can chain the methods together into a single loop. I will suggest a commit (once I find out how to)

Kasdejong: Alright, it seems like i'm not able to suggest it the way I wanted to. I would like to suggest something along the lines of:
```cs
internal static List<CSDC> RemoveCode(this ICollection<CSDC> concepts, string code)
{
    return concepts.getNonMatchingCodes(c => c.Code == code);
}

private static List<CSDC> getNonMatchingCodes(this ICollection<CSDC> concepts, Predicate<CSDC> predicate)
{
    var filtered = concepts.Where(c => !predicate(c)).ToList();
    
    foreach (var concept in filtered.Where(concept => concept.Concept?.Count is > 0))
    {
        concept.Concept = concept.Concept.getNonMatchingCodes(predicate).ToList();
    }
    
    return filtered;
}

PR #2877, opened 2024-09-27
Added unit test for generating Snapshots for multiple nested extensions

### LINKED ISSUE #2799: Snapshot generator does not expand complex extensions defined in a an extension.
The bug was originally submitted to Forge:
https://firely.atlassian.net/browse/FOR-872

The project to reproduce the issue is attached.
The issue is for file: NLLArticleInfoCorrect.xml.

The snapshot generator method Generate returns 2 of the same errors:

`https://electronichealth.se/fhir/StructureDefinition/NLLArticleInfo2 (Element Extension.extension.extension has neither a type nor a nameReference.) 
`

They correspond to the extension element definitions for **currentProductName** and **history**.
These two element definitions are not expanded.

 - FHIR Version: R4
 - Version: 5.8.1

[swedishnationalmedicationlist(5).zip](https://github.com/user-attachments/files/15788483/swedishnationalmedicationlist.5.zip)

mmsmits:
Fixes: #2799
SnapshotGenerator already works for this, but nice to not have to write this unit test ever again in the future

### DIFF
--- a/src/Hl7.Fhir.Specification.Shared.Tests/Snapshot/SnapshotGeneratorTest.cs
+++ b/src/Hl7.Fhir.Specification.Shared.Tests/Snapshot/SnapshotGeneratorTest.cs
@@ -35,7 +35,6 @@ using Hl7.Fhir.Support;
 using Hl7.Fhir.Utility;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using NSubstitute;
-using NSubstitute.Exceptions;
 using System;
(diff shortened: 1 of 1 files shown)


Kasdejong (approved):
We do have TestProfileArtifactSource in the SDK as well (Specification.Shared.Tests/Snapshot/TestProfileArtifactSource.cs). Maybe we can make that the new standard. I'm not sure if it matters for this one. At least the style is consistent :)

PR #2886, opened 2024-10-08
Remove all non-inheritable extensions from the base profile before snapshotting

### LINKED ISSUE #1814: Make sure we do not propagate "publishing extensions" while creating a snapshot
The .NET snapshot generator still propagates a set of extensions that do not make sense to be propagated from a base/core profile into a derived profile.

A full solution is described here: https://jira.hl7.org/browse/FHIR-28441, but this requires authors of some extensions to align with this solution too - which won't be the case for older extensions.  So, we should at least make sure we don't propagate the list of extensions enumerated in Lloyd comment at the bottom of the page (https://jira.hl7.org/browse/FHIR-28441?focusedCommentId=177188&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-177188).

mmsmits:
fixes #1814


src/Hl7.Fhir.Specification.STU3.Tests/Snapshot/SnapshotGeneratorTest.cs
```diff
@@ -2433,8 +2433,8 @@ private static bool isAlmostExactly(ElementDefinition elem, ElementDefinition ba
             }
 
             // Also ignore any Changed extensions on base and diff
-            elemClone.RemoveAllConstrainedByDiffExtensions();
-            baseClone.RemoveAllConstrainedByDiffExtensions();
+            elemClone.RemoveAllNonInheritableExtensions();
```
ewoutkramer, line 2436: Does this method replace the original ones? So RemoveallNonInheritableExtension replaces RemoveAllConstrainedByDiffExtensions?  If so, should we not remove/obsolete the original method?

src/Hl7.Fhir.STU3/Specification/Snapshot/SnapshotGeneratorExtensions.cs
```diff
@@ -73,6 +72,47 @@ public static void RemoveAllConstrainedByDiffExtensions<T>(this IEnumerable<T> e
             }
         }
 
+        internal static void RemoveAllNonInheritableExtensions(this Element element)
```
ewoutkramer: Can we use a bit code in a shared project instead of copy/paste? I know the snapget is unique to both, but maybe this overlaps?

ewoutkramer (changes requested):

src/Hl7.Fhir.STU3/Specification/Snapshot/SnapshotGeneratorExtensions.cs
mmsmits: I moved the extensions to base.

src/Hl7.Fhir.Specification.STU3.Tests/Snapshot/SnapshotGeneratorTest.cs
mmsmits: Well, I thought about that, but didn't dare to since Forge might be using the ConstraintByDiff extension one maybe?
And that is a bit faster (I don't know if it might be noticeable in a UI).
@Rob5045  any feedback here?

Rob5045: Forge does not use the RemoveAllConstrainedByDiffExtensions extension.

mmsmits: Thanks, I'll Obsolete is then

src/Hl7.Fhir.STU3/Specification/Snapshot/SnapshotGeneratorExtensions.cs
Rob5045: I'm working on supporting extensions on element definnition properties in Forge. That would mean the UI won't be cluttered with these I guess? That would be nice.
image

src/Hl7.Fhir.Specification.STU3.Tests/Snapshot/SnapshotGeneratorTest.cs
Rob5045: I'm working on supporting extensions on element definition properties in Forge. That would mean that the UI won't be cluttered with these I guess? That would be nice.
image

Rob5045: How about http://hl7.org/fhir/StructureDefinition/elementdefinition-bindingName? I see that one all the time as well.

mmsmits: This is not on the list:
We only made this fix for category 5 of this list:
https://jira.hl7.org/browse/FHIR-28441?focusedCommentId=177188&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-177188

src/Hl7.Fhir.Base/Specification/Snapshot/SnapshotGeneratorExtensions.cs
```diff
@@ -53,6 +54,7 @@ public static void RemoveConstrainedByDiffExtension(this IExtendable element)
         }
 
         /// <summary>Recursively removes all instances of the <see cref="CONSTRAINED_BY_DIFF_EXT"/> extension from the specified element definition and all it's child objects.</summary>
+        [Obsolete("Use RemoveAllConstrainedByDiffExtensions(this Element element) instead.")]
```
ewoutkramer:
```suggestion
        [Obsolete("Use RemoveAllNonInheritableExtensions(this Element element) instead.")]
```

src/Hl7.Fhir.Base/Specification/Snapshot/SnapshotGeneratorExtensions.cs
```diff
@@ -64,6 +66,7 @@ public static void RemoveAllConstrainedByDiffExtensions(this Element element)
         }
 
         /// <summary>Recursively removes all instances of the <see cref="CONSTRAINED_BY_DIFF_EXT"/> extension from all the elements and their respective child objects.</summary>
+        [Obsolete("Use RemoveAllConstrainedByDiffExtensions(this Element element) instead.")]
```
ewoutkramer:
```suggestion
        [Obsolete("Use RemoveAllNonInheritableExtensions(this IEnumerable<T> elements) instead.")]
```

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/Specification/Snapshot/SnapshotGeneratorExtensions.cs
mmsmits: Oops

ewoutkramer (approved):

PR #2909, opened 2024-10-10
changed type of FhirEvalContext.TerminologyService

### LINKED ISSUE #2590: `FhirEvaluationContext.TerminologyService` should be of type `ICodeValidationTerminologyService`
In the current implementation, the terminology service defined within the `FhirEvaluationContext` class is of the type `ITerminologyService`, which is overly broad for our specific needs. Our sole requirement within FhirPath is to utilize the `ValueSetValidateCode()` method. Therefore, it would be more appropriate to reconfigure `FhirEvaluationContext.TerminologyService` to be of the type `ICodeValidationTerminologyService` to align with our focused functionality.

Kasdejong:
## Description
Changed the type of FhirEvaluationContext.TerminologyService to ICodeValidationTerminologyService

## Related issues
Closes #2590


src/Hl7.Fhir.Base/FhirPath/FhirEvaluationContext.cs
```diff
@@ -69,7 +69,7 @@ public FhirEvaluationContext(ScopedNode node)
         /// </summary>
         public static new FhirEvaluationContext WithResourceOverrides(ITypedElement? resource, ITypedElement? rootResource = null) =>
             (FhirEvaluationContext)EvaluationContext.WithResourceOverrides(resource, rootResource);
-        public ITerminologyService? TerminologyService { get; set; }
```
alexzautke, line 72: Wouldn't this hurt us when trying to implement the other terminology-related fhirpath functions except member-of?

Kasdejong: See discussion in #2590. I am not sure, just doing what I'm told :)

alexzautke:
Haha. See, I had the same argument last year.

ewoutkramer:
Yeah, and I think you are right. We should probably pick an interface that has memberOf() as well.  Of course, it's just a matter of time before people want more, though. But that does not mean all termservices also implement it. That's the beauty of making it explicit.

alexzautke:
Would it make sense to add a way for the ITerminologyService to report which methods are supported by the service?

Kasdejong:
Both ValueSetValidateCode (memberOf) and Subsumes are defined in ICodeValidationTerminologyService. It is just operations like Closure and Expand which are not in this interface.

ewoutkramer (approved):

PR #2918, opened 2024-10-16
#2901 Add support for overflow to POCOs

### LINKED ISSUE #2901: Add overflow Dictionary to POCOs
We need to add a Dictionary to base, alter the implementation of IReadOnlyDictionary to include the elements from this dictionary, maybe implement IDictionary (or the equivalent writer functionality from GEtElementPairs). 

Also, to cater for incorrect data, we need to add support for ErrorAnnotations, and supporting classes.

ewoutkramer:
## Description
This PR is the first shot at adding support for overflow to POCOs. It should not introduce major breaking changes. 

## Remarks
⚠️ I had introduced the `this[]` operator for setting values, as we already had a getter for it hidden behind `IReadOnlyDictionary`.  This meant that an existing `this[]` operator on the `Parameter` resource had to be removed as it conflicted with this operator in `Base`. Later I implemented this operator explicitly, so the problem does not exist anymore, but just for future flexibility I am removing this minor feature. This is a breaking change.

⚠️ Since pocos now implement IDictionary, they implement ICollection, which triggered bugs since we used a comparison to `ICollection<T>` to detect whether an element repeated or not. I changed the method in ReflectionHelper to detect `IList<>` instead. This is a breaking change. 

⚠️  Also review and pull  https://github.com/FirelyTeam/fhir-codegen/pull/42, which are the changes to the codegen necessary to produce POCO's for this PR.

## Related issues
Fixes #2901.


ewoutkramer:
`error CP0002: Member 'Hl7.Fhir.Model.Parameters.ParameterComponent Hl7.Fhir.Model.Parameters.this[string].get' exists on [Baseline] lib/net8.0/Hl7.Fhir.Base.dll but not on lib/net8.0/Hl7.Fhir.Base.dll [D:\a\1\s\src\Hl7.Fhir.Base\Hl7.Fhir.Base.csproj]`

Use `GetSingle()` instead.

ewoutkramer:
`error CP0002: Member 'bool Hl7.Fhir.Utility.ReflectionHelper.IsTypedCollection(System.Type)' exists on [Baseline] lib/net8.0/Hl7.Fhir.Base.dll but not on lib/net8.0/Hl7.Fhir.Base.dll [D:\a\1\s\src\Hl7.Fhir.Base\Hl7.Fhir.Base.csproj]`

Use `IsListCollection()`. instead

Kasdejong:
There are too many things I want to comment on to do this via PR comments. Not because it's bad, but because I have not worked on this so I do not know what complications impacted your design :) We should discuss in person or over slack.

Kasdejong (approved):

PR #2921, opened 2024-10-16
Defined IScopedNode

### LINKED ISSUE #2894: Design and extend "IScopedNode"
We need a new interface with at least Parent (and maybe Resolve()). This is the "ITypedElement" for POCOs. See #2893 for more discussion on the design need and issues for this interface.

Kasdejong:
Defined IScopedNode, which is what ITypedElement was once supposed to be. It contains the set of information necessary for validation and FP and is implemented explicitly by Base and through ScopedNode by ITypedElement.

Closes #2894.

## Breaking changes
moved ScopedNodeExtensions from ElementModel to Model to reflect its new role.


src/Hl7.Fhir.Base/ElementModel/ScopedNode.cs
```diff
@@ -87,6 +88,19 @@ private ScopedNode(ScopedNode parentNode, ScopedNode? parentResource, ITypedElem
         /// <inheritdoc/>
         public string Name => Current.Name;
 
+        /// <summary>
+        /// Will be replaced by a different implementation in the future.
+        /// </summary>
+        public NodeType Type => this switch
+        {
+            { AtResource: true } when Current.Children("contained").Any() => NodeType.DomainResource | NodeType.Resource,
```
ewoutkramer, line 96: Being a DomainResource is something structural, most DomainResource instances won't actually have contained resources, so this will not work if you need structural information.  I guess you need this for the %rootResource stuff?  Well, we have nothing better I am afraid, but this is certainly flakey.

src/Hl7.Fhir.Base/ElementModel/ScopedNodeExtensions.cs
```diff
@@ -1,145 +0,0 @@
...
-
-using Hl7.Fhir.Rest;
-using Hl7.Fhir.Support.Poco;
-using Hl7.FhirPath.Sprache;
-using System;
-using System.Linq;
-
-#nullable enable
-
-namespace Hl7.Fhir.ElementModel
```
ewoutkramer, line 17: This is a breaking change - you should put it in the PR.  I think if you do not rename this from ScopedNodeExtensions to ScopedNodeHelpers, then you would still have compile compatibility?

src/Hl7.Fhir.Base/Model/IScopedNode.cs
```diff
@@ -0,0 +1,255 @@
...
+    /// Enumerates the children of this instance. If name is given, only the children with that name are returned.
+    /// </summary>
+    /// <param name="name">If given, specifies the children to return based on this parameter.</param>
+    /// <returns>An IEnumerable containing the children of this instance, including scope data.</returns>
+    new IEnumerable<IScopedNode> Children(string? name = null);
+
+    /// <summary>
+    /// Name of the node, e.g. "active", "value".
+    /// </summary>
+    new string Name { get; }
```
ewoutkramer, line 51: I cannot remember why we need to do a `new` here.  For `Children()`, sure, since the return type is different, but this is exactly the property in the Base?

src/Hl7.Fhir.Base/Model/IScopedNode.cs
```diff
@@ -0,0 +1,255 @@
...
+        return scan?.Parent; // return the container (DomainResource around contained)
+    }
+
+    /// <summary>
+    /// Resolve the given reference within the context of the given node. This node should be of type reference.
+    /// </summary>
+    /// <param name="maybeScopedNode">A node representing a reference</param>
+    /// <param name="externalResolver">An external resolver</param>
+    /// <returns></returns>
+    /// TODO: This method should not be generic! It should work on IScopedNode ONLY! this is just for testing
```
ewoutkramer: When are we resolving this TODO?

src/Hl7.Fhir.Base/Model/IScopedNode.cs
```diff
@@ -0,0 +1,255 @@
...
+                return true;
+            }
+        }
+        
+        result = null;
+        return false;
+    }
+}
+
+public static class ScopedNodeHelpers
```
ewoutkramer: I feel this should be in a separate file, in fact we had that already (ScopedNodeExtensions)

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/ElementModel/ScopedNode.cs
Kasdejong: I was aware that it wasn't very solid, but for now I only use it to determine whether I should try to resolve in its contained resources, so no harm done. "HasContained" would be a better name for this since that is essentially what it is used for now, but it would violate the concept of the NodeType enum

src/Hl7.Fhir.Base/ElementModel/ScopedNodeExtensions.cs
Kasdejong: Do we consider this breaking generally? if they used the extension method syntax before, this would not be source breaking, only if they use ScopedNodeExtensions.Resolve(node, ...) this is breaking. 

[EDIT] In any case I moved it from elementModel to Model, so it is breaking nonetheless

src/Hl7.Fhir.Base/Model/IScopedNode.cs
Kasdejong: I used new for every property that will stop being defined in IScopedNode if we unlink it from ITypedElement. Since IScopedNode : ITypedElement and ITypedElement already has the property Name, we need new here.

src/Hl7.Fhir.Base/Model/IScopedNode.cs
Kasdejong: We need to unlink from ITE first. I don't remember exactly why, but it was the case.

Kasdejong: I will try to make this non-generic and see what errors I get

src/Hl7.Fhir.Base/ElementModel/ScopedNode.cs
ewoutkramer: But if the rule is that we should not go past a DomainResource boundary, how can we make sure we don't, even if the DomainResource has no contained resources? Or is that not relevant?

src/Hl7.Fhir.Base/ElementModel/ScopedNodeExtensions.cs
ewoutkramer: Most people won't do the explicit `ScopedNodeExtensions.Resolve(node, ...)`, but it is sometimes necessary, e.g. if these extension methods are in assemblies that are referenced with an `external alias` (which is a technique used to use R3/R4 etc versions of the SDK in a single application).  So, we don't have such a rule, but we might say: "If the extension method is in an assembly that is not a satellite (so in base or conformance), then such a namechange is ok".

But then again, why rename this in the first place?

ewoutkramer: Moving it is something we *could* make bw compatible (dotnet has features for moving things to different assemblies), but we don't have to, since this is a major new release. Just make sure you put it in the wiki with breaking changes!

ewoutkramer (approved):

PR #2922, opened 2024-10-17
Remove the Charset from the Accept Header and replace it with the Accept-Charset Header.

whyfate:
## Description
In the HTTP standard, `Accept` Header does not include charset.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept


Kasdejong:
This change is binary-breaking, but not source-breaking. Please update the package validation compatibility supression file (steps are in the readme) and include them in your PR!

whyfate:
> This change is binary-breaking, but not source-breaking. Please update the package validation compatibility supression file (steps are in the readme) and include them in your PR!

Is it like this？

ewoutkramer (approved):

ewoutkramer:
Thanks, @whyfate !

PR #2939, opened 2024-10-18
2930 Clean up IsNested/BackboneType attribute information

### LINKED ISSUE #2930: Remove `IsNested` property from the FhirTypeAttribute
This property has been superceded by the `[BackboneElement]` attribute, which also specifies the backbone path.

ewoutkramer:
Fixes #2930.

* I removed the old `IsResource` attribute parameter, since that is not necessary anymore, resources will always be derived from `Resource`.
* I changed FhirType.IsNestedType to FhirType.IsBackboneType to align with common terminology.
* I changed the name of the FhirType to be the definition path of the BackboneElement, since that is generally considered to be its name. This also means we don't need a separate BackboneTypeAttribute(.DefinitionPath) anymore.

⚠️ This removes IsResource from the FhirTypeAttribute. One can just check against the `Resource` subtype instead.
⚠️ This renames the IsNestedType parameter on FhirTypeAttribute to IsBackboneType. 
⚠️ This removes the `BackboneTypeAttribute` class. Check IsBackboneType, the type's name is the definition path.
⚠️ The DefinitionPath property of ClassMapping is now gone, since it is the same as the ClassMapping.Name for backbone types.

Kas: Also changed the signature of an evaluation context method, which seemed like the right thing to do to fix a failing unit test.


ewoutkramer:
Review https://github.com/FirelyTeam/fhir-codegen/pull/43 too.

Kasdejong (approved):

PR #2943, opened 2024-10-22
Use proper snapshot generation settings, fixed obsolete warnings

Kasdejong:
## Description
- Changed the SnapshotGenerator to use the new RegenerationBehaviour setting. Also aligned this behaviour more with the "expected" behaviour.
- Removed the obsolete warnings from our own code
- In doing so, changed the signature of EvaluationContext.WithResourceOverrides. This is breaking, but since this member has only existed for a month or so and the current signature was confusing and impractical, we deemed it necessary to change this BEFORE 6.x.x

### DIFF
--- a/src/Hl7.Fhir.Specification.Shared.Tests/Snapshot/SnapshotGeneratorManifestTests.cs
+++ b/src/Hl7.Fhir.Specification.Shared.Tests/Snapshot/SnapshotGeneratorManifestTests.cs
@@ -799,6 +799,8 @@ namespace Hl7.Fhir.Specification.Tests
                 Id = id ?? throw new ArgumentNullException(nameof(id));
                 Assert.AreEqual(id, generated.Id);
                 this.Tracer = this.Trace;
+
+                this.WithResourceOverrides(Generated);
             }
 
             void Trace(string msg, IEnumerable<ITypedElement> elems)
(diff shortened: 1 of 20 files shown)


src/Hl7.Fhir.Specification.Shared.Tests/Snapshot/SnapshotGeneratorManifestTests.cs
```diff
@@ -788,7 +788,7 @@ class SnapshotEvaluationContext : FhirEvaluationContext
 
             public SnapshotEvaluationContext(
                 string testPath, IResourceResolver resolver, string id,
-                StructureDefinition input, StructureDefinition generated) : base(generated.ToTypedElement())
+                StructureDefinition input, StructureDefinition generated)
```
ewoutkramer, line 791: Shouldn't this set the WithResource()?

src/Hl7.FhirPath.R4.Tests/PocoTests/FhirPathTest.cs
```diff
@@ -200,34 +200,40 @@ public void TestFhirPathRootResource()
...
+            EvaluationContext ctx = new FhirEvaluationContext();
+            Assert.AreEqual("contained-1", testNode.Scalar("%resource.id", ctx));
+            Assert.AreEqual("patient-1", testNode.Scalar("%rootResource.id", ctx));
 
             // focus on the id of the contained resource
-            ctx = new FhirEvaluationContext(patBundle.Select("entry.first().resource.contained.id")?.FirstOrDefault() as ScopedNode);
-            Assert.AreEqual("contained-1", patBundle.Scalar("%resource.id", ctx));
-            Assert.AreEqual("patient-1", patBundle.Scalar("%rootResource.id", ctx));
+            testNode = patBundle.Select("entry.first().resource.contained.id")?.FirstOrDefault() as ScopedNode;
+            ctx = new FhirEvaluationContext();
```
ewoutkramer: Do we need to re-create a new context every time?

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #2953, opened 2024-10-25
Replace PocoBuilder to support overflow

### LINKED ISSUE #2947: Adapt PocoBuilder to use overflow
- [ ] Copy & paste the code...call it "PocoBuilder2" ;-)
- [ ] Simplify code to be more performant and less flexble.
- [ ] Let it use overflow for unknown elements.

ewoutkramer:
Wrote a replacement for PocoBuilder (for now, called PocoBuilderNew, they are internal classes anyway) that implements the new overflow and uses the Dynamic datatype classes for unknown types.  

* I have also replaced the `ToPoco()` extension methods to use the new builder
* I have patched the old builder to eventually call the new builder.
* Consequently I was able to throw away all the old ComplexTypeReader stuff.

Fixes #2947 .

## Breaking changes
* The `DefaultModelFactory` has been removed. It had not been in use for years, except by our own tests. With the cleanup of the PocoBuilder (ITypedElement->POCO) infrastructure, we decided to remove this class.


src/Hl7.Fhir.Base/ElementModel/PocoBuilderNew.cs
```diff
@@ -0,0 +1,211 @@
...
+using Hl7.Fhir.Introspection;
+using Hl7.Fhir.Model;
+using Hl7.Fhir.Utility;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using ET = Hl7.Fhir.ElementModel.Types;
+using ElementMappingInfo = (Hl7.Fhir.Introspection.ClassMapping Class, Hl7.Fhir.Introspection.PropertyMapping? Property);
+
+namespace Hl7.Fhir.Serialization;
```
Kasdejong: Possibly better suited for Elementmodel namespace

src/Hl7.Fhir.Base/ElementModel/PocoBuilderNew.cs
```diff
@@ -0,0 +1,211 @@
...
+            return (mapping, elementMapping);
+        }
+
+        // Ok, so the node does not have a type, or the type is not known. So, let's use one
+        // of the applicable dynamic types. If the node has a resource name (but it was unknown),
+        // we'll create a DynamicResource. If the node has a Value, it's a DynamicPrimitive,
+        // otherwise it's a DynamicDataType.
+        // Design question: there might be a "strict" option, where we will not create Dynamic types
+        // for unknown types, but throw an error instead.
+        if(node.Value is not null)
```
Kasdejong: Check if instancetype starts with a lowercase letter, since even a primitive could have only extensions

src/Hl7.Fhir.Base/Model/Base.cs
```diff
@@ -111,14 +113,21 @@ protected virtual bool TryGetValue(string key, [NotNullWhen(true)] out object? v
...
 
-
 /// <summary>
 /// A dynamic data type that can hold any element.
 /// </summary>
-public class DynamicDataType : DataType
+[Serializable]
+[DataContract]
+[FhirType("DynamicDataType","http://fire.ly/fhir/StructureDefinition/DynamicDataType")]
+public class DynamicDataType : DataType, IDynamicType
```
Kasdejong: Let's consider making another part of base, and keeping the Base.cs file as close to the core model as possible

src/Hl7.Fhir.Support.Poco.Tests/Model/TypedElementToPocoTests.cs
```diff
@@ -0,0 +1,101 @@
...
+                Contact = [new Patient.ContactComponent
+                {
+                    Name = new HumanName(family: "Doe", given: ["Jane"]),
+                    Relationship = [new CodeableConcept("http://nu.nl", "relation")]
+                }]
+            };
+
+        var poco = toPoco(subject);
+
+        poco.Active.Should().Be(true);
```
Kasdejong: Maybe try this with IsExactlyEqualTo?

Kasdejong (changes requested):

Kasdejong (approved):

PR #2973, opened 2024-11-19
Remove Children, NamedChildren and IDictionary

### LINKED ISSUE #2927: Remove properties Children and NamedChildren
These properties list the properties of an instance, and are replaced by GetElementPairs() and IReadOnlyDictionary. We should remove them or rewrite these as (obsolete) extension methods, so they can be used by people (most notably Forge) to have a migration path.  We could keep the original properties in, but throwing NotImplemented and add an Obsolete to them to point to the new extension methods.

ewoutkramer:
**Breaking Changes**

In our effort to simplify, this PR removes three ways to navigate down the tree:

* Base.Children
* Base.NamedChildren
* IReadOnlyDictionary/IDictionary.

This also makes sure none of our code is dependent on IDictionary anymore, and we are using "Base" (or ITypedElement) instead to navigate.

Fixes #2927.

⚠️ Review https://github.com/FirelyTeam/fhir-codegen/pull/45 too.


Kasdejong (changes requested):
We should document the removal of `AsReadOnlyDictionary` in the wiki pages. Apart from that, no comments.

Kasdejong (approved):

PR #2975, opened 2024-11-20
Revert snapshot generation settings

mmsmits:
This PR reverts the previously added "simplification" of the snapshot generation settings, since they introduced some regressions.


mmsmits:
It makes me sad that we have to do this now, because I still believe this is correct. But is causes too many regressions for which we currently can identify the reason. 
I'll put an issue on the backlog to check if these regressions are correct.

Kasdejong (approved):

PR #2987, opened 2024-12-03
Moved SearchParamDefinition from STU3+ to a separate class in Base.

### LINKED ISSUE #2531: Migrate SearchParamDefinition into the Base versionless level
The SearchParamDefinition class is an internal class used in each of the versions ModelInfo classes to hold the code generated search parameter code.
I'd love to have this definitional class in Base so that it can be used in generic services that handle more than one version.
My specific case is for a versionless Search Expression validator - It will be using this class for it's information, and likely have an extension method to take the version specific SearchParameter and produce this class for validations (and also be able to scan any allready pre-generated.
As shown in this code:
https://github.com/brianpos/UploadFIG/blob/98408a0d8ad33993d57e90905f8ed22e0c360a77/UploadFIG/ObjectValidator.cs#L88

Migrate the class into the Base assembly

The only other alternatives for me are to replicate the code in my validator to each version, or have my own generic class and request project users provide the mapping.

ewoutkramer:
SearchParamDefinition used to be in the sattelites, since it has a `Target` element that was bound to ResourceTypes. Now, with the publication of R5, we have a "version independent" list of resources, so by using this enum we are able to move SearchParamDefinition out to the Base assembly.

This also required some changes to code generation, which need to be review as well: https://github.com/FirelyTeam/fhir-codegen/pull/49.

Since the SPD class was nested in ModelInfo (which is in the sattellites), I needed to move this class out of ModelInfo, which is a breaking change as well.

## Breaking changes ##
* `ModelInfo.SearchParamDefinition` is moved out of `ModelInfo` and now lives in the assembly root namespace.
* The element `Target` is changed from `ResourceType` to `VersionIndependentResourceTypesAll`.
* `VersionIndependentResourceTypesAll` is moved from R5 to the Base assembly.


Kasdejong:
maybe a PR description? I'll review regardless :)

ewoutkramer:
working on it.

mmsmits:
fixes #2531

Kasdejong (approved):

Issue #2903, opened 2024-10-09
Make StructureDefinition version agnostic


Kasdejong:
We should move StructureDefinition to the Base namespace

mmsmits:
A property is changed from non-repeating to repeating. Changed are too drastic to have one single base.
We can use our new overflow in SDK 6 to make agnostic code around StructureDefinitions.

PR #2988, opened 2024-12-05
Introduce IEqualityComparer instead of Matches/IsExactly

### LINKED ISSUE #2928: Turn Matches and IsExactly into `EqualityComparer<Base>`
We can rewrite Matches and IsExactly in terms of GetElementPairs(), or even IReadOnlyDictionary. This could be done as in-place replacements for the current methods (who then don't need to be virtual anymore), but we could also turn them into a `ExactEqualityComparer<T>` and `PatternEqualityComparer<T>` to align with the normal .NET strategy of implementing `System.Collections.Generic.EqualityComparer<Base>` for these kind of purposes.

Again, as a guide to the user, we can leave the old methods in, throw not implemented and add an Obsolete with a message to indicate you should use the comparers.

ewoutkramer:
This PR removes the generated Matches/IsExactly and replaces it with a re-useable `CompareChildren()` function. The original Matches/IsExactly have been turned into extension methods. It is now possible to create custom comparers for other types of equality.

For efficiency, I also made the dictionary support functions public, since that was still a minor todo.

Fixes #2928.

⚠️ Make sure to also review https://github.com/FirelyTeam/fhir-codegen/pull/50 with the changes to the codegenerator.

## Breaking changes ##
* The `IDeepComparable` interface has disappeared, the `Matches` and `IsExactly` now simply take a `Base` as parameter. Also, it is now possible to write new custom comparison operations using the new `CompareChildren()` function, given a custom implementation of `IEqualityComparer<T>`.


Kasdejong:
Wow! trying to view the changes almost crashes my browser... it uses 3GB of memory, maybe some GH pagination would be in order :) I can't help but notice you moved StructureDefinition from Conformance to Shims.Base. Is that intentional right now, or was that a change you made locally as a PoC for #2903?

Kasdejong (approved):

PR #2952, opened 2024-10-25
Fix wrong attachment url datatype in PocoSummaryProvider

mmsmits:
## Description
Changed datatype of attachment.Url from Url to Uri in STU3.

## Related issues
fixes: https://github.com/FirelyTeam/firely-validator-api/issues/383

## Testing
Added new roundtrip test of comparing the SDSummaryProvider vs PocoSummaryProvider for all datatypes.

PR #3007, opened 2025-01-13
2931 Remove ISystemAndCode

### LINKED ISSUE #2931: Remove `ISystemAndCode` in favor of `ICoded`
`ISystemAndCode` is only used to get the system+code out of a `Code<T>`.  There is a less direct route already available via ICoded, which will work on any coded type to get a `Fhir.Coding` type, which has a system+code.

- [ ] Extend this route by also having all bindable datatypes implement ICoding, which will make the current extension method to get a coding out of a DataType unnecessary. Make sure the generated Resource.ICoding implementation is calling this interface, instead of the extension method. Remove the extension method.
- [ ] Remove ISystemAndCode, this is a breaking change.

ewoutkramer:
Fixes #2931.

Made the use of the existing `ICoded` interface more general: it now has been applied to all datatypes, and each bindeable datatype now knows how to transform itself into a list of codes.  This means the `ISystemAndCode` (only useful for Code<T>) and BindableAttribute have become superfluous, so they have been removed.


src/Hl7.Fhir.Base/Model/Generated/Attachment.cs
```diff
@@ -209,7 +209,7 @@ public string Url
         if (value == null)
           UrlElement = null;
         else
-          UrlElement = new Hl7.Fhir.Model.FhirUri(value);
+          UrlElement = new Hl7.Fhir.Model.FhirUrl(value);
```
Kasdejong, line 212: I remember seeing this before... are we sure this is correct? I believe so but at this point I dont trust it anymore :)

Kasdejong (approved):

src/Hl7.Fhir.Base/Model/Generated/Attachment.cs
mmsmits: This is an incorrect change I believe, this should be `FhirUri`, the more generic datatype.
At least according to this PR: https://github.com/FirelyTeam/firely-net-sdk/pull/2952/files

mmsmits (changes requested):

src/Hl7.Fhir.Base/Model/Generated/Attachment.cs
ewoutkramer: This was a side effect of me re-running generation. It was accidental, but if we don't do it here, it will come with any next PR that does regeneration again (I think that's Kas upcoming PR).

The change is correct, this property is specific to R4 (and higher), and it's a FHIR Url in those versions. It was an Uri in STU3.  I think I caught this mistake at the last moment, and never ran the generator again.

I think we should just bite the bullet, creating a separate PR for this change is maybe a bit much.

mmsmits (approved):

PR #3017, opened 2025-01-17
2936 Make comparisons consistent

### LINKED ISSUE #2932: Hide ToSystemType conversions behind an interface.
The following datatypes (to which Instant should be added) have methods to convert their values to a CQL type.

*Code
*CodeableConcept
*Code<T>
*Coding
*Date
*FhirDateTime
*Time
*NEW Instant

There is no way to polymorphically call these, which would be really useful in the implementation of `IScopedNode.Value`.  So, create an interface with a single `ToSystemType()` method, hide these methods behind an explicit implementation and adapt the implementation of IScopedNode.Value to use these methods instead.

### LINKED ISSUE #2936: What to do with equality/comparisons etc.
We have tons of comparison operations, duplicated between operations on the POCOs and on the CQL types.  Can we make up our mind and implement them on one or the other?

* We can convert any primitive FHIR type to the CQL type, and the CQL type have the comparison operations
* We could assume all IScopedNode trees end in a standard FHIR POCO datatype, and then we can call the operators on those types.
*

ewoutkramer:
This PR removes comparisons from the FHIR types, in favor of using the comparisons on the CQL types. Also, it simplifies the public surface of the CQL types:

* Remove comparison operators from FHIR Date/Time/DateTime/instant
* Replace Result<T> with normal TryXXXX pattern to align better with normal C# use (unfortunately).
* Add convenience methods + interface `IToSystemPrimitive` for consistently converting from a FHIR primitive to a CQL primitive.

@Kasdejong : The last feature will make sure that doing comparisons in FhirPath is a simple `IToSystemPrimitive.TryConvertToSystemType()` plus + `TryCompare()`. We should refactor that in the FP engine.

Still Todo:
- [x] I discovered we have not moved Ratio to Base (which we should) - this would also implement IToSystemPrimitive.
- [x] We now have TryToSystemXXX() and ToSystemXXX() on some types, and just ToSystemXXX() on others. Unify.
- [x] Some types have/had a nullable return for ToSystemXXXX(), which would return `null` if the Value of the primitive was null. I have now removed that, but double check that that makes sense. Esp with TryXXX: would you return false if you would return null if there is no value, or is that allowed?
- [x] Many of these operations don't have doccomments, need to add that.
- [x] Unit tests
- [x] Document breaking changes

Fixes #2936, also fixes #2932.


ewoutkramer:
On nullability of `ToSytemXXX()`, I think we should align with #2990, so throw if the value is null.  In the longer term this is also the proposed solution for other "illegal" values in the primitives.  In that case Value will throw, and probably so should all the `Toxxxx()` functions.  This means the datatype contains illegal/incomplete information, which is possible since the datatypes also participate in "overflow" (for now, using ObjectValue) and can contain illegal information. Rather than allowing a user to inadvertedly continue working with that information, we should throw when he/she tries to "do" something  with it.

src/Hl7.Fhir.Base/Model/Generated/XHtml.cs
```diff
@@ -82,8 +82,9 @@ protected internal override Base DeepCopyInternal()
       return instance;
     }
 
+    protected internal override SystemPrimitive.Any TryConvertToSystemTypeInternal() => null;
```
mmsmits: This should be in the non-generated partial class

src/Hl7.Fhir.Base/Model/Canonical.cs
```diff
@@ -27,130 +27,139 @@ POSSIBILITY OF SUCH DAMAGE.
...
+using P=Hl7.Fhir.ElementModel.Types;
 using Hl7.Fhir.Utility;
 using System;
 
 #nullable enable
 
-namespace Hl7.Fhir.Model
+namespace Hl7.Fhir.Model;
+
+public partial class Canonical: P.IToSystemPrimitive
```
mmsmits: Is this interface implemented by the base type?

src/Hl7.Fhir.Base/Model/Date.cs
```diff
@@ -35,7 +35,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace Hl7.Fhir.Model;
 
-public partial class Date
+public partial class Date: P.IToSystemPrimitive
```
mmsmits: Same?

src/Hl7.Fhir.Base/Model/FhirUrl.cs
```diff
@@ -30,32 +30,42 @@ POSSIBILITY OF SUCH DAMAGE.
 #nullable enable
 
 using System;
+using System.Diagnostics.CodeAnalysis;
+using P = Hl7.Fhir.ElementModel.Types;
 
-namespace Hl7.Fhir.Model
+namespace Hl7.Fhir.Model;
+
+public partial class FhirUrl: P.IToSystemPrimitive
```
mmsmits: Same?

src/Hl7.Fhir.Base/Model/Time.cs
```diff
@@ -37,7 +37,7 @@ POSSIBILITY OF SUCH DAMAGE.
 
 namespace Hl7.Fhir.Model;
 
-public partial class Time
+public partial class Time: P.IToSystemPrimitive
```
mmsmits: Same?

mmsmits (changes requested):

mmsmits (approved):

PR #3025, opened 2025-01-22
Attribute validation error message now provides valid FHIR definitional paths.

### LINKED ISSUE #3008: Attribute Validation error messages contain incorrect/incomplete location paths
**Describe the bug**
The location included in error messages are different for resources parsed from xml/json or created as objects
These should be consistent and use the fhir definitional paths (which are in the `FhirElement` attribute on the properties)
And if there is no positional information, then the line/position data should be excluded.
The current code returns a message such as `..., At Patient.Text.DivElement.Value, line , position`

https://github.com/FirelyTeam/firely-net-sdk/blob/c32024bfa360e3838c2c633d5b35bc0b4ec58b44/src/Hl7.Fhir.Base/Validation/CodedValidationException.cs#L90-L97

**To Reproduce**
Check this unit test:
```c#
 public void TestXhtmlValidation()
 {
     var p = new Patient
     {
         Text = new Narrative() { Div = "<div xmlns='http://www.w3.org/1999/xhtml'><p>should be valid</p></div>", Status = Narrative.NarrativeStatus.Generated }
     };
     DotNetAttributeValidation.Validate(p, true);

     p.Text.Div = "<div xmlns='http://www.w3.org/1999/xhtml'><p>should not be valid<p></div>";
     validateErrorOrFail(p, true);

     p.Text.Div = "<div xmlns='http://www.w3.org/1999/xhtml'><img onmouseover='bigImg(this)' src='smiley.gif' alt='Smiley' /></div>";
     validateErrorOrFail(p, true);
 }
```
The error message produced will have a location of `Patient.Text.DivElement.Value` where the actual location should be `Patient.text.div`
It also currently includes empty line/position info 
`"... At Patient.Text.DivElement.Value, line , position" `
this extra empty data should be excluded,

**Expected behavior**
The location should be accurate, and missing information should be absent from the message.

**Possible discussion**
Should these paths actually be different?

p.s. I've got code that resolves these issues, though I haven't checked if there are any other side effects as yet.

**Version used:**
 - FHIR Version: R4
 - Version: 5.11.1

brianpos:
## Description
Update the handling of attribute validation errors to use the same structure as the FHIR definitional paths.

## Related issues
Closes|Fixes|Resolves #3008 

## Testing
Unit tests updated for attribute validation to include the corrected location.

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Base/Introspection/FhirElementAttribute.cs
+++ b/src/Hl7.Fhir.Base/Introspection/FhirElementAttribute.cs
@@ -35,6 +35,7 @@ using System;
 using System.Collections.Generic;
 using System.ComponentModel.DataAnnotations;
 using System.Linq;
+using System.Reflection;
 
 #nullable enable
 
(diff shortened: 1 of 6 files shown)


brianpos:
This is due to the attribute supporting multiple. I'll update the code to handle that (and work out which of the multiple is in context at that point during processing)

https://github.com/FirelyTeam/firely-net-sdk/blob/631f01e226c82b19af04798e78434473dbaefd1b/src/Hl7.Fhir.Base/Model/Generated/OperationOutcome.cs#L331-L332

ewoutkramer:
We've redone this in SDK6.0, so I will close this PR. When this is necessary/wanted for SDK 5.0, that's ok, but since it has test errors (and is by now behind on SDK5), these will need to be fixed first.

PR #3033, opened 2025-02-07
2781 Work on ObjectValue

### LINKED ISSUE #2781: Streamline primitive's Value, ObjectValue and "original text"
The Primitive POCO's currently have an internal representation that is the same as the type of their `Value` property. They are listed in the [IReadOnlyDictionary](https://docs.fire.ly/projects/Firely-NET-SDK/en/latest/model/readonlydictionary.html) documentation.

POCO's also have an `ObjectValue` property, that should contain their internal representation (so the same as Value), but can also be used to store a verbatim string (from the json or xml) if it could not be parsed into the format required by the `Value` (i.e. an integer in Integer that was unparseable in the XML attribute).

For some operations on these primitives, the internal representation is not good enough, i.e. FhirDateTime needs to parse the string value in it to do comparisons. Conversely, some of the constructors of FhirDateTime pass in a discrete date time (already parsed), but since it stores a string internally, these are formatted to a string and then thrown away.

The current setup has a few disadvantages:
* There is unnecessary formatting/parsing going on for many types
* It is unclear what `ObjectValue` will contain, as its type is object and it can both contain a valid Value or text. I often have to refer back to our documentation to figure out what could be in ObjectValue.
* We do not keep the exact text of the source (json/xml, whatever) for roundtripping, we only do that for unparseable data. 

We should fix these disadvantages, maybe like so:

* Make ObjectValue a readonly property or add `ToObjectValue()` on `PrimitiveType`. 
* Have a "OriginalText" property that is set by the parser/raw text constructor and gets erased when its overwritten via its POCO interface. Ideally this would be read-only, and can be set by an explicit `Parse()` on the datatype, but this would require the new static interface members (Parse) on a IPrimitiveType or somesuch.
* Keep the `Value` property as is
* Add SetValue(string originalText, object parsed) - if we want to allow different parsers/sources to be able to supply unparsed/parsed values into the primitive, so it does not have to be reserialized if an underlying format already has a usable form.

ewoutkramer:
Fixes #2781.

This PR contains several improvements to how ObjectValue and validation are handled on primitives:

* ObjectValue now always aligns with the json literals, so the ObjectValue is now string for Instance (not DateTimeOffset), and string for Base64Binary (not byte[]). 
* It has delayed parsing from ObjectValue->Value and delayed serialization from Value->ObjectValue for all types.
* It unifies the validation of the ObjectValue across all primitives: all operations/properties on primitives that access Value will make sure the data is valid, and throw CodedValidationExceptions otherwise.
* Because the primitives now handle validation of ObjectValue, the parsers don't have to do it anymore, and they have been simplified to mostly just copy the raw value on the line to the ObjectValue, and let the new Primitive validation handle the rest.

⚠️ Make sure to also review and pull https://github.com/FirelyTeam/fhir-codegen/pull/52


src/Hl7.Fhir.Base/ElementModel/PocoElementNode.cs
```diff
@@ -195,7 +195,7 @@ internal object ToITypedElementValue()
                     Integer64 fint64 => fint64.Value,
                     PositiveInt pint => pint.Value,
                     UnsignedInt unsint => unsint.Value,
-                    Base64Binary { Value: { } b64 } => PrimitiveTypeConverter.ConvertTo<string>(b64),
+                    Base64Binary { ObjectValue: { } b64 } => b64,
```
Kasdejong, line 198: If we merge the FP engine refactoring PR before this one, it might be appropriate to do this for all CQL primitives in THIS PR instead of another one. Just a suggestion.

src/Hl7.Fhir.Base/Model/Base64Binary.cs
```diff
@@ -38,28 +43,100 @@ namespace Hl7.Fhir.Model;
...
-        new(Convert.FromBase64String(base64Data));
+    [FhirElement("value", IsPrimitiveValue = true, XmlSerialization = XmlRepresentation.XmlAttr, InSummary = true,
+        Order = 30)]
+    [DeclaredType(Type = typeof(P.String))]
+    [DataMember]
+    public byte[]? Value
+    {
+        get
+        {
+            if (ValidateObjectValue(null) is {} error)
```
Kasdejong, line 54: What is the purpose in having the side effect here and returning the error? could we not just have something like "ParseObjectValue" return the parsed value or throw? Is this prep for our error annotations??

ewoutkramer: That's right, in PrimitiveType, the implementation of the attribute validation ALSO calls this method (as does an additional helper method `HasValidValue()` on PrimitiveType.

src/Hl7.Fhir.Base/ElementModel/PocoElementNode.cs
ewoutkramer: This PR does it for all known mappings from a POCO to a ITypedElement value, as documented in ITypedElement. Which other types are you expecting?

ewoutkramer: But we can merge yours and see what happens.

src/Hl7.Fhir.Base/Model/Base64Binary.cs
ewoutkramer: But I have now added documentation that this function actually has a side effect in some classes, which is updating the internal cached value.

Kasdejong (approved):

Issue #2778, opened 2024-05-01
Better diagnostics when resolvers fail


ewoutkramer:
The `IResourceResolver` interface allows resolvers to return "null" when something "cannot be resolved". But the cause of something not being resolvable remains unclear. This has caused many hours of chasing down bugs where the resource could be resolved, but (most prominently) a StructureDefiniton could not be snapshotted. 

It would be way better if we could somehow return a resource OR an error. Of course, this is a massively breaking change. But smaller changes would mean the error can be easily ignored. Throwing exceptions also is not a great solution, still a lot of code is currently checking for null.

Since the most problematic of these is the SnapshotSource, a temporary "hack" to check SnapshotSource.Outcome after resolution MAY work, but its not great.

mmsmits:
Could also throw a specific exception.
Whatever we make up, it's breaking.

PR #3044, opened 2025-02-17
Enable resolvers to be able to report errors

andrzejskowronski:
## Description
Adds an `TryResolve` overloads to resolver interfaces, and obsoletes the regular ones.
Default interface implementation used to make it backwards compatible, while producing warnings still.

## Related issues
Closes [#2778](https://github.com/FirelyTeam/firely-net-sdk/issues/2778)

### DIFF
--- a/src/Hl7.Fhir.Base/Specification/Source/CachedResolver.cs
+++ b/src/Hl7.Fhir.Base/Specification/Source/CachedResolver.cs
@@ -75,7 +75,7 @@ namespace Hl7.Fhir.Specification.Source
         /// <remarks>Return data from memory cache if available, otherwise load on demand from the internal artifact source.</remarks>
         public async Task<Resource> ResolveByUriAsync(string url)
         {
-            var result = await TryResolveByUriAsync(url);
+            var result = await TryResolveByUriAsync(url).ConfigureAwait(false);
             return result.Value;
         }
 
@@ -91,7 +91,7 @@ namespace Hl7.Fhir.Specification.Source
         /// <remarks>Return data from memory cache if available, otherwise load on demand from the internal artifact source.</remarks>
         public async Task<Resource> ResolveByUriAsync(string url, CachedResolverLoadingStrategy strategy)
         {
-            var result = await TryResolveByUriAsync(url, strategy);
+            var result = await TryResolveByUriAsync(url, strategy).ConfigureAwait(false);
             return result.Value;
         }
         
@@ -120,7 +120,7 @@ namespace Hl7.Fhir.Specification.Source
         /// <remarks>Return data from memory cache if available, otherwise load on demand from the internal artifact source.</remarks>
         public async Task<Resource> ResolveByCanonicalUriAsync(string url)
         {
-            var result = await TryResolveByCanonicalUriAsync(url);
+            var result = await TryResolveByCanonicalUriAsync(url).ConfigureAwait(false);
             return result.Value;
         }
 
--- a/src/Hl7.Fhir.Base/Specification/Source/CachedResolver.cs
+++ b/src/Hl7.Fhir.Base/Specification/Source/CachedResolver.cs
@@ -98,7 +98,7 @@ namespace Hl7.Fhir.Specification.Source
         /// <summary>Retrieve the artifact with the specified url.</summary>
         /// <param name="url">The url of the target artifact.</param>
         /// <param name="strategy">Option flag to control the loading strategy.</param>
-        /// <returns>A <see cref="Resource"/> instance, or <c>null</c> if unavailable.</returns>
+        /// <returns>A <see cref="ResolverResult"/> instance, with either <see cref="ResolverResult.Value"/> or <see cref="ResolverResult.Error"/>.</returns>
         /// <remarks>Return data from memory cache if available, otherwise load on demand from the internal artifact source.</remarks>
         public async Task<ResolverResult> TryResolveByUriAsync(string url, CachedResolverLoadingStrategy strategy)
         {
@@ -110,9 +110,11 @@ namespace Hl7.Fhir.Specification.Source
         [Obsolete("CachedResolver now works best with asynchronous resolvers. Use TryResolveByCanonicalUriAsync() instead.")]
         public Resource ResolveByCanonicalUri(string url) => TryResolveByCanonicalUri(url).Value;
 
+        /// <inheritdoc cref="TryResolveByUriAsync(string)" />
         [Obsolete("CachedResolver now works best with asynchronous resolvers. Use TryResolveByUriAsync() instead.")]
         public ResolverResult TryResolveByUri(string uri) => TaskHelper.Await(() => TryResolveByUriAsync(uri));
 
+        /// <inheritdoc cref="TryResolveByCanonicalUriAsync(string)" />
         [Obsolete("CachedResolver now works best with asynchronous resolvers. Use TryResolveByCanonicalUriAsync() instead.")]
         public ResolverResult TryResolveByCanonicalUri(string uri) => TaskHelper.Await(() => TryResolveByCanonicalUriAsync(uri));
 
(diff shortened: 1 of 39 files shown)


src/Hl7.Fhir.Base/Specification/Source/CachedResolver.cs
```diff
@@ -90,25 +90,44 @@ public Resource ResolveByUri(string url, CachedResolverLoadingStrategy strategy)
...
             if (url == null) throw Error.ArgumentNull(nameof(url));
             return await _resourcesByUri.Get(url, strategy).ConfigureAwait(false);
         }
 
         /// <inheritdoc cref="ResolveByCanonicalUriAsync(string)" />
         [Obsolete("CachedResolver now works best with asynchronous resolvers. Use ResolveByCanonicalUriAsync() instead.")]
-        public Resource ResolveByCanonicalUri(string url) => TaskHelper.Await(() => ResolveByCanonicalUriAsync(url));
+        public Resource ResolveByCanonicalUri(string url) => TryResolveByCanonicalUri(url).Value;
+
+        public ResolverResult TryResolveByUri(string uri) => TaskHelper.Await(() => TryResolveByUriAsync(uri));
```
ewoutkramer, line 115: Shouldn't we mark these with the same Obsolete message as ResolveByCanonicalUri above?

src/Hl7.Fhir.Base/Specification/Source/IResourceResolver.cs
```diff
@@ -19,12 +19,43 @@ public interface IResourceResolver : ISyncOrAsyncResourceResolver
     {
         /// <summary>Find a resource based on its relative or absolute uri.</summary>
         /// <param name="uri">A resource uri.</param>
+        [Obsolete("This does not provide information about the kind of error that lead to us returning null. Consider implementing TryResolveByUri instead.")]
```
ewoutkramer: This message is geared to the implementers of this interface, but *users* of this interface will also see it, right?  So, should we not rephrase this for the end user?

src/Hl7.Fhir.Base/Specification/Source/IResourceResolver.cs
```diff
@@ -19,12 +19,43 @@ public interface IResourceResolver : ISyncOrAsyncResourceResolver
...
-
         /// <summary>Find a (conformance) resource based on its canonical uri.</summary>
         /// <param name="uri">The canonical url of a (conformance) resource.</param>
+        [Obsolete("This does not provide information about the kind of error that lead to us returning null. Consider implementing TryResolveByCanonicalUri instead.")]
         Resource ResolveByCanonicalUri(string uri);
+
+        /// <summary>Find a (conformance) resource based on its canonical uri.</summary>
+        /// <param name="uri">The canonical url of a (conformance) resource.</param>
+        /// <returns><see cref="ResolverResult"/> with an actual resource, or the <see cref="ResolverResult.Error"/>.</returns>
+        ResolverResult TryResolveByUri(string uri)
```
ewoutkramer, line 32: I think we should use the opportunity (=changing this file) to add nullability to the declarations.

src/Hl7.Fhir.Base/Specification/Source/MultiResolver.cs
```diff
@@ -73,53 +73,81 @@ public void Pop()
...
-        public Resource ResolveByUri(string uri) => TaskHelper.Await(() => ResolveByUriAsync(uri));
+        public Resource ResolveByUri(string uri) => TryResolveByUri(uri).Value;
 
         public async Task<Resource> ResolveByUriAsync(string uri)
+        {
+            var resource = await TryResolveByUriAsync(uri);
+            return resource.Value;
+        }
+
+        [Obsolete("MultiResolver now works best with asynchronous resolvers. Use ResolveByCanonicalUriAsync() instead.")]
```
ewoutkramer: ...or rather TryResolveByUriAsync!

src/Hl7.Fhir.Base/Specification/Source/IResourceResolver.cs
```diff
@@ -19,12 +19,43 @@ public interface IResourceResolver : ISyncOrAsyncResourceResolver
...
+            if (resource is not null) 
+                return resource;
+            
+            return ResolverException.NotFound();
+        }
+        
+        /// <summary>Find a resource based on its relative or absolute uri.</summary>
+        /// <param name="uri">A resource uri.</param>
+        /// <returns><see cref="ResolverResult"/> with an actual resource, or the <see cref="ResolverResult.Error"/>.</returns>
+        ResolverResult TryResolveByCanonicalUri(string uri)
```
ewoutkramer, line 47: Do we still need a non-async version of this?  Or is our codebase still so dependent on sync resolvers that that would be impractical?

src/Hl7.Fhir.Conformance/Specification/Source/CommonDirectorySource.cs
```diff
@@ -579,20 +579,42 @@ public IEnumerable<ArtifactSummary> ListSummaries()
...
+        public ResolverResult TryResolveByUri(string uri)
         {
             if (uri == null) throw Error.ArgumentNull(nameof(uri));
             var summary = GetSummaries().ResolveByUri(uri, _inspector);
-            return summary is not null ? loadResourceInternal<Resource>(summary) : null;
+
+            if(summary is null)
+                return ResolverException.ArtifactSummaryNoMatch(uri);
+            
+            var resource = loadResourceInternal<Resource>(summary);
```
ewoutkramer: This function will actually throw already, should we catch that and package in a ResolverResult?

src/Hl7.Fhir.STU3/Specification/Source/ResourceResolverExtensions.cs
```diff
@@ -38,7 +38,9 @@ public static StructureDefinition FindExtensionDefinition(this IResourceResolver
         /// <returns>Returns a StructureDefinition if it is resolvable and defines an extension, otherwise <c>null</c>.</returns>
         public static async Tasks.Task<StructureDefinition> FindExtensionDefinitionAsync(this IAsyncResourceResolver resolver, string uri)
         {
-            if (await resolver.ResolveByCanonicalUriAsync(uri).ConfigureAwait(false) is not StructureDefinition sd) return null;
+            var result = await resolver.TryResolveByCanonicalUriAsync(uri).ConfigureAwait(false);
```
ewoutkramer, line 41: For this PR really to have effect on the quality, we need to have TryXXXX versions of these too.  It's these functions that are used in the SDK itself (like in the Snapshot generator etc) - so these are the actual "users" of this change. I think it is wise to make a separate PR for that, as it will concentrate on things like: "how is the snapshot generator going to report back errors from its resolvers?".

src/Hl7.Fhir.Base/Specification/Source/ResolverException.cs
```diff
@@ -0,0 +1,55 @@
+using Hl7.Fhir.Model;
+using Hl7.Fhir.Utility;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Hl7.Fhir.Specification.Source;
+
+public class ResolverException : CodedException
```
ewoutkramer, line 13: Make sure you document public classes and members - at least the new ones in this PR that you have introduced!

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/Specification/Source/IResourceResolver.cs
andrzejskowronski: I would personally keep it for parity sake, since we want people to actually take advantage of the errors reported.
I would probably consider skipping that if we were to obsolete the sync interface completelly though.

src/Hl7.Fhir.STU3/Specification/Source/ResourceResolverExtensions.cs
ewoutkramer: I have created issues #3053 and #3054 to cover for this!

ewoutkramer (approved):

PR #3052, opened 2025-02-19
moved To... extension methods to be more consistent in where they can be found

### LINKED ISSUE #2992: Reorg ToXml/ToTypedElement/ToSourceNode files
These extension methods are in duplicate files in Base/STU3shims, their names and classes don't make sense any more too. Maybe put them in one place?

Kasdejong:
refactored extension methods around the elementmodel to now be in the appropriate classes

resolves #2992


src/Hl7.Fhir.Base/ElementModel/VersionedConversionExtensions.cs
```diff
@@ -0,0 +1,202 @@
+﻿/*
```
ewoutkramer: I think the filename should be `TypedElementExtensions.cs`, right?

src/Hl7.Fhir.Base/FhirPath/Functions/EqualityOperators.cs
```diff
@@ -48,7 +48,7 @@ public static class EqualityOperators
         // Note that the Equals as defined by FhirPath/CQL only returns empty when one or both of the arguments
         // are empty. Otherwise, it will return either false or true. Uncomparable values (i.e. datetimes
         // with incompatible precisions) are mapped to false, as are arguments of different types.
-        // This function is different from the one in the TypedElementExtensions, as that one does a structural
+        // This function is different from the one in the VersionedConversionExtensions, as that one does a structural
```
ewoutkramer: And this renamed too.

src/Hl7.Fhir.Base/Serialization/FhirJsonBuilderExtensions.cs
```diff
@@ -21,112 +21,13 @@ namespace Hl7.Fhir.Serialization;
...
-    public static byte[] ToJsonBytes(this ITypedElement source, bool pretty = false)
-        => SerializationUtil.WriteJsonToBytes(source.WriteTo, pretty);
-
-    /// <inheritdoc cref="ToJsonBytes"/>
-    [Obsolete("Async support will be removed in the next major release, please use the non-async version instead")]
-    public static async Task<byte[]> ToJsonBytesAsync(this ITypedElement source, bool pretty = false)
-        => await SerializationUtil.WriteJsonToBytesAsync(source.WriteToAsync, pretty).ConfigureAwait(false);
-
-    private static void writeTo(this JObject root, JsonWriter destination)
+    internal static void writeTo(this JObject root, JsonWriter destination)
```
ewoutkramer: Do we need a public class that just has internal methods?

src/Hl7.Fhir.Base/Serialization/FhirXmlBuilderExtensions.cs
```diff
@@ -7,136 +7,24 @@
...
 using System.Threading;
 using System.Threading.Tasks;
 using System.Xml;
 using System.Xml.Linq;
-using PocoNode = Hl7.Fhir.ElementModel.PocoNode;
 
 namespace Hl7.Fhir.Serialization;
 
-public static class FhirXmlBuilderExtensions
+public static class XDocumentExtensions
```
ewoutkramer: This public class seems to have internal methods only (in the diff).

ewoutkramer (changes requested):

ewoutkramer (changes requested):
And add the breaking changes to the wiki!  These will actually cause pain since if you used the SDK to work with multiple versions of FHIR in one solution, you will have used the names of the Extension classes, so its worth documenting.....

src/Hl7.Fhir.Base/ElementModel/VersionedConversionExtensions.cs
Kasdejong: yep, this was changed accidentally, ill correct it

src/Hl7.Fhir.Base/FhirPath/Functions/EqualityOperators.cs
Kasdejong: idem

PR #3070, opened 2025-03-04
3048 Move working with modes to parsers

### LINKED ISSUE #3049: Make FhirXml/JsonParser use the POCO-based parsers
We'll leave FhirXmlNode and FhirJsonNode untouched, but in the effort to make the POCO-based stack the default, this should happen asap.

This also helps for unifying the settings.

We should combine ParserSettings and [FhirXml/FhirJson]PocoDeserializerSettings

TODO after PR review:

- [x] Rename BaseFhirJson/XmlPocoDeserializer.cs to BaseFhirJson/XmlParser.cs
- [x] Rename ParserSettings.cs to DeserializerSettings.cs

ewoutkramer:
Fixes #3049.

Using with fixed desserialization modes was something we invented for FhirSerilizationEngines, but it turned out we really liked the concept to be useable for creating parsers in general and also for the FhirJsonConverter.  To make this possible, I did the following:

* Move the filtering logic to the json+xml deserializers
* Simplify the code in FhirJsonConverter for modes, since it does not require acrobatics with the engines anymore.
* Simplify the engines, since they can just rely on the parser's implementations of modes.

Since I originally started out working on #3049, there are some (minor) changes in this PR that actually are part of the future #3049.


src/Hl7.Fhir.Base/Serialization/BaseFhirJsonPocoDeserializer.cs
```diff
@@ -29,35 +28,15 @@ namespace Hl7.Fhir.Serialization;
 /// <remarks>The serializer uses the format documented in https://www.hl7.org/fhir/json.html. </remarks>
 public class BaseFhirJsonPocoDeserializer
 {
-    /// <summary>
-    /// Initializes an instance of the deserializer.
-    /// </summary>
-    /// <param name="assembly">Assembly containing the POCO classes to be used for deserialization.</param>
-    public BaseFhirJsonPocoDeserializer(Assembly assembly) : this(assembly, new())
```
Kasdejong, line 36: is there a reason we do not keep this constructor around?

Kasdejong (commented):
Set to approve, just worried about how many people might use the assembly to construct their parsers. It is public, correct?

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonPocoDeserializer.cs
ewoutkramer: All other BaseXXXXX (BaseDirectorySource, BaseFhirSerializer, Base etc) have a ModelInspector. But you could be right, I'll add it back, and I'll obsolete it first.

Kasdejong (approved):

Issue #3056, opened 2025-02-21
Implement overflow according to the design


ewoutkramer:
Based on the design (#2908 ),  now implement overflow.

Based on the design this mostly means:

* Regeneraring SetValue to handle "incorrect" values
* Regenerating the setters to also update overflow
* Making sure that getters throw on incorrect values
* Change/generate ValidateObject (on Base I think) so that it will go through the overflow and report the encountered errors.

But see the design for more details ;-)

Issue #2982, opened 2024-11-27
Make attribute validation sensitive to FHIR version


ewoutkramer:
So far, .NET attribute validation is not aware of the `Since` attribute. This was never a problem, as all attributes that have `Since` do not take part in validation, they just provide metadata.

This needs to change (and was already not completely true):
* `NotMapped` should validate whether an element still appears in a FHIR version and then report an error.
* `FhirElement` can repeat and should validate whether an element is valid for a given FHIR version (not validated at this moment)
* `AllowedTypes` has been promoted to a versioned attribute, since the set of allowed types can vary per version.

We should probably extend the `ValidationContext` with a property (as we've done before, the pattern exists) that represents the applicable FHIR version (or highest if let out).

mmsmits:
Consider putting ModelInspector on Base, so the property validator can figure out which properties to use.

PR #3079, opened 2025-03-13
Allow the model to store invalid data and give detailed exceptions about it

Kasdejong:
## Description
- Allows the model to store invalid data in the overflow and report errors on it when the matching getter is called
- Also added exceptions for when invalid data is found in the corresponding overflow entry when the property is being requested

## Related issues
implements #3056 

Codegen PR: FirelyTeam/fhir-codegen#55


ewoutkramer:
```
      /// <summary>
      /// Defines how the part works
      /// </summary>
      /// <remarks>This uses the native .NET datatype, rather than the FHIR equivalent</remarks>
      [IgnoreDataMember]
      public string? Definition
      {
        get => _DefinitionElement?.Value;
        set
        {
          DefinitionElement = value is null ? null : new Hl7.Fhir.Model.Canonical(value);
          OnPropertyChanged("Definition");
        }
      }
```

The helper properties are exposing the null markers, and won't trigger an exception. Maybe let the getters just use the normal property instead of the backing field, that should solve the problem

ewoutkramer:
We've hit a subtle snag here, which is characteristic of everything we've with moving over to the POCOs:  Since POCO is King, then so is the model, defined by the attributes on the POCOs. However, we don't always stick to it for now - mostly because of the  "Since" parameters on the attributes. This means that for SOME versions of FHIR a datatype is different, or an element did not yet exist. We're not using this information at all in the new model validation. We did in the parsers (they are using the right version of the ModelInspector), but our (attribute) validators don't account for it yet.  Also, when we return the type of the element in the PocoNode, we return the physical type, which is not always the FHIR type (this is, I think, only relevant for the `ofType` in FhirPath).

The good news is: we have the information. Since we're dealing with POCOs and this is encoded on the POCOs. However, we've shied away from using metadata most of the times (although we annotated ModelInspector sometimes I think), and the ValidationContext has no notion of FHIR version yet. 

So, I have moved #2982 on this sprint, since I think we should do _something_ with the FHIR version while validating the model:
* While parsing, this is simple, the parser has the ModelInpector, and can pass it to the model validation
* While calling "attribute validation" outside of the parsers, we'lll probably have to pass in the ModelInspector to make the attributes/IValidateObject sensitive to FHIR version.

A nice last design discussion!

src/Hl7.Fhir.Base/Model/Base.cs
```diff
@@ -46,8 +47,10 @@ public abstract partial class Base : IAnnotatable, INotifyPropertyChanged
     /// FHIR Type Name
     /// </summary>
     public virtual string TypeName => GetType().Name;
-
+    
     private Dictionary<string, object>? _overflow = null;
+
```
ewoutkramer, line 52: Also known as `_overflow?.Any() == true`.

src/Hl7.Fhir.Base/Model/Generated/BackboneElement.cs
```diff
@@ -107,10 +125,16 @@ public override bool TryGetValue(string key, [NotNullWhen(true)] out object? val
 
     public override Base SetValue(string key, object? value)
     {
+      if(value is not (null or Hl7.Fhir.Model.Base or IEnumerable<Hl7.Fhir.Model.Base>)) throw new ArgumentException("Value must be a Base or IEnumerable<Base>", nameof(value));
       switch (key)
       {
         case "modifierExtension":
-          ModifierExtension = (List<Hl7.Fhir.Model.Extension>?)value!;
+          if (value is not (List<Hl7.Fhir.Model.Extension> or null))
```
ewoutkramer, line 133: If something implements IEnumerable, but is not a list, it will end up in the overflow. Is that correct?  Orioginally I had some code to construct a list based on an IEnumerable (If it is a list, then assign directly, else construct a new list with the ienumerable contents). Would that make sense?

src/Hl7.Fhir.Base/Serialization/FhirJsonException.cs
```diff
@@ -28,24 +28,24 @@ namespace Hl7.Fhir.Serialization;
 public class FhirJsonException : ExtendedCodedException
 {
     public const string EXPECTED_START_OF_OBJECT_CODE = "JSON101";
-    public const string RESOURCETYPE_SHOULD_BE_STRING_CODE = "JSON102";
-    public const string NO_RESOURCETYPE_PROPERTY_CODE = "JSON103";
+    public const string RESOURCETYPE_SHOULD_BE_STRING_CODE = "JSON102"; // non-duplicate
```
ewoutkramer, line 31: What does "non-duplicate" mean?

src/Hl7.Fhir.Base/Validation/CodedValidationException.cs
```diff
@@ -40,6 +41,13 @@ public class CodedValidationException : ExtendedCodedException
     public const string LITERAL_INVALID_CODE = "PVAL124";
     public const string POSITIVE_INT_MUST_BE_POSITIVE_CODE = "PVAL125";
     public const string UNSIGNED_INT_MUST_NOT_BE_NEGATIVE_CODE = "PVAL126";
+    public const string EXPECTED_OBJECT_NOT_PRIMITIVE_CODE = "PVAL127";
```
ewoutkramer, line 44: I wonder whether we need different error codes here, since they are all basically type mismatches. Another problem is that the terminology is Json specific. Or maybe this means we should have different codes for it, so the parsers can override the text with the correct term for the serialization format....

src/Hl7.Fhir.Base/Validation/CodedValidationException.cs
```diff
@@ -40,6 +41,13 @@ public class CodedValidationException : ExtendedCodedException
...
+    public const string EXPECTED_OBJECT_NOT_ARRAY_CODE = "PVAL128";
```
ewoutkramer, line 45: Maybe we should also check whether the distinction between recoverable/non-recoverable is still correct in the json and xml error list.

src/Hl7.Fhir.Support.Poco.Tests/Model/PocoDictionaryTests.cs
```diff
@@ -45,13 +46,13 @@ public void ResourceAcceptsOverflow()
         var pat = new Patient();
 
         // setting an existing property to an incorrect type should fail.
```
ewoutkramer: Setting an existing property to a non-Base type should fail.

src/Hl7.Fhir.Base/Validation/CodedValidationException.cs
Kasdejong: I believe the goal here is to be FHIR-specific, not XML or JSON specific. Object/Primitive/Array should probably be renamed to their FHIR equivalents, I will adjust this

src/Hl7.Fhir.Base/Model/Generated/BackboneElement.cs
Kasdejong: note: we require list here

ewoutkramer (approved):

Issue #2878, opened 2024-10-01
Add FhirSerializationEngine.Validation


ewoutkramer:
Currently, the parser is catching errors that would more naturally fit to be caught by the validator, i.e. cadinality errors.  Or errors about parsing enums (which are actually terminology errors).  We can introduce a new validator mode that lets these errors through, so the validator can catch them.  We could call this mode `Validation`

alexzautke:
This will be done by the parsers themselves not by FhirSerializationEngine. See https://github.com/FirelyTeam/firely-net-sdk/issues/3048

ewoutkramer:
I tried to do it as part of restructuring FhirSerializationEngine, but it's too early to add this, as we first need to find out which errors will be moved from the parsers to the POCO attribute validation. After having done this, we should actually go over all error categories and make sure they are still up to date.

ewoutkramer:
The tests in src/Hl7.Fhir.Support.Poco.Tests/NewPocoSerializers/FhirSerializationEngineTests.cs have become almost useless after the work on overflow and parsers, so we will need to review them to find differences between the modes again.

PR #3080, opened 2025-03-17
Fallback to dynamic/guessed types on unknown data

### LINKED ISSUE #3050: Allow the parsers to fill POCOs with unknown/incorrect data
This also requires handling erroneous data in such a way that the errors are accurate when fetching the incorrectly filled properties.

Quick list of errors that SHOULD be handled by base (not the parsers) for reference:

- Unknown elements
- Expected primitive, got object or vice versa
- Expected single, got array or vice versa
- choice where no choice allowed
- unknown choice
- unknown resource

andrzejskowronski:
## Description
Targets poco branch for ease of viewing just the relevant changes, and will require the other pr.

## Related issues
Closes #3050


src/Hl7.Fhir.Base/Serialization/BaseFhirJsonPocoDeserializer.cs
```diff
@@ -849,11 +1023,16 @@ private static (PropertyMapping? propMapping, ClassMapping? propValueMapping, Fh
...
-            return string.IsNullOrEmpty(typeSuffix)
-                ? (null, ERR.CHOICE_ELEMENT_HAS_NO_TYPE(ref r, path.GetInstancePath(), propertyMapping.Name))
-                : inspector.FindClassMapping(typeSuffix) is { } cm
-                    ? (cm, null)
-                    : (null, ERR.CHOICE_ELEMENT_HAS_UNKOWN_TYPE(ref r, path.GetInstancePath(), propertyMapping.Name, typeSuffix));
+            ClassMapping? choiceMapping = null;
+            if(!string.IsNullOrEmpty(typeSuffix))
+                choiceMapping = inspector.FindClassMapping(typeSuffix);
+            
+            choiceMapping ??= inspector.FindClassMapping(nameof(DynamicDataType));
```
ewoutkramer: But this could also be a DynamicPrimitive - which we could heuristically determine from the fact that there is a `_` or a primitive value. But we don't know at this point, so maybe we need a method ToDynamicPrimitive() that switches to a primitive if we discover that later on.

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonPocoDeserializer.cs
```diff
@@ -849,11 +1023,16 @@ private static (PropertyMapping? propMapping, ClassMapping? propValueMapping, Fh
...
+            
+            if(choiceMapping is not null)
+                return (choiceMapping, null);
+            
+            return (null, ERR.CHOICE_ELEMENT_HAS_UNKOWN_TYPE(ref r, path.GetInstancePath(), propertyMapping.Name, typeSuffix));
```
ewoutkramer: This should never happen.  In fact, the NewPocoBuilder has constants for the classmappings for  DynamicDataType, and this will never be null when we use those.

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonPocoDeserializer.cs
```diff
@@ -236,8 +241,10 @@ private void deserializeObjectInto<T>(
     {
         if (reader.TokenType != JsonTokenType.StartObject)
```
ewoutkramer: I think we should be more flexible in our thinking: we will only call this function IF we are at an object. I know at this moment we are doing it if we are expecting an object (because of classmappings etc), but since those are just hints now, we should probably never end up in this function if we don't actually see an object on the wire.

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonPocoDeserializer.cs
```diff
@@ -310,6 +317,142 @@ private void deserializeObjectInto<T>(
...
+                
+                var choiceMapping = _inspector.FindClassMapping(subSpan.ToString());
+                if (choiceMapping is not null)
+                    return (span[..i].ToString(), choiceMapping);
+            }
+            return (propertyName, null);
+        }
+    }
+
+    private object? deserializeUnexpectedJsonValue(string propertyName, ref Utf8JsonReader reader, FhirJsonPocoDeserializerState state, bool stayOnLastToken, ClassMapping? propertySuggestion = null)
```
ewoutkramer: Nice, I think this is going to be the main workhorse, in fact we could call this straight from the main TryDeserializeResouirce and TryDeserializeObject functions - we're only *expecting* something, but we're no longer bound to it, so this dynamic function is really nice.

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonPocoDeserializer.cs
```diff
@@ -342,7 +485,7 @@ FhirJsonPocoDeserializerState state
 
         // There might be an existing value, since FhirPrimitives may be spread out over two properties
         // (one with, and one without the '_')
-        var existingValue = propertyMapping.GetValue(target);
+        target.TryGetValue(propertyMapping.Name, out var existingValue);
```
ewoutkramer: Note that we've moved from using IL (in GetValue) to our generated function now, which might be a bit slower.

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonPocoDeserializer.cs
```diff
@@ -335,6 +342,142 @@ private void deserializeObjectInto<T>(
...
+
+        // move past property name
+        reader.Read();
+
+        object? result;
+        if(choiceType is not null)
+        {
+            state.Path.EnterElement(name, null, choiceType.IsPrimitive);
+            
+            result = DeserializeFhirPrimitive(value as PrimitiveType, name, choiceType, null, ref reader, new(), state);
```
ewoutkramer: This does not need to be a primitive, right? Choice types can be non-primitives too,

ewoutkramer: What if it is a list?

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonPocoDeserializer.cs
```diff
@@ -305,7 +312,7 @@ private void deserializeObjectInto<T>(
 
                 try
                 {
-                    state.Path.EnterElement(propMapping!.Name, !propMapping.IsCollection ? null : 0, propMapping.IsPrimitive);
+                    state.Path.EnterElement(propMapping!.Name, isEnteringJsonArray(ref reader) ? 0 : null, propMapping.IsPrimitive);
                     deserializePropertyValueInto(target, currentPropertyName, propMapping, propValueMapping!, ref reader, objectParsingState, state);
```
ewoutkramer: Maybe we should try to merge deserializeUnknownPropertiesInto and deserializePropertyValueInto into a single function? (it needs to do the same thing, with "the only difference" that we don't have type information)

ewoutkramer: Maybe by making sure we always have metadata, but then the metadata for DynamicPrimitive/DynamicType/DynaimcResource etc.

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/Model/Base.Extensions.cs
```diff
@@ -65,4 +65,19 @@ internal static void CopyToInternal(this Dictionary<string, object> source, Dict
             };
         }
     }
+
+    internal static DynamicPrimitive ToDynamicPrimitive(this Base instance)
+    {
+        var primitive = new DynamicPrimitive();
```
ewoutkramer, line 71: Dont forget to set the name of the type! Maybe we should remove the parameterless constructor so we are forced to think about it (even if passing null as the name is fine, at least it is explicit).

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonSerializer.cs
```diff
@@ -242,6 +242,14 @@ private void serializeFhirPrimitive(string elementName, PrimitiveType value, Utf
             writer.WritePropertyName(elementName);
             SerializePrimitiveValue(value.ObjectValue, writer, requiredType);
         }
+        // Since the object is not null, try overflow, but only if no extension data defined
```
ewoutkramer, line 245: I dont really understand whats going on, but it looks like we're serializing the contents of the overflow into the primitive element. This is invalid fhir json, so I'd like to understand better which problem we are solving here. It seems the overflow should be put under the _element?

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
```diff
@@ -838,11 +943,16 @@ private static (PropertyMapping? propMapping, ClassMapping? propValueMapping, Fh
...
-            return string.IsNullOrEmpty(typeSuffix)
-                ? (null, ERR.CHOICE_ELEMENT_HAS_NO_TYPE(ref r, path.GetInstancePath(), propertyMapping.Name))
-                : inspector.FindClassMapping(typeSuffix) is { } cm
-                    ? (cm, null)
-                    : (null, ERR.CHOICE_ELEMENT_HAS_UNKOWN_TYPE(ref r, path.GetInstancePath(), propertyMapping.Name, typeSuffix));
+            ClassMapping? choiceMapping = null;
+            if(!string.IsNullOrEmpty(typeSuffix))
+                choiceMapping = inspector.FindClassMapping(typeSuffix);
+            
+            choiceMapping ??= inspector.FindClassMapping(nameof(DynamicDataType));
```
ewoutkramer, line 950: We should introduce a non-null ClassMapping.DynamicDataType property (and use that in the NewPocoBuilder too), so the last line of this function (which will and should never be reached) can be removed.

src/Hl7.Fhir.Base/Serialization/BaseFhirXmlDeserializer.cs
```diff
@@ -597,13 +731,18 @@ private static (PropertyMapping? propMapping, ClassMapping? propValueMapping, Fh
...
-            return string.IsNullOrEmpty(typeSuffix)
-                ? (null, ERR.CHOICE_ELEMENT_HAS_NO_TYPE(r, path.GetInstancePath(), propertyMapping.Name))
-                : inspector.FindClassMapping(typeSuffix) is { } cm
-                    ? (cm, null)
-                    : (null, ERR.CHOICE_ELEMENT_HAS_UNKOWN_TYPE(r, path.GetInstancePath(), propertyMapping.Name, typeSuffix));
+            
+            if(!string.IsNullOrEmpty(typeSuffix))
+                choiceMapping = inspector.FindClassMapping(typeSuffix);
+
+            choiceMapping ??= inspector.FindClassMapping(nameof(DynamicDataType));
```
ewoutkramer, line 740: We should introduce a non-null ClassMapping.DynamicDataType property (and use that in the NewPocoBuilder too), so the last line of this function (which will and should never be reached) can be removed.

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
```diff
@@ -408,8 +351,154 @@ FhirJsonPocoDeserializerState state
...
+            propertyValueMapping ??= _inspector.FindClassMapping(nameof(DynamicDataType))!;         
+            
+            if(existingValue is Base)
+                state.Errors.Add(ERR.DUPLICATE_PROPERTY(ref reader, state.Path.GetInstancePath(), propertyName));
+            
+            result = deserializeSingleValue(ref reader, propertyValueMapping, state);
+        }
+        else if (reader.TokenType == JsonTokenType.StartArray)
+        {
+            var peeked = peekTypeNested(ref reader);
```
ewoutkramer, line 409: Is there a way to avoid these peeks (since they do cost a bit of time) for the case where we do have a mapping?

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
```diff
@@ -548,19 +636,24 @@ FhirJsonPocoDeserializerState state
                 // don't read any new data into the primitive here
                 reader.Read();
             }
+            else if(reader.TokenType == JsonTokenType.StartArray)
```
ewoutkramer, line 639: We should report an error here about nested lists not being allowed in FHIR.

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
```diff
@@ -513,7 +601,7 @@ FhirJsonPocoDeserializerState state
         if (reader.TokenType != JsonTokenType.StartArray)
         {
             state.Errors.Add(ERR.EXPECTED_START_OF_ARRAY(ref reader, state.Path.GetInstancePath()));
-            oneshot = true;
+            return existingList;
```
ewoutkramer, line 604: Remove "oneshot" logic from deserializeFhirPrimitiveList as well.

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
```diff
@@ -127,38 +128,40 @@ public bool TryDeserializeObject(Type targetType, ref Utf8JsonReader reader, [No
     {
         if (reader.TokenType != JsonTokenType.StartObject)
         {
-            state.Errors.Add(ERR.EXPECTED_START_OF_OBJECT(ref reader, state.Path.GetInstancePath(), reader.TokenType));
-            reader.Recover();  // skip to the end of the construct encountered (value or array)
-            return null;
+            var target = (_inspector.FindClassMapping(typeof(DynamicResource))?.Factory() as Resource)!;
```
ewoutkramer, line 131: I think we should just throw.

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
```diff
@@ -225,13 +228,12 @@ private void deserializeObjectInto<T>(
         bool stayOnLastToken = false) where T : Base
     {
         if (reader.TokenType != JsonTokenType.StartObject)
-        {
-            state.Errors.Add(ERR.EXPECTED_START_OF_OBJECT(ref reader, state.Path.GetInstancePath(), reader.TokenType));
-            reader.Recover();  // skip to the end of the construct encountered (value or array)
+        {            
+            deserializePropertyInto(target, "value", ref reader, state, stayOnLastToken, new());
```
ewoutkramer, line 232: derserializeObjectInto is called from several sites, and I think this particular fix for encountering a non-object would not work for all of them. E.g. when this is called in the situation where we are handling the "complex"/extension part of a primitive, it might logically override the original ObjectValue.  In some cases where deserializeObjectInto is called, it cannot be anything else than an object (because we have already checked for it).  So maybe we should make sure that all callers have verified and have handled the situation where this is not an object, and then throw here if it still is (and then it is a programmer error).

src/Hl7.Fhir.Serialization.Shared.Tests/SerializationExcexptionHandlersXmlPoco.cs
```diff
@@ -313,10 +482,6 @@ public void XMLInvalidPropertyOrdering()
                 DebugDump.OutputXml(oc);
                 DebugDump.OutputXml(ex.PartialResult);
 
-                Assert.AreEqual("Patient.active", oc.Issue[0].Expression.First());
```
ewoutkramer, line 316: Why is this gone?

src/Hl7.Fhir.Serialization.Shared.Tests/SerializationExcexptionHandlersXmlPoco.cs
```diff
@@ -358,28 +518,17 @@ public void XMLInvalidPropertyDetected()
                 DebugDump.OutputXml(oc);
                 DebugDump.OutputXml(ex.PartialResult);
 
-                Assert.AreEqual("Doe2", (ex.PartialResult as Patient).Name[1].Family);
-
-                Assert.AreEqual("Patient.name[1]", oc.Issue[0].Expression.First());
```
ewoutkramer, line 363: Are we not detecting this anymore, or will we again when we have merged Kas work?

src/Hl7.Fhir.Shared.Tests/Serialization/ResourceParsingTests.cs
```diff
@@ -28,7 +28,7 @@ public partial class ResourceParsingTests
         [TestMethod]
         public void ConfigureFailOnUnknownMember()
         {
-            const string xml = "<Patient xmlns='http://hl7.org/fhir'><gender value='ox'/><daytona></daytona></Patient>";
+            const string xml = "<Patient xmlns='http://hl7.org/fhir'><gender value='ox'/><daytona value='test'></daytona></Patient>";
```
ewoutkramer, line 31: What happened in the original case, without the value?

src/Hl7.Fhir.Support.Poco.Tests/NewPocoSerializers/FhirJsonDeserializationTests.cs
```diff
@@ -33,8 +33,8 @@ public void PrimitiveValueCannotBeComplex()
 
     [DataTestMethod]
     [DataRow("OperationOutcome", null)]
-    [DataRow("OperationOutcomeX", ERR.UNKNOWN_RESOURCE_TYPE_CODE)]
-    [DataRow("Meta", null)]
+    [DataRow("OperationOutcomeX", ERR.RESOURCE_TYPE_NOT_A_RESOURCE_CODE)]
```
ewoutkramer, line 36: Why isn't this unknown resource type anymore?

src/Hl7.Fhir.Support.Poco.Tests/NewPocoSerializers/FhirJsonDeserializationTests.cs
```diff
@@ -200,17 +201,16 @@ public static IEnumerable<object[]> TestDeserializeResourceData
             yield return
             [
                 new { resourceType = "Doesnotexist", crap = 5 }, JsonTokenType.EndObject,
-                ERR.UNKNOWN_RESOURCE_TYPE_CODE
             ];
             yield return
             [
                 new { resourceType = nameof(OperationOutcome), crap = 5 }, JsonTokenType.EndObject,
-                ERR.UNKNOWN_PROPERTY_FOUND_CODE
+                COVE.INCORRECT_CARDINALITY_MIN_CODE
```
ewoutkramer, line 208: This will be different again when we integrate Kas's changes.

src/Hl7.Fhir.Support.Poco.Tests/NewPocoSerializers/FhirJsonDeserializationTests.cs
```diff
@@ -285,40 +285,37 @@ public void TestData(Type t, object testObject, JsonTokenType token, Action<obje
...
             ERR.ARRAYS_CANNOT_BE_EMPTY_CODE);
         yield return data<ContactDetail>(
             new { name = "Ewout", telecom = new object[] { new { system = "phone" }, new { systemX = "b" } } },
-            checkData, ERR.UNKNOWN_PROPERTY_FOUND_CODE);
+            checkData);
         yield return data<ContactDetail>(
             new { name = "Ewout", _telecom = new object[] { new { system = "phone" }, new { systemX = "b" } } },
-            checkData, ERR.USE_OF_UNDERSCORE_ILLEGAL_CODE, ERR.UNKNOWN_PROPERTY_FOUND_CODE);
-        yield return data<ContactDetail>(new { name = new[] { "Ewout" } }, ERR.EXPECTED_PRIMITIVE_NOT_ARRAY_CODE);
+            checkData);
```
ewoutkramer, line 317: Why is this use of underscore illegal gone?

src/Hl7.Fhir.Support.Poco.Tests/NewPocoSerializers/FhirJsonDeserializationTests.cs
```diff
@@ -511,26 +507,32 @@ public void TestRecovery()
 
             assertErrors(dfe.Exceptions, [
                 COVE.LITERAL_INVALID_CODE,
```
ewoutkramer, line 509: I think we should review this test after Kas's work has been merged, then we can verify the original tests versus the completed one with both your PRs

src/Hl7.Fhir.Support.Poco.Tests/NewPocoSerializers/FhirSerializationEngineTests.cs
```diff
@@ -56,12 +56,12 @@ public void TestParsingEM(string data, string? permissive, string? strict, strin
     [TestMethod]
     [DataRow(CORRECTXML, null, null, null, DisplayName = "Correct XML")]
     [DataRow(PERMISSIVEXML, null, POCO_EMPTY_VALUE, POCO_EMPTY_VALUE, DisplayName = "Permissive XML")]
-    [DataRow(BWCOMPATIBLEXML, POCO_UNKNOWN_ELEMENT, POCO_UNKNOWN_ELEMENT, null, DisplayName = "Backwards-compatible XML")]
-    [DataRow(POCO_WRONGXML, POCO_EXPECTED_OBJECT, POCO_EXPECTED_OBJECT, POCO_EXPECTED_OBJECT, DisplayName = "Wrong XML")]
+    [DataRow(BWCOMPATIBLEXML, null, null, null, DisplayName = "Backwards-compatible XML")]
```
ewoutkramer, line 59: This isn't testing much anymore, it was supposed to test the differences between CORREECt/permissive/bw compat etc, but it no longer functions as a good test, we will need to write another one. I will take this up as part of https://github.com/FirelyTeam/firely-net-sdk/issues/2878

src/Hl7.Fhir.Support.Poco.Tests/TestData/fp-test-patient-errors-recovered.json
```diff
@@ -1,16 +1,87 @@
 {
   "resourceType": "Patient",
```
ewoutkramer, line 2: We should manually review the new output (recovered + dumped errors) to verify this after pulling the PR.

ewoutkramer (changes requested):

ewoutkramer (changes requested):
We're working on this already....

src/Hl7.Fhir.Shared.Tests/Serialization/ResourceParsingTests.cs
andrzejskowronski: It would previously stop processing when it encounters unknown element, now parses it. 
When it would now process it, it will complain about object not having content or value.

src/Hl7.Fhir.Support.Poco.Tests/NewPocoSerializers/FhirJsonDeserializationTests.cs
andrzejskowronski: Fixed that one :)

src/Hl7.Fhir.Support.Poco.Tests/NewPocoSerializers/FhirJsonDeserializationTests.cs
andrzejskowronski: Fixed!

ewoutkramer:
Superceded by #3094.

PR #3098, opened 2025-04-09
Allow Resolve() to resolve versioned references

### LINKED ISSUE #3097: Fix in sdk-6
See parent issue for more details. This issue concerns the SDK-6 specific fix.

Kasdejong:
## Description
Changed ResourceIdentity and resolve implementation on PocoNode to now correctly resolve versioned references. Note that this has not been fixed on ScopedNode. It will be, once we pull SDK-5 to SDK-6 again.

## Related issues
fixes #3097 

## Testing
new unit test for testing this behaviour, as well as existing unit test framework


src/Hl7.Fhir.Base/ElementModel/PocoNodeExtensions.cs
```diff
@@ -277,6 +279,23 @@ internal static IEnumerable<PocoNode> Where<T>(this PocoListNode pln, Func<T, bo
...
+    /// </summary>
+    /// <param name="node"></param>
+    /// <param name="path"></param>
+    /// <returns></returns>
+    public static IEnumerable<PocoNode> NavigateTo(this PocoNode node, string path)
+    {
+        var parts = path.Split(['.', '[', ']'], StringSplitOptions.RemoveEmptyEntries);
+
+        return parts.Aggregate<string, IEnumerable<PocoNode>>(node, (current, part) =>
+            Regex.IsMatch(part, @"^\d+$")
```
ewoutkramer: Is `Int32.TryParse()` not a better/faster alternative for this check? You'll have the integer ready when you need it (now you need an expensive regex + parse).

ewoutkramer: I would try to return IReadOnlyCollection or at least turn the enumeration into a list, since there's a good chance the result of NavigateTo will be re-run by the caller, which is quite expensive.

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/ElementModel/PocoNodeExtensions.cs
Kasdejong: Agreed with the first one, but for the second one: should the caller not decide this for themselves?

ewoutkramer: It was a constant source of pain in ITypedElement, so I'd rather pre-compute it, yes.

Kasdejong: Decided: we leave it like this for now to maintain consistent with other PocoNode methods, we investigate enumerating to IReadOnlyCollections later. See also #3114

ewoutkramer (approved):

PR #3099, opened 2025-04-09
Fixed behaviour of versioned resources in bundles

### LINKED ISSUE #3096: Fix in sdk-5
See parent issue for more details. This issue concerns the SDK-5 specific fix for the parent issue

Kasdejong:
## Description
- Fixes duplicate key exception when trying to index a bundle that contains resources with the same fullUrl
- Allows versioned references to reference versioned resources in bundles

## Related issues
Resolves #3096 

## Testing
One new unit test for this behaviour, as well as the existing resolve framework

### DIFF
--- a/src/Hl7.Fhir.Base/ElementModel/ScopedNode.cs
+++ b/src/Hl7.Fhir.Base/ElementModel/ScopedNode.cs
@@ -219,7 +219,7 @@ namespace Hl7.Fhir.ElementModel
             
             if (InstanceType == "Bundle")
             {
-                var referenceEntryPairs = new List<KeyValuePair<string, ScopedNode>>();
+                var referenceEntryPairs = new List<KeyValuePair<string?, ScopedNode>>();
                 var versionedEntries = Current.Children("entry").Where(entry => entry.Children("resource").Children("meta").Children("versionId").Any());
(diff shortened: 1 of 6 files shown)


src/Hl7.Fhir.Base/ElementModel/ScopedNode.cs
```diff
@@ -211,16 +213,31 @@ public BundledResource(string? fullUrl, ScopedNode? resource) : this()
         /// <summary>
         /// When this node is the root of a Bundle, retrieves the bundled resources in its Bundle.entry.
         /// </summary>
-        public IEnumerable<BundledResource> BundledResources()
+        internal IEnumerable<BundledResource> BundledResources()
```
ewoutkramer: That's a breaking change, not possible in 5.x, unless it is fundamentally broken.

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/ElementModel/ScopedNode.cs
Kasdejong: It is fundamentally broken technically because of our unversioned resource trick, but we can keep it public for now if anyone uses it. It is quite a new field I believe.

ewoutkramer (dismissed):

ewoutkramer:
@Kasdejong - not building - suppressions?

ewoutkramer (approved):

PR #3112, opened 2025-04-23
Be careful not to accidentally create an empty overflow dictionary. M…

### LINKED ISSUE #3111: HasOverflow also returns true when overflow is empty.
We are just checking whether a dictionary has been created, and return true even if the dictionary is empty.

This also demonstrated that there are spots where we are "touching" the overflow (and thus creating a dictionary) when not necessary. We should remove them as much as possible.

ewoutkramer:
## Description
Be careful not to accidentally create an empty overflow dictionary. Made sure HasOverflow actually works.

## Related issues
Fixes #3111.

⚠️ Review https://github.com/FirelyTeam/fhir-codegen/pull/57 first.


Kasdejong (approved):
The fact that we have NONE of this in the generated code (I checked) says a lot about what came first and what came after :)

PR #3137, opened 2025-05-01
3136 Support adding custom properties using ClassMappings

### LINKED ISSUE #3136: Add support for custom elements
Make it possible to alter classmappings so you can remove and add property mappings to allow parsing unknown elements.

ewoutkramer:
Fixes #3136.


src/Hl7.Fhir.Base/Introspection/PropertyMapping.cs
```diff
@@ -228,6 +253,32 @@ public static bool TryCreate(PropertyInfo prop, [NotNullWhen(true)] out Property
...
+        if (Nullable.GetUnderlyingType(implementingType) is { } underlyingType) implementingType = underlyingType;
+
+        if (inspector.FindOrImportClassMapping(implementingType) is not {} propertyTypeMapping)
+            throw new InvalidOperationException($"Custom property {name} is of type " +
+                                                $"{implementingType}, for which a classmapping cannot be found.");
+
+        return new PropertyMapping(declaringClass, name, nativeProperty: null, propertyType)
+        {
+            Choice = allowedTypes is not null ? ChoiceType.DatatypeChoice : ChoiceType.None,
+            Order = Int32.MaxValue,
```
andrzejskowronski: We should have it as an optional parameter too.

andrzejskowronski (changes requested):
Looks good, minor thing and will still need to update the breaking changes doc.

src/Hl7.Fhir.Base/Introspection/PropertyMapping.cs
ewoutkramer: Agree, most of the properties would be. So I am thinking about turning the "CustomProperty" function into a constructor instead.

andrzejskowronski (approved):

PR #3164, opened 2025-05-27
Propagate position information through poco builder

Kasdejong:
## Description
Made the new poco builder look for position information annotations on the incoming typed element, and forward them to the constructed Poco

Added some new tests to confirm this behaviour. Note that this only works for the concrete PositionInfo class, NOT with IPositionInfo

### DIFF
--- a/src/Hl7.Fhir.Base/ElementModel/NewPocoBuilder.cs
+++ b/src/Hl7.Fhir.Base/ElementModel/NewPocoBuilder.cs
@@ -17,6 +17,8 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
+using System.Reflection;
 using ET = Hl7.Fhir.ElementModel.Types;
 
 namespace Hl7.Fhir.ElementModel;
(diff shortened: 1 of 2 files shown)


src/Hl7.Fhir.Base/ElementModel/NewPocoBuilder.cs
```diff
@@ -43,6 +45,10 @@ public Base BuildFrom(ITypedElement source)
     private Base readFromElement(ITypedElement node, ClassMapping classMapping)
     {
         var newInstance = buildNewInstance(classMapping);
+        
+        // copy over the position information, if available.
+        if(node.Annotation<PositionInfo>() is {} positionInfo)
```
andrzejskowronski: As we talked, we'll need to also handle `JsonSerializationDetails` and `XmlSerializationDetails` as those are the concrete ones that are actually used in FhirJsonNode/FhirXmlNode and will get lost otherwise.

andrzejskowronski (changes requested):

andrzejskowronski (approved):

PR #3168, opened 2025-06-02
Expose MemberName by passing it through the ValidationContext

### LINKED ISSUE #3167: Expose MemberName on the CodedValidationException
Previously ValidationResult exposed `MemberNames` returning a collection of property names affected by the validation issue.

We should provide this information as well in the current validation exceptions

andrzejskowronski:
## Description
Adds name of the property that's being validated on the PocoValidationContext to report in case of issues.

## Related issues
Closes #3167

### DIFF
--- a/src/Hl7.Fhir.Base/Validation/CodedValidationException.cs
+++ b/src/Hl7.Fhir.Base/Validation/CodedValidationException.cs
@@ -139,7 +139,7 @@ public class CodedValidationException : ExtendedCodedException
             code, message, path,
             context?.LineNumber, context?.LinePosition, 
             issueSeverity, issueType,
-            context?.MemberName ?? memberName);
+            memberName ?? context?.MemberName);
 
         return codedException;
     }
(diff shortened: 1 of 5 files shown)


src/Hl7.Fhir.Base/Validation/CodedValidationException.cs
```diff
@@ -123,27 +124,32 @@ public CodedValidationException(
...
         var codedException = new COVE(
-            code,
-            message,
-            path,
-            context?.LineNumber,
-            context?.LinePosition, issueSeverity, issueType);
+            code, message, path,
+            context?.LineNumber, context?.LinePosition, 
+            issueSeverity, issueType,
+            context?.MemberName ?? memberName);
```
Kasdejong: tad nitpicky, but should these not be the other way around? I would expect a manual optional parameter to always override the context-inferred name

Kasdejong (changes requested):

src/Hl7.Fhir.Base/Validation/CodedValidationException.cs
andrzejskowronski: Fair, fixed!

Kasdejong (approved):

PR #3169, opened 2025-06-03
Fix line info being erroneously included in location messages

andrzejskowronski:


Kasdejong (approved):
is there no changes to the manifest? how did someone even find this? :)

andrzejskowronski:
Ha, thanks for the question, made a double take!
Those would show up on manual calls to `.Validate()` where we had the path, but no lines

PR #3195, opened 2025-06-25
Make sure that we update the http client's timeout

### LINKED ISSUE #3176: Timeout not producing consistent results when changed.
**Describe the bug**
Setting client timeouts produce inconsistent behaviour.

Been looking at changing some of the timeout behaviour from the client. The default timeout is currently 100 seconds. If that is breached I get a nice stack trace. All good. If I change that value via FhirClientSettings to a lower number, I get a 200 and no stack trace when the timeout is reached. Additionally if I increase that value about 100 seconds I get the default stack trace but at 100 seconds. 

**To Reproduce**
Add FhirClientSettings with the Timeout parameter to the FhirClient. 
Create a wiremock server that returns with a delay.
Change the values below 100 seconds
Change the values above 100 seconds.

**Expected behavior**
To get a stack trace on my configured timeout.

**Version used:**
 - Version: 6.0.0-alpha1 but also tested with 5.11.7

ewoutkramer:
Fixes #3176.


alexzautke (approved):

ewoutkramer:
Solution will never work, cannot change timeout on an HttpClient that has been used. And you cannot create new ones, as this will keep on allocating new ports.

PR #3199, opened 2025-07-01
Merge sdk5 into sdk6

ewoutkramer:
A periodic update to bring the fixes from SDK5 into SDK6.

Mainly:
* R6 support
* Fixes to resolution logic and versioning.

ℹ️ This PR has many files since it contains the newly generated files for FHIR R6.


src/Hl7.Fhir.Base/ElementModel/ScopedNode.cs
```diff
@@ -223,10 +222,25 @@ public IEnumerable<BundledResource> BundledResources()
             
             if (InstanceType == "Bundle")
             {
-                var referenceEntryPairs = from e in this.Children("entry")
-                    let fullUrl = e.Children("fullUrl").FirstOrDefault()?.Value as string
-                    let resource = e.Children("resource").FirstOrDefault() as ScopedNode
-                    select new KeyValuePair<string, ScopedNode>(fullUrl, resource);
+                var referenceEntryPairs = new List<KeyValuePair<string?, ScopedNode>>();
```
andrzejskowronski, line 225: I believe in sdk6 we had this logic stay the same as the actual issue was solved in other way, @Kasdejong can you confirm that is indeed the case or should it stay as is?
If so feel free to resolve this comment :)

andrzejskowronski (approved):

src/Hl7.Fhir.Base/ElementModel/ScopedNode.cs
ewoutkramer: You are correct....This is exactly what I wanted to discuss with @Kasdejong this morning, and he confirmed then this was the intention. But let's double check!

Kasdejong: The SDK-6 version of "ScopedNode" does not support this functionality, precisely because I knew this would be forward merged into SDK-6 at some point. I did two separate fixes, one on scopednode on dev, one on poconode on dev6 :)

PR #3200, opened 2025-07-01
Add global json for CodeQl

mmsmits:


mmsmits:
Not necessary for develop yet. Since it doesn't specify net9.0

PR #3142, opened 2025-05-07
Start development phase 5.11.6

andrzejskowronski:

PR #3233, opened 2025-08-01
3142 Use custom properties & improve Json parser

ewoutkramer:
## Description
* By using the new custom PropertyInfo we have been able to simplify the code a bit. 
* Also, the way unknown elements are dealt with is more in line with how the Xml deserializers does it, giving a more consistent experience to the user (the parsed model containing unknown elements differ less between was xml or json).
* Finally, since the model is now flexible, the code has been simplified quite a bit - there is no difference anymore between dealing with arrays of primitives or arrays of complex datatypes, shaving off code duplication.

## Related issues
Fixes #3142


src/Hl7.Fhir.Base/Model/DynamicDataType.cs
```diff
@@ -48,6 +50,18 @@ protected internal override Base DeepCopyInternal()
...
+    /// <summary>
+    /// Uses reflection emit to create a subclass of DynamicResource with the given type name.
+    /// </summary>
+    /// <returns></returns>
+    public static Type CreateSubclass(string typeName)
+    {
+        if (string.IsNullOrEmpty(typeName))
+            throw new ArgumentNullException(nameof(typeName));
+
+        return DynamicTypeFactory.CreateDynamicResourceSubclass(typeName);
```
andrzejskowronski: Pretty sure that will break on iOS with the no-codegen requirement.

src/Hl7.Fhir.Support.Poco.Tests/Serialization/snapshots/FhirJsonDeserializationTests.LessSimpleVerify.verified.txt
andrzejskowronski, line 1: Artifact of removed test, so we should probably delete it.

src/Hl7.Fhir.Support.Poco.Tests/Serialization/snapshots/FhirJsonDeserializationTests.SimpleVerify.verified.txt
andrzejskowronski, line 1: Artifact of removed test, so we should probably delete it.

src/Hl7.Fhir.Support.Poco.Tests/Serialization/FhirJsonDeserializationTests.cs
```diff
@@ -12,38 +12,70 @@
...
+        _settings = new VerifySettings();
+        _settings.UseDirectory("snapshots");
+        _settings.DisableDiff();
+    }
+
+    [TestMethod]
+    public async T.Task VerifyVerifier() =>
+        await VerifyChecks.Run();
+
+    internal string buildVerifierPath([CallerFilePath] string sourceFile = "")
```
andrzejskowronski: We'll probably need some helper method if we were to introduce it in other tests, but that can be a separate issue.

andrzejskowronski (changes requested):

src/Hl7.Fhir.Base/Model/DynamicDataType.cs
ewoutkramer: Oh, this was scratch code that is not used at all. It should be removed completely.

src/Hl7.Fhir.Support.Poco.Tests/Serialization/FhirJsonDeserializationTests.cs
ewoutkramer: It sounds attractive, but we do not really have cross-repository utility libraries currently. Or maybe you mean within this repo. Yeah, that we should do - within this test project at least.

andrzejskowronski: Yeah I meant specifically within sdk for now, cross repository utils would be much more effort :D
Just thought of it since the `Serialization` part will obviously be different based on where test file is relative to the project base, so it would probably help to avoid the path confusion in the future.

andrzejskowronski (approved):

PR #3239, opened 2025-08-07
FS back-compatibility SDK tweaks

andrzejskowronski:


src/Hl7.Fhir.Base/ElementModel/NewPocoBuilder.cs
```diff
@@ -111,6 +111,7 @@ private static void raiseFormatError(string message, string location)
 
     private static Base buildNewInstance(ClassMapping mapping)
     {
+        if (mapping.NativeType.IsAbstract) return new DynamicResource();
```
ewoutkramer: Shouldn't this depend on mapping.IsResource? Otherwise create a DynamicDataType (I guess that's what we need for ba ckbones?)

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/ElementModel/NewPocoBuilder.cs
andrzejskowronski: Fair, fixed ;)

ewoutkramer (approved):

PR #3242, opened 2025-08-11
error on empty element validation

### LINKED ISSUE #3230: Add POCO validation equivalent of "ele-1"
Currently, the XML and JSON parser report "empty objects" (Json: OBJECTS_CANNOT_BE_EMPTY, Xml: ELEMENT_HAS_NO_VALUE_OR_CHILDREN).

This should not be a syntax error, but a model error, caught by the attribute validator, since this is also illegal when constructing a POCO manually. So, we should add to Base's ValidateObject() the check that it has elements (or a value in case of a primitive).

OBJECTS_CANNOT_BE_EMPTY can then remain in json for the purelty syntactic error:

```
primitive: 4
_primitive: {}
```

which would otherwise go unnoticed.

Kasdejong:
## Description
Removed xml-specific check for value or children in favour of a model-based one. Json empty objects will still have their own syntax exceptions, because:
```
primitive: 5
_primitive: {}
```
still results in a valid primitive instance, but an error should still be thrown.

## Related issues
Closes #3230


src/Hl7.Fhir.Base/Model/PrimitiveType.cs
```diff
@@ -59,7 +59,9 @@ bool P.IToSystemPrimitive.TryConvertToSystemType([NotNullWhen(true)] out P.Any?
     protected internal override IReadOnlyCollection<CodedValidationException> ValidateInvariants(
         PocoValidationContext validationContext)
     {
-        var baseResults = base.ValidateInvariants(validationContext);
+        IReadOnlyCollection<CodedValidationException> baseResults = [];
+        if (this.JsonValue is null)
```
ewoutkramer: Looks like an optimization (since you know the bases' validation are only about emptiness and JsonValue is an important factor), but if there would be other validations on the base, it would be dependent on the JsonValue on whether they are run or not. That is pretty implicit like this. If that's what you meant, maybe add a comment to tell the reader about it.

src/Hl7.Fhir.Support.Poco.Tests/Serialization/FhirXmlDeserializationTests.cs
```diff
@@ -20,10 +20,10 @@ namespace Hl7.Fhir.Support.Poco.Tests;
 public class FhirXmlDeserializationTests
 {
     [TestMethod]
-    public void SerializingErroneousResource_Should_ThrowExpectedErrors() => testRecovery(false, "TestData");
+    public void SerializingErroneousResource_Should_ThrowExpectedErrors() => testRecovery(false, "../../../TestData");
 
     [TestMethod]
-    [Ignore]
+    // [Ignore]
```
ewoutkramer: This should be ignored. And we should make it a verifier test, like the json one!

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #3248, opened 2025-08-14
Handle long literal parsing in Integer64 validation

andrzejskowronski:


alexzautke:
Can we add a unit test for this?

PR #3250, opened 2025-08-14
Feature/open choice types in model

### LINKED ISSUE #3231: Detect DynamicData in POCO choice elements
The Json and Xml parser are currently detecting unknown types (Json+Xml: CHOICE_ELEMENT_HAS_UNKNOWN_TYPE), but this should really be caught by the model, since dynamic data can also be used when manually constructing POCOs and it is not syntactical knowledge that disappears after parsing.

This is partially done by [AllowedTypes] (since the dynamic type will not be a member of the allowed types), but since "open types" (like Extension.value) are not marked with [AllowedTypes], this error is not caught there.

We can either repair this now, or change the error raised by the json/xml parsers to be the same COVE errors as used by the [AllowedTypes] attribute, so we can repair this later.

Kasdejong:
## Description
Added support for open choice types to the model. The model will now report an invalid choice type issue if the suffix for an open type is unknown or missing, and it will no longer accept types that are not in the list of valid choice types for a specific version of FHIR

## Related issues
Resolves #3231 

also review FirelyTeam/fhir-codegen#65


src/Hl7.Fhir.R4B/Model/Generated/StructureMap.cs
```diff
@@ -1946,7 +1946,7 @@ public string? Type
...
...ing),typeof(Hl7.Fhir.Model.Time),typeof(Hl7.Fhir.Model.UnsignedInt),typeof(Hl7.Fhir.Model.FhirUri),typeof(Hl7.Fhir.Model.FhirUrl),typeof(Hl7.Fhir.Model.Uuid),typeof(Hl7.Fhir.Model.Address),typeof(Hl7.Fhir.Model.Age),typeof(Hl7.Fhir.Model.Annotation),typeof(Hl7.Fhir.Model.Attachment),typeof(Hl7.Fhir.Model.CodeableConcept),typeof(Hl7.Fhir.Model.Coding),typeof(Hl7.Fhir.Model.ContactPoint),typeof(Hl7.Fhir.Model.Count),typeof(Hl7.Fhir.Model.Distance),typeof(Hl7.Fhir.Model.Duration),typeof(Hl7.Fhir.Model.HumanName),typeof(Hl7.Fhir.Model.Identifier),typeof(Hl7.Fhir.Model.Money),typeof(Hl7.Fhir.Model.Period),typeof(Hl7.Fhir.Model.Quantity),typeof(Hl7.Fhir.Model.Range),typeof(Hl7.Fhir.Model.Ratio),typeof(Hl7.Fhir.Model.ResourceReference),typeof(Hl7.Fhir.Model.SampledData),typeof(Hl7.Fhir.Model.Signature),typeof(Hl7.Fhir.Model.Timing),typeof(Hl7.Fhir.Model.ContactDetail),typeof(Hl7.Fhir.Model.Contributor),typeof(Hl7.Fhir.Model.DataRequirement),typeof(Hl7.Fhir.Model.Expression),typeof(Hl7.Fhir.Model.ParameterDefinition),typeof(Hl7.Fhir.Model.RelatedArtifact),typeof(Hl7.Fhir.Model.TriggerDefinition),typeof(Hl7.Fhir.Model.UsageContext),typeof(Hl7.Fhir.Model.Dosage),typeof(Hl7.Fhir.Model.Meta))]
```
ewoutkramer: Unfortunately, this is incorrect.  In R4 this was an open type, but -invisible to us- in R4B this is no longer an open type, as it needs to allow for for example Meta (which is not an open type), so here, the generated list should remain.

It's impossible to distinguish, so I propose only using your new method to determine open types to work when producing base/conformance, and otherwise just generate the full list as specified in the satellite?

src/Hl7.Fhir.Support.Poco.Tests/Serialization/snapshots/FhirJsonDeserializationTests.SerializingErroneousResource_Should_ThrowExpectedErrors.verified.txt
```diff
@@ -620,26 +620,40 @@
       Message: Choice element 'value' should be suffixed by a type. At Patient.gender.extension[2].value, line 302, position 18.
     },
     {
-      Type: FhirJsonException,
+      Type: CodedValidationException,
+      MemberName: value,
       IssueSeverity: Error,
       IssueType: Value,
-      BaseErrorMessage: Choice element 'value' is suffixed with an unrecognized type 'SuperDecimal'.,
+      BaseErrorMessage: Value is of type 'UnknownType_Patient.gender.extension[2].value', which is not an allowed choice.,
```
ewoutkramer, line 627: This looks like a new error to me in comparison to the previously verified version, is that correct?

src/Hl7.Fhir.Support.Poco.Tests/Serialization/snapshots/FhirJsonDeserializationTests.SerializingErroneousResource_Should_ThrowExpectedErrors.verified.txt
```diff
@@ -1172,11 +1186,11 @@
     },
     {
       "url": "http://example.org/StructureDefinition/extension-without-type",
-      "valueDecimal": 28
+      "valueUnknownType_Patient.gender.extension[2].value": 28
```
ewoutkramer, line 1189: Ah, that's because we're not assuming Decimal anymore....

src/Hl7.Fhir.Support.Poco.Tests/Serialization/snapshots/FhirXmlDeserializationTests.SerializingErroneousResource_Should_ThrowExpectedErrors.verified.txt
```diff
@@ -396,17 +396,6 @@
       ErrorCode: PVAL128,
       Message: Found unknown attribute 'valueString'. At Patient.contained[10].valueString, line 70, position 18.
     },
-    {
```
ewoutkramer, line 399: Is it ok that this disappeared?

ewoutkramer (changes requested):

src/Hl7.Fhir.R4B/Model/Generated/StructureMap.cs
Kasdejong: meta IS an open type though... at least, in the modelinfo, it is...

src/Hl7.Fhir.Support.Poco.Tests/Serialization/snapshots/FhirJsonDeserializationTests.SerializingErroneousResource_Should_ThrowExpectedErrors.verified.txt
Kasdejong: yes. We used to take a guess, and the guessed type would obviously always be compatible with extension.Value[x]

src/Hl7.Fhir.Support.Poco.Tests/Serialization/snapshots/FhirXmlDeserializationTests.SerializingErroneousResource_Should_ThrowExpectedErrors.verified.txt
Kasdejong: yes. The parser errors for unknown suffixes are gone, same as in json! only in json, we took a guess, and in xml we already made a dynamic type apparently...

src/Hl7.Fhir.R4B/Model/Generated/StructureMap.cs
ewoutkramer: And so we discovered a discrepancy between the open types as written in the *text* of the spec and what is generated at `Extension.value[x]`.  We decided to take `Extension.value[x]`'s list as leading.

ewoutkramer (approved):

Issue #3251, opened 2025-08-14
QuestionnaireItemType missing Question in STU3/R4


andrzejskowronski:
From 5 -> 6 seems the enum value disappeared from the enum for some reason. but are still there in R4B & R5.

ewoutkramer:
The reason is that the question code is *abstract*, and should never appear in real instances. It's only there to reflect a hierarchy of codes. We noticed this during SDK5, but could not remove this code until a major new release. It also means it should also not be there in R4B and R5, I will need to check how they can still be present there.

ewoutkramer:
I also noted that @andrzejskowronski has been changing generated files, which is doomed to result in lost work. Would it be an idea to conform to the CQL SDK's habit of including ".g" (so Patient.g.cs) in the name to make this clearer?

andrzejskowronski:
I think it's clear with the `\Generated\` path, but I'm fine with either way ;)

PR #3252, opened 2025-08-14
Restore enum missing from codegen

andrzejskowronski:


src/Hl7.Fhir.R4/Model/Generated/Questionnaire.cs
```diff
@@ -84,6 +84,12 @@ public enum QuestionnaireItemType
       [EnumLiteral("display"), Description("Display")]
       Display,
       /// <summary>
+      /// An item that defines a specific answer to be captured, and which may have child items. (the answer provided in the QuestionnaireResponse should be of the defined datatype).
```
ewoutkramer, line 87: You cannot manually add code to generated files, these will be overwritten next time. 

Also, this code is no longer generated on purpose (it was accidentally generated, but we could never remove it until a major new release). This code is *abstract* (It's in italics in https://hl7.org/fhir/R4/valueset-item-type.html) and will never appear in an instance, only the codes below it in hierarchy, so it should not be part of this enum.

Your PR has no context or description on why you are doing this, so did you encounter this code somewhere? Did someone report a bug?

ewoutkramer (changes requested):

ewoutkramer:
Fixes #3251.

@andrzejskowronski - can you please keep adding these "fixes" notes so your PRs refer back to the original issues? It's hard for a reviewer to get context otherwise!

src/Hl7.Fhir.R4/Model/Generated/Questionnaire.cs
andrzejskowronski: Actually generated it as part of codegen, but don't have a PR ready for that yet, and hoped to get approval/feedback before Monday :D
Encountered when updating Enterprise validator to sdk6, but I can drop it instead, whichever makes more sense.
The enum is generated differently because we don't run expansion if the elements are already there, and the element in question is not in STU3/R4, but is in the R4B/R5/R6.

andrzejskowronski: So, instead of finding how to add the element in, I'll instead make sure it is removed in further versions ;)

ewoutkramer: Haha, yes, I double checked, it is indeed generated in R5 etc.

ewoutkramer: A simple check for IsAbstract seems to fix the issue in `WriteEnum()`.

PR #3276, opened 2025-09-08
Poco conversion fixes

andrzejskowronski:
Improve Poco conversions: 

- avoiding building a new poco if we already have one of requested type (via PocoNode)
- handling the case where PocoNode got constructed with no information of the type of data (Dynamic...), but later gets converted to a specific Poco, where we now are aware of a correct type of the data

### DIFF
--- a/src/Hl7.Fhir.Base/ElementModel/NewPocoBuilder.cs
+++ b/src/Hl7.Fhir.Base/ElementModel/NewPocoBuilder.cs
@@ -65,8 +65,8 @@ internal class NewPocoBuilder(ModelInspector inspector, PocoBuilderSettings? set
             if (newInstance is DynamicPrimitive)
                 objectValue = value;
             // we're trying to add the type data to untyped values
-            else if (node is PocoNode { Poco: IDynamicType } && classMapping.PrimitiveValueProperty is not null)
-                objectValue = PrimitiveTypeConverter.ConvertTo(value, classMapping.PrimitiveValueProperty.ImplementingType);
+            else if (node is PocoNode { Poco: IDynamicType } && value is string s && classMapping.PrimitiveValueProperty is not null)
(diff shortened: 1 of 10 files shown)


alexzautke (approved):

src/Hl7.Fhir.Base/ElementModel/TypedElementOnSourceNode.cs
```diff
@@ -282,6 +283,8 @@ private static bool tryParse(string value, Type primitiveType, [NotNullWhen(true
             if (info.IsChoiceElement)
             {
                 var suffix = current.Name.Substring(info.ElementName.Length);
+                if (string.IsNullOrEmpty(suffix) && current is PocoNode { Poco: DataType { TypeName: {} tn}})
```
ewoutkramer, line 286: This might not be necessary anymore.

src/Hl7.Fhir.Base/ElementModel/NewPocoBuilder.cs
```diff
@@ -229,8 +266,10 @@ private ClassMapping getClassMapping(string dynTypeName) =>
         throw Error.InvalidOperation($"Cannot find ClassMapping for type '{dynTypeName}'.");
 
     private ClassMapping getClassMapping(Type t) =>
-        inspector.FindOrImportClassMapping(t) ??
-        throw Error.InvalidOperation($"Cannot find ClassMapping for type '{t.Name}'.");
+        inspector.FindClassMapping(t) ?? 
+        (ClassMapping.TryCreate(inspector, t, out var newMapping)
```
ewoutkramer, line 270: This might not be necessary anymore.

ewoutkramer (commented):
I think we should clean up a few cases that we think are not necessary anymore + provide some more remarks, so this remains better to understand in a year or so.....

PR #3277, opened 2025-09-15
Expose extension methods for PocoNode

andrzejskowronski:


src/Hl7.Fhir.Base/Model/PocoNodeExtensions.cs
```diff
@@ -161,7 +161,7 @@ private static IEnumerable<PocoNode> parents(this PocoNodeOrList node)
     /// </summary>
     /// <param name="node"></param>
     /// <returns></returns>
-    internal static string? FindFullUrl(this PocoNodeOrList node)
+    public static string? FindFullUrl(this PocoNodeOrList node)
```
cknaap, line 164: I can imagine that you need these 2 (FindFullUrl & GetParentResource) to work together for Bundle.entry.fullUrl & ~.resource. But if you call them separately, 
- `parents()` is evaluated twice
- you still have to check whether the parent resource is part of the same entry (it should be, of course, but it is not obvious)
Should we instead offer something like FindParentBundleEntry?

src/Hl7.Fhir.Base/Model/PocoNodeExtensions.cs
```diff
@@ -161,7 +161,7 @@ private static IEnumerable<PocoNode> parents(this PocoNodeOrList node)
...
     {
         var entry = node.parents().FirstOrDefault(n => n.Poco is Bundle.EntryComponent);
         return entry?.Child<PrimitiveNode>("fullUrl")?.Value as string;
```
cknaap: Why not
```suggestion
        return (entry?.Poco as Bundle.EntryComponent)?.FullUrl;
```
?

src/Hl7.Fhir.Base/Model/PocoNodeExtensions.cs
```diff
@@ -335,4 +346,20 @@ internal static string SerializeToString(this PocoNode pn, bool pretty)
...
         return serializer.SerializeToString(pn.Poco, pretty, rootName: rootName);
     }
+
+    /// <summary>
+    /// Converts the specified PocoNode instance to its JSON representation.
+    /// </summary>
+    /// <param name="pn">The PocoNode instance to serialize.</param>
+    /// <param name="pretty">Indicates whether the JSON output should be formatted for readability.</param>
+    /// <returns>A JSON string representing the given PocoNode.</returns>
+    public static string ToJson(this PocoNode pn, bool pretty)
```
cknaap, line 359: This functionality was already accessible through the ITypedElement interface: `Hl7.Fhir.ElementModel.TypedElementExtensions.ToJson` (the other change in the PR). It was not added on PocoNode before. Was that simply an oversight, or for some reason deliberate?

src/Hl7.Fhir.Base/Model/PocoNodeExtensions.cs
```diff
@@ -335,4 +346,20 @@ internal static string SerializeToString(this PocoNode pn, bool pretty)
...
+    {
+#pragma warning disable CS0618// Type or member is obsolete
```
cknaap, line 361: It looks strange to use obsolete calls in new code. Especially now you're making it public and will have to support this method indefinitely.
Will you ever be able to solve that?

cknaap (changes requested):

src/Hl7.Fhir.Base/Model/PocoNodeExtensions.cs
andrzejskowronski: Yes I believe mostly oversight about issues with overload resolution between ITypedElement and ISourceNode versions

src/Hl7.Fhir.Base/Model/PocoNodeExtensions.cs
cknaap: The obsolete ForType is used in other places as well, it is obsolete to guide the public away from it. 
Check with unit tests whether/when FindInspector might return the common Base inspector where you'd expect a FHIR version specific inspector.

cknaap: Will add those tests in a separate tasks.

cknaap (approved):

PR #3279, opened 2025-09-19
Ensure correct overload gets called for parents location

andrzejskowronski:
Wrong overload getting called would result in different behavior at node vs at child node
`Extension.valueReference` -> `Extension.value.reference`


alexzautke (commented):
Is it possible to add a unit test for this?

alexzautke (approved):

PR #3281, opened 2025-09-22
Bugfix/append text

### LINKED ISSUE #3280: The text append notation ... is applied twice in certain cases
**Description**
When an element has a type profile then the an element merge is performed twice.
Normally this is not a problem as the second merge will just replace the result of the first merge but when appending text as a result of the ... notation then text is appended twice.

**To Reproduce**
This branch contains two unit tests to reproduce the problem and a possible fix (feel free to solve it in a different way).
https://github.com/FirelyTeam/firely-net-sdk/tree/bugfix/append-text

**Expected behavior**
Text should only be appended once.

**Screenshots**
Example from Forge
Image

**Version used:**
 - FHIR Version: R5
 - Version: 5.12.2

**Solutions**
The solution in the branch is to mark an element that text has been appended with an annotation and only append if there is no annotation.

Other solutions:
- Prevent multiple merging of the same element (out of my league...)
- Compare the end of the text with the text that is to be appended and skip if they are the same

alexzautke:
Fixes #3280

### DIFF
--- a/src/Hl7.Fhir.Conformance/Specification/Snapshot/SnapshotGeneratorAnnotations.cs
+++ b/src/Hl7.Fhir.Conformance/Specification/Snapshot/SnapshotGeneratorAnnotations.cs
@@ -231,5 +231,55 @@ namespace Hl7.Fhir.Specification.Snapshot
 
         #endregion
 
+        #region Annotation: AppendedText
+
+        /// <summary>
+        /// Custom annotation for elements and properties in the <see cref="StructureDefinition.SnapshotComponent"/>
+        /// that are constrained by the <see cref="StructureDefinition.DifferentialComponent"/>.
+        /// </summary>
+        [Serializable]
+        sealed class AppendedTextAnnotation
+        {
+            //
+        }
+
+        /// <summary>Annotate the specified snapshot element to indicate that it is constrained by the differential.</summary>
+        internal static void SetAppendedTextAnnotation(this Base element)
+        {
+            if (element == null) { throw Error.ArgumentNull(nameof(element)); }
+            element.AddAnnotation(new AppendedTextAnnotation());
+        }
+
+        /// <summary>Remove any existing differential constraint annotation from the specified snapshot element.</summary>
+        internal static void RemoveAppendedTextAnnotation(this Base element)
+        {
+            if (element == null) { throw Error.ArgumentNull(nameof(element)); }
+            element.RemoveAnnotations<AppendedTextAnnotation>();
+        }
+
+        /// <summary>Recursively remove any existing differential constraint annotations from the specified snapshot element and all it's children.</summary>
+        internal static void RemoveAllAppendedTextAnnotations(this Base element)
+        {
+            if (element == null) { throw Error.ArgumentNull(nameof(element)); }
+            element.RemoveAppendedTextAnnotation();
+            foreach (var child in element.Children)
+            {
+                child.RemoveAllAppendedTextAnnotations();
+            }
+        }
+
+        /// <summary>Recursively remove any existing differential constraint annotations from the specified snapshot elements and all their children.</summary>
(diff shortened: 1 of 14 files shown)


src/Hl7.Fhir.Conformance/Specification/Snapshot/SnapshotGeneratorAnnotations.cs
```diff
@@ -20,6 +20,67 @@ namespace Hl7.Fhir.Specification.Snapshot
...
+            }
+        }
+
+        /// <summary>Recursively remove any existing snapshot generator annotations from the specified snapshot elements and all their children.</summary>
+        internal static void RemoveAllSnapshotGeneratorAnnotations<T>(this IEnumerable<T> elements) where T : Base
+        {
+            if (elements == null) { throw Error.ArgumentNull(nameof(elements)); }
+            foreach (var elem in elements)
+            {
+                elem.RemoveSnapshotGeneratorAnnotation();
```
ewoutkramer: If this function is called `RemoveAllSnapshotGeneratorAnnotations><T>` then this should call `elem.RemoveAllSnapshotGeneratorAnnotations()` or `elem.RemoveSnapshotGeneratorAnnotation()`, otherwise I guess the name of the function is wrong!

src/Hl7.Fhir.Conformance/Specification/Snapshot/SnapshotGeneratorAnnotations.cs
```diff
@@ -20,6 +20,67 @@ namespace Hl7.Fhir.Specification.Snapshot
...
+        #region Used internally by Snapshot Generator
+
+        internal static void RemoveSnapshotGeneratorAnnotation(this Base element)
+        {
+            element.RemoveConstrainedByDiffAnnotation();
+            element.RemoveAppendedTextAnnotation();
+        }
+
+        /// <summary>Recursively remove any existing snapshot generator annotations from the specified snapshot element and all it's children.</summary>
+        internal static void RemoveAllSnapshotGeneratorAnnotations(this Base element)
```
ewoutkramer, line 32: Maybe add "Recursive" to the end, instead of the "All" halfway to make the difference more clear.

src/Hl7.Fhir.Conformance/Specification/Snapshot/SnapshotGeneratorAnnotations.cs
ewoutkramer: In the code that you rewrote, this indeed called the "all" version...

ewoutkramer (changes requested):

src/Hl7.Fhir.Conformance/Specification/Snapshot/SnapshotGeneratorAnnotations.cs
Rob5045: Other similar methods use All as well (e.g. RemoveAllNonInheritableExtensions). Not saying that All is better but I leave it to the SDK team to change them from All to Recursive.

Rob5045:
@ewoutkramer I processed your comments, I have added the fix for STU3 and moved the unit test to the shared tests.

Rob5045:
With STU3 being separate from the rest there is a lot of code duplication. Maybe it is a good idea to have a shared test project for code that can be used for all FHIR versions?

ewoutkramer:
> With STU3 being separate from the rest there is a lot of code duplication. Maybe it is a good idea to have a shared test project for code that can be used for all FHIR versions?

There is a "Common" test project for tests that work against all branches, and otherwise we use shared projects to create tests that really depend on resources (i.e. Patient)!

ewoutkramer:
That said, I know there is still quite some duplication - originally these were separate repos and Marco did his best to deduplicate, but certainly in the tests, there is still a long way to go (luckily not so much in the real code).

ewoutkramer (approved):

PR #3291, opened 2025-09-27
Added check on absolute content reference when expanding elements

### LINKED ISSUE #3272: Snapshot generator method ExpandElement does not handle absolute content references
The snapshot generator method ExpandElement does not expand elements that have an absolute content reference (e.g. _http://hl7.org/fhir/StructureDefinition/Composition#Composition.section_). Elements with an relative content reference (e.g. _#Composition.section_) do get expanded.

The method getStructureForContentReference does not check for absolute content references and tries to find a core type that does not exit (e.g. "ttp://hl7"):

```
var contentReference = elementDef.ContentReference; // e.g. "#Questionnaire.item"

// [WMR 20181212] TODO: Handle logical models, where StructureDefinition.type returns an uri

var coreType = nav.StructureDefinition?.Type
    // Fall back to root element name...?
    ?? ElementDefinitionNavigator.GetPathRoot(contentReference.Substring(1));

```

Rob5045:
Fixes https://github.com/FirelyTeam/firely-net-sdk/issues/3272


ewoutkramer:
HI Rob, can you check whether this is related: https://github.com/FirelyTeam/firely-net-sdk/pull/2054?

This is from 2 years ago, but I thought it tried to fix the same problem. Apparently not, but I'm not sure there is no overlap...

ewoutkramer (approved):

PR #3320, opened 2025-10-14
Make our develop branch 6.0

Kasdejong:
## Description
We should make 6.0 our base version for the SDK, so that the github index page shows 6.0, and it is clear that we would rather accept PRs against 6.0.


andrzejskowronski (dismissed):

andrzejskowronski:
Done in #3333 due to merge conflicts

Issue #3327, opened 2025-10-24
A ValueSet resource expanded by LocalTerminologyService/ValueSetExpander causes NullReferenceException during serialization process.


v-isyamauchi-gh:
:information_source: Do you want to ask something or start a discussion, then go [here](https://github.com/FirelyTeam/firely-net-sdk/discussions)

:information_source: Is your bug report about Firely Server Facade, please send an email to [server@fire.ly](mailto:server@fire.ly) or submit a bug report to Jira if you are a paying customer.

**Describe the bug**
A ValueSet resource expanded by LocalTerminologyService/ValueSetExpander causes NullReferenceException during serialization process (in PocoElementNode [ctor](https://github.com/FirelyTeam/firely-net-sdk/blob/61cc68da1affa6c4121dabe561564a28cf97915e/src/Hl7.Fhir.Base/ElementModel/PocoElementNode.cs#L50) due to 'definition' parameter being null).

The issue appears to be the 'expansion' field added by the expander having 'property' field ([ValueSet.expansion.contains.property](https://hl7.org/fhir/R5/valueset.html)) that exists in R5 or above, not in lower versions.

**To Reproduce**
Run the code below with Fhir version R4 or below.
```
    var server = new LocalTerminologyService(
        new MultiResolver(new CachedResolver(ZipSource.CreateValidationSource())));
    var parameters = new Parameters()
    {
        Parameter = new List<Parameters.ParameterComponent>
        {
            new Parameters.ParameterComponent()
            {
                Name = "url",
                Value = new FhirUri("http://hl7.org/fhir/ValueSet/diagnostic-service-sections"),
            },
        },
    };
    var resource = await server.Expand(parameters);
    var s = await resource.ToJsonAsync();
```

**Expected behavior**
The json serialization should succeed (the expander should not add any property not existing in a specific Fhir version).

**Screenshots**

**Version used:**
 - FHIR Version: R4(B) or below
 - Version: 5.11.4

**Additional context**

ewoutkramer:
I think this and #3329 are related.

andrzejskowronski:
We are not really able to make expander not assign the property since the assembly the logic resides in is shared between all R4+ versions with no distinction. That said the serialization should be working on preview package and in the future releases.

PR #3339, opened 2025-10-30
Merge fix for #3327 for SDK6

andrzejskowronski:
## Description
Fix legacy typed element path for possible NRE


andrzejskowronski:
Warning not to merge the develop updates into the support branch, @mmsmits can we protect the support branch so it won't be suggested?

Kasdejong (dismissed):

mmsmits:
@andrzejskowronski I don't really completely understand what you mean, but I applied the same protections now as the normal develop branch. Is that enough?

alexzautke:
@mmsmits I think we need to relax the requirement that the PR must be up-to-date as we can't currently merge it

ewoutkramer:
Oops, I think I merged it ;-)

ewoutkramer:
Or did I, no I just updated this PR. Or maybe this is the problem... this PR did not have its own branch, Shouldn't it have had that?

ewoutkramer:
Ok, I reverted the merge, which temporarily fixes the problem haha, since develop now is in the history of support,.,,,,

alexzautke:
Should we do a force push here and bring it back to the old state?

ewoutkramer:
There is only one possible reaction to force pushing on a public repo!  It's something with a tree and a rope 🤣 

I reverted it, so we don't need to take deparate measures ;-)

ewoutkramer:
Mmmm....this PR now has 300+ changes.

I think we have to step back and analyze what's going on.

ewoutkramer:
I don't think you can use merge here, and this should have been a cherry pick.  But I am confused now: this is new for the SDK, but the FS team did this before, right?

mmsmits:
For FirelyServer we also cherry pick fixes that need to go from a support branch into develop. That's the only way. I suggest we close this PR, create a new branch from develop, cherry pick the fixes and merge that.

PR #3341, opened 2025-11-03
Added call to RemoveAllNonInheritableExtensions to ExpandElementAsync

### LINKED ISSUE #3059: Non inheritable extensions are not removed when expanding a snapshot
When generating a snapshot non inheritable extensions are removed from elements. However when expanding an element they are not.

Rob5045:
## Related issues
Resolves #3059 .

## Testing
Updated existing unit test.

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Base/Specification/Snapshot/SnapshotGeneratorExtensions.cs
+++ b/src/Hl7.Fhir.Base/Specification/Snapshot/SnapshotGeneratorExtensions.cs
@@ -99,7 +99,16 @@ namespace Hl7.Fhir.Specification.Snapshot
 
         internal static bool HasNonInheritableExtensions(this IExtendable element)
         {
-            return element != null && _nonInheritableExtensions.Any(ext => element.GetExtension(ext) != null);
+            if (element == null) 
+                return false;
+
+            if (element.Extension.Any(ext => _nonInheritableExtensions.Contains(ext.Url)))
+                return true;
+
+            if (element is IModifierExtendable modifierElement)
+                return modifierElement.ModifierExtension.Any(ext => _nonInheritableExtensions.Contains(ext.Url));
+
+            return false;
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


src/Hl7.Fhir.Base/Specification/Snapshot/SnapshotGeneratorExtensions.cs
```diff
@@ -88,6 +88,20 @@ public static void RemoveAllConstrainedByDiffExtensions<T>(this IEnumerable<T> e
...
+            if (element == null) return false;
+            if (element.HasNonInheritableExtensions()) return true;
+#pragma warning disable CS0618 // Type or member is obsolete
+            return element.Children().OfType<Element>().Any(child => child.HasAnyNonInheritableExtensions());
+#pragma warning restore CS0618 // Type or member is obsolete
+        }
+
+        internal static bool HasNonInheritableExtensions(this IExtendable element)
+        {
+            return element != null && _nonInheritableExtensions.Any(ext => element.GetExtension(ext) != null);
```
ewoutkramer: I am a bit worried about the number of iterations here:  for each element in the profile we go over all `_nonInheritableExtensions`, and for each of those we scan the extension list to match a string. That sounds like NxMxP, where P is a low number.

If we'd make _nonInheritableExtensions a hashset, then we could reverse this bit and do `element.Extensions.Any(e => _nonInheritableExtensions.Contains(e.Url.Value))` (or something along those lines). That would probably save a lot of scans.

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/Specification/Snapshot/SnapshotGeneratorExtensions.cs
Rob5045: You were right. Optmizing it as you suggested will increase performance. A quick test in Forge showed a decrease of the load time by 5 - 10%.

mmsmits (approved):

ewoutkramer (approved):

PR #3356, opened 2025-12-02
VONK-9012 implemented CachingTerminologyService

### LINKED ISSUE #3190: Create a cached ITerminologyService
Create a TermServiceCacher implementation of ITerminologyService (or at least for the interface that includes codesystem-validate-code and valueset-validate-code) and let it cache the calls to a terminology service. If a certain number of hits per minute for a valueset has been reached, ask for the expansion and then answer locally. Note that this requires care: we should make sure the local semantics for the check are the same as the external semantics. In any doubt (e.g. parameters that we don't understand) => refer back to the original terminology service.

Kasdejong:
## Description
Implemented CachingTerminologyService

Fixes #3190

### DIFF
--- a/src/Hl7.Fhir.Base/Specification/Terminology/CachingTerminologyService.cs
+++ b/src/Hl7.Fhir.Base/Specification/Terminology/CachingTerminologyService.cs
@@ -118,15 +118,10 @@ internal static class ParametersExtensions
         var hash = new HashCode();
         foreach (var parameter in parameters.Parameter)
         {
-            hash.Add(parameter.Name);
-            if (parameter.Value != null)
-                hash.Add(getTerminologyValueHashCode(parameter.Value));
-            if (parameter.Resource != null)
+            if (parameter.getPartHashCode() is { } hashPart)
+                hash.Add(hashPart);
+            else
                 return null;
-            foreach (var part in parameter.Part)
-            {
-                hash.Add(getPartHashCode(part));
-            }
         }
         return hash.ToHashCode();
     }
@@ -141,7 +136,10 @@ internal static class ParametersExtensions
             return null;
         foreach (var subpart in part.Part)
         {
-            hash.Add(getPartHashCode(subpart));
+            if(subpart.getPartHashCode() is { } subpartHash)
+                hash.Add(subpartHash);
+            else
+                return null;
         }
         return hash.ToHashCode();
     }
@@ -168,13 +166,16 @@ internal static class ParametersExtensions
                 break;
             case Coding coding:
                 hash.Add(coding.System);
+                hash.Add(coding.Version);
                 hash.Add(coding.Code);
                 hash.Add(coding.Display);
                 break;
             case CodeableConcept concept:
+                hash.Add(concept.Text);
                 foreach (var coding in concept.Coding)
                 {
                     hash.Add(coding.System);
+                    hash.Add(coding.Version);
                     hash.Add(coding.Code);
                     hash.Add(coding.Display);
                 }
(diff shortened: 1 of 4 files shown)


src/Hl7.Fhir.Base/Specification/Terminology/CachingTerminologyService.cs
```diff
@@ -0,0 +1,136 @@
...
+            if (parameter.Value != null)
+                hash.Add(parameter.Value.GetHashCode());
+            if (parameter.Resource != null)
+                return null;
+            foreach (var part in parameter.Part)
+            {
+                hash.Add(part.Name);
+                if (part.Value != null)
+                    hash.Add(part.Value.GetHashCode());
+                if (part.Resource != null)
```
ewoutkramer: A `part` can have `parts` again. In fact, this is a recursive structure, so I would expect a recursive call here: a function that turns a Parameters.ParameterComponent into a hashcode. So, then you can just call this function straight from the foreach (instead of the current body), and that function would call itself on its parts.

src/Hl7.Fhir.Base/Specification/Terminology/CachingTerminologyService.cs
```diff
@@ -0,0 +1,136 @@
...
+
+    private readonly MemoryCacheEntryOptions _entryOptions;
+
+    public Task<Parameters> ValueSetValidateCode(Parameters parameters, string? id = null, bool useGet = false)
+    {
+        return parameters.GetParameterComponentHashCode() is { } hash
+            ? _cache.GetOrCreate<Task<Parameters>>(hash, entry =>
+                {
+                    entry.SetOptions(_entryOptions);
+                    return _terminologyService.ValueSetValidateCode(parameters, id, useGet);
```
ewoutkramer, line 51: I would cache the result, not the Task. Unless you really know what you are doing, which I am not.

src/Hl7.Fhir.Base/Specification/Terminology/CachingTerminologyService.cs
ewoutkramer: There is, in fact, a `GetOrCreateAsync()` available on cache, so that signals to me we should await the task and store the result.

ewoutkramer: A lucky side-effect is that you are capturing exceptions in the Task<T>, so calling the same failed operation again will raise the exception again. If you don't use Task<T>, you'll have to build that yourself.

In any case, I'd add a unit test for the case where a service call throws a FhirOperationException (or more generally any exception).

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/Specification/Terminology/CachingTerminologyService.cs
```diff
@@ -0,0 +1,192 @@
...
+            case FhirString fhirString:
+                hash.Add(fhirString.Value);
+                break;
+            case FhirUri uri:
+                hash.Add(uri.Value);
+                break;
+            case Coding coding:
+                hash.Add(coding.System);
+                hash.Add(coding.Code);
+                hash.Add(coding.Display);
```
ewoutkramer, line 171: Add `coding.Version`.

src/Hl7.Fhir.Base/Specification/Terminology/CachingTerminologyService.cs
```diff
@@ -0,0 +1,192 @@
...
+                break;
+            case CodeableConcept concept:
+                foreach (var coding in concept.Coding)
+                {
+                    hash.Add(coding.System);
+                    hash.Add(coding.Code);
+                    hash.Add(coding.Display);
```
ewoutkramer, line 180: Add `concept.Text`

ewoutkramer (changes requested):

ewoutkramer (approved):

PR #3369, opened 2025-12-11
Remove $findmatches

ewoutkramer:
We introduced $findmatches, but it has been removed from FHIR R6 and … we don't need it after all.

Also, I undid the tuning of Lookup (introducing an `id` parameter), since I'd like to review the use of "id" (truly meaningless in ITerminologyService as it requires the context of a database of resources to match and id to a canonical), much like that I'd like to review (basically remove) `useGet` (as this has nothing to do with ITS).

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ If you want to participate in this project, we're using [Git Flow][nvie] for our
 
 ## Binary Compatibility
 Our build pipeline automatically checks for binary compatibilty and fails when we break it. In some cases we can make an exception to break binary compatitibility. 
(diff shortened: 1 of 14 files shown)


mmsmits (dismissed):

alexzautke (dismissed):

alexzautke:
@ewoutkramer The build is failing

ewoutkramer:
Gosh, I didn't even bother to compile it, it seems....

andrzejskowronski (approved):

PR #3370, opened 2025-12-15
Validate strings for leading/trailing whitespaces

andrzejskowronski:
## Description
We should notify that strings have invalid leading/trailing whitespaces in stricter XML parsing modes.

### DIFF
--- a/src/Hl7.Fhir.Base/Serialization/FhirXmlException.cs
+++ b/src/Hl7.Fhir.Base/Serialization/FhirXmlException.cs
@@ -79,7 +79,7 @@ public class FhirXmlException(
     internal static FhirXmlException ATTRIBUTE_SHOULD_HAVE_BEEN_AN_ELEMENT(XmlReader reader, string instancePath, string attributeName) => Initialize(reader, instancePath, ATTRIBUTE_SHOULD_HAVE_BEEN_AN_ELEMENT_CODE, $"Attribute '{attributeName}' should have been an element with a `value` property, not an attribute.", "Expected element", OO_Sev.Error);
     
     // XML strings do not accept leading or trailing whitespaces and will become trimmed during parsing
-    internal static FhirXmlException STRING_SHOULD_NOT_HAVE_LEADING_TRAILING_WHITESPACE(XmlReader reader, string instancePath, string attributeName) => Initialize(reader, instancePath, STRING_SHOULD_NOT_HAVE_LEADING_OR_TRAILING_WHITESPACE, $"Attribute '{attributeName}' should not contain a leading or trailing whitespaces.", "Invalid whitespace", OO_Sev.Warning);
+    internal static FhirXmlException STRING_SHOULD_NOT_HAVE_LEADING_TRAILING_WHITESPACE(XmlReader reader, string instancePath, string attributeName) => Initialize(reader, instancePath, STRING_SHOULD_NOT_HAVE_LEADING_OR_TRAILING_WHITESPACE, $"Attribute '{attributeName}' should not contain  leading or trailing whitespace.", "Invalid whitespace", OO_Sev.Warning);
 
     /// <summary>
     /// An issue is allowable for backwards compatibility if it could be caused because an older parser encounters data coming from a newer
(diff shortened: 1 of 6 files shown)


ewoutkramer:
But leading spaces are not illegal......  The spec says (https://hl7.org/fhir/datatypes.html#primitive):

> Leading and Trailing whitespace is allowed, but SHOULD be [removed when using the XML format](https://hl7.org/fhir/xml.html#whitespace).

So, this is allowed (just not in XML, but that's a format specific thing). Are we failing to live by this rule somewhere else so that we need to make it consistent?

ewoutkramer (changes requested):
Trailing whitespaces are allowed per the spec, so I do not agree on this change.

andrzejskowronski:
Oh, whoops, seems I misread/misremembered it! And seems tests agree :D
I'll fix it in a sec, seems it's just XML parser that needs to care more about it then! ;)

src/Hl7.Fhir.Base/Serialization/FhirXmlException.cs
```diff
@@ -76,6 +77,9 @@ public class FhirXmlException(
...
     internal static FhirXmlException ELEMENT_SHOULD_HAVE_BEEN_AN_ATTRIBUTE(XmlReader reader, string instancePath, string elementName) => Initialize(reader, instancePath, ELEMENT_SHOULD_HAVE_BEEN_AN_ATTRIBUTE_CODE, $"Element '{elementName}' should have been encoded as an attribute.", "Expected attribute", OO_Sev.Error);
     internal static FhirXmlException ATTRIBUTE_SHOULD_HAVE_BEEN_AN_ELEMENT(XmlReader reader, string instancePath, string attributeName) => Initialize(reader, instancePath, ATTRIBUTE_SHOULD_HAVE_BEEN_AN_ELEMENT_CODE, $"Attribute '{attributeName}' should have been an element with a `value` property, not an attribute.", "Expected element", OO_Sev.Error);
+    
+    // XML strings do not accept leading or trailing whitespaces and will become trimmed during parsing
+    internal static FhirXmlException STRING_SHOULD_NOT_HAVE_LEADING_TRAILING_WHITESPACE(XmlReader reader, string instancePath, string attributeName) => Initialize(reader, instancePath, STRING_SHOULD_NOT_HAVE_LEADING_OR_TRAILING_WHITESPACE, $"Attribute '{attributeName}' should not contain a leading or trailing whitespaces.", "Invalid whitespace", OO_Sev.Warning);
```
ewoutkramer:
```suggestion
    internal static FhirXmlException STRING_SHOULD_NOT_HAVE_LEADING_TRAILING_WHITESPACE(XmlReader reader, string instancePath, string attributeName) => Initialize(reader, instancePath, STRING_SHOULD_NOT_HAVE_LEADING_OR_TRAILING_WHITESPACE, $"Attribute '{attributeName}' should not contain  leading or trailing whitespace.", "Invalid whitespace", OO_Sev.Warning);
```

ewoutkramer (changes requested):

ewoutkramer (approved):

ewoutkramer:
So, in the end, we're still trimming the spaces but also raise an issue (of type warning).

Note that this is not validation - having trailing whitespace is fine, so validating data with such whitespace will not raise an error.  It's just that the *xml* parser is trimming whitespace and raising a warning, as suggested in the spec.

mmreith:
Okay its merged already, but i.m.h.o. Firely shouldn't modify anything automatically. The user should always do that, be he needs a hint (error or warning). But what is about JSON-files? Matthijs told me, that JSON files will get cutted too?

alexzautke:
The trimming already happened before, the addition is only that now it’s also reported back as a warning. Note the FHIR spec mentions that the is SHOULD happen: https://build.fhir.org/xml.html

mmreith:
This behaviour is acceptable for xml, but why JSONs will be treated in the same way (Matthijs told me)? So i'm not sure if this problem really arises "before" packaging?

alexzautke:
This is for XML only. @wardweistra @MatthijsvdWielen is there something happening with JSON on the Simplifier side as well?

andrzejskowronski:
To summarize what will be happening:

1. From a starting point of XML, if serialized to JSON, the strings will be trimmed, **with a warning emitted during parsing of original XML**.
2. From a starting point of JSON, if serialized to XML, the strings will be trimmed, as that is required from XML specification itself. There will be no warnings, as that's the expected, intended, and documented behavior.
3. From a starting point of JSON, if serialized back to JSON, whitespaces will remain in the property values, as there's no constraints there.

Also worth keeping in mind, only the bolded out part is a change, everything else is working the same for a long time now.

PR #3389, opened 2026-01-13
Bump version number

ewoutkramer:


src/firely-net-sdk.props
mmsmits: Can you bump the baseline version to 6.0.2 as well?

mmsmits (changes requested):

mmsmits (dismissed):

mmsmits (dismissed):

mmsmits (approved):

PR #3394, opened 2026-01-20
VONK-9235: Base for strongly typed and pre-validated TerminologyService implementations

andrzejskowronski:
## Description
Provides a base class for standardized parameter/error handling, validation and strongly typed property access in Terminology implementations


src/Hl7.Fhir.Base/Utility/CallerArgumentExpressionAttribute.cs
```diff
@@ -0,0 +1,9 @@
+#if NETSTANDARD2_1
+namespace System.Runtime.CompilerServices;
+
+[AttributeUsage(AttributeTargets.Parameter)]
+internal sealed class CallerArgumentExpressionAttribute(string parameterName) : Attribute
+{
+    public string ParameterName { get; } = parameterName;
```
Kasdejong, line 7: ahhhh we have so many of these now :))

src/Hl7.Fhir.Shims.Base/Specification/Terminology/BaseTerminologyService.Common.cs
```diff
@@ -0,0 +1,171 @@
...
+/// <item><description><b><see cref="ValidateCode(CodeSystem, Code, string?, string?, FhirBoolean?)"/></b> - Core validation logic. Most services implement this.</description></item>
+/// <item><description><b><see cref="ResolveValueSet(Canonical)"/></b> - Required for ValueSet operations. Must return a ValueSet or throw.</description></item>
+/// <item><description><b><see cref="ResolveCodeSystem(Canonical)"/></b> - Required for CodeSystem operations. Must return a CodeSystem or throw.</description></item>
+/// <item><description><see cref="Expand(ExpandParameters)"/> - For ValueSet expansion support</description></item>
+/// <item><description><see cref="BuildLookupResult(CodeSystem, CodeSystem.ConceptDefinitionComponent)"/> - For code system lookup operations</description></item>
+/// <item><description><see cref="Translate(TranslateParameters)"/> - For concept map translations</description></item>
+/// <item><description><see cref="Closure(ClosureParameters)"/> - For closure table maintenance (rarely needed)</description></item>
+/// <item><description><see cref="Subsumes(SubsumesParameters)"/> - For subsumption testing</description></item>
```
Kasdejong, line 128: I guess now that we have a default implementation of subsumption, part of this comment is obsolete.

src/Hl7.Fhir.Shims.Base/Specification/Terminology/CustomValueSetTerminologyService.cs
Kasdejong, line 1: massive kudos, this is exactly what I wanted to achieve eventually :)

src/Hl7.Fhir.Specification.Shared.Tests/Terminology/LocalTerminologyServiceTests.cs
Kasdejong, line 1: should we document/treat the new requirements for terminology operations as a behavioural breaking change? I get that us not requiring them before was technically a bug according to the FHIR specification, but this might break a lot of existing assumptions by third parties using our code, right?

I think it leads to enough potential errors to make SDK 6.1

Kasdejong (approved):
Should we merge this now or wait until you have finished on FS?

src/Hl7.Fhir.Base/Utility/CallerArgumentExpressionAttribute.cs
andrzejskowronski: True, I'm thinking it might be simpler if we'll include some polyfills package instead, but that's for another issue ;)

src/Hl7.Fhir.Specification.Shared.Tests/Terminology/LocalTerminologyServiceTests.cs
andrzejskowronski: Not sure if there's anything required that wasn't before, more our tests didn't complain because we weren't reaching the code path where that would be checked, and now we verify first, run logic later.
That said I think 6.1 indeed makes sense :D

src/Hl7.Fhir.Shims.Base/Specification/Terminology/BaseTerminologyService.Common.cs
andrzejskowronski: They can actually still override it if they prefer different behavior, so I think it still applies, but that's why I moved it all the way to the bottom :D

PR #3397, opened 2026-01-21
Support null focus on fhirpath operations

### LINKED ISSUE #3390: FhirPath Expression without a PocoNode
given an expression like: `@2020-01-01 - 365 days` previously it was possible to: 
```cs
var fpCompiler = new FhirPathCompiler(
        new SymbolTable()
            .AddStandardFP()
            .AddFhirExtensions()
    );
var exp = fpCompiler.Compile("@2020-01-01 - 365 days");
var res = exp(null!, new EvaluationContext()).FirstOrDefault();
```

but with newer versions, you get an exception that the node parameter is null. Not sure where exactly the change was made.

I want to call the compiled expression with null

or

provider a different method or Compile overload that will return a different delegate that doesn't need a node.

supplying something like `exp(new Patient().ToPocoNode(), new EvaluationContext())` works, but not ideal.

**Additional context**
n/a

Kasdejong:
## Description
Make sure we support null focus on FP operations. We supported this before, so it makes sense for us to support it now.

Tested by editing some of the tests which used dummies to now use `null!` instead.

Note: we default to [] now, which is more correct.

## Related issues
Fixes #3390


andrzejskowronski (changes requested):
Code itself seems fine, but I think it might be too big of a breaking change to public surface

Kasdejong:
@andrzejskowronski I'm hoping it will not be source-breaking for most people, since the type in the old signature implements the type in the new signature.

src/Hl7.Fhir.Base/FhirPath/EvaluationContext.cs
```diff
@@ -48,12 +48,12 @@ public EvaluationContext(PocoNode? resource, PocoNode? rootResource, IDictionary
     /// <summary>
     /// The data represented by <c>%rootResource</c>.
     /// </summary>
-    public PocoNode? RootResource { get; set; }
+    public IEnumerable<PocoNode>? RootResource { get; set; }
```
ewoutkramer: This is a breaking change for the `get`, right?

ewoutkramer (dismissed):

src/Hl7.Fhir.Base/FhirPath/EvaluationContext.cs
Kasdejong: You're right. I'll revise

andrzejskowronski (dismissed):

andrzejskowronski (approved):

Issue #3053, opened 2025-02-20
Improve error messages from ZipSource


ewoutkramer:
Once SDK 6.0 and especially #2778 is done, we can add extra checks to the ZipSource to detect empty directories, failed unpacks etc, so we can finally fix one of the most common support issues in the SDK.

ewoutkramer:
I don't think these are breaking changes, just better use of the new infrastructure, so I have moved this out of the "need to do this in SDK 6.0 timeframe".

Narnian12:
I have been discussing this issue with Ewout, and Ewout has been greatly helpful in helping me walk through the code base alongside the issue. Here is a summary of my analysis and proposal before I work on building a PR for this.

## Problem Analysis

I believe one place where potential exceptions have been hiding is in `ZipCacher.Refresh` (lines 110-117). Here is a code snippet of the try-catch:

```csharp
try
{
  ZipFile.ExtractToDirectory(ZipPath, dir.FullName);
}
catch
{
  Clear();
}
```

The `Clear()` function clears the cache, but the catch does not propagate any exceptions upward. `ZipFile.ExtractToDirectory` [can throw many kinds of exceptions](https://learn.microsoft.com/en-us/dotnet/api/system.io.compression.zipfile.extracttodirectory?view=net-10.0), and I suspect that method has the highest likelihood of throwing helpful exceptions.

## Proposed Changes

- Remove the try/catch from the `ZipCacher.Refresh()` function altogether
    - The Refresh method calls `ZipFile.ExtractToDirectory()`, which appears to be the method that should throw the most exceptions - a cursory look has it potentially throwing 10+ different types of exceptions
    - This should stop swallowing all the exceptions that `ExtractToDirectory()` might be throwing - originally, when an exception was thrown, the catch statement just cleared the cache, but did not rethrow
- Add a try/catch to the CommonZipSource.createSource() function
    - This function calls the `ZipCacher.Refresh()` method, so the original try/catch should basically get moved up a level
    - This try/catch will include a simple retry - if the `ZipCacher.Refresh()` function throws an exception, the catch statement will execute the Refresh method again. If it throws again, it will propagate the exception upward

I would appreciate any feedback on this analysis! Thank you in advance - this is my first attempt at contributing to Firely, but I hope it will not be my last :)

Narnian12:
Hopefully #3399 will resolve most of this, but can definitely see it in action first.

PR #3399, opened 2026-01-24
Add ZipSource Exception and Retry Logic

Narnian12:
## Description
A more detailed summary of the issue can be found [here](https://github.com/FirelyTeam/firely-net-sdk/issues/3053#issuecomment-3779250824).

* `ZipCacher.Refresh()` will no longer swallow exceptions
    * The try/catch was removed surrounding the `getCachedZipDirectory()` call
* Updated `ZipCacher.GetContentDirectory()`
    * The line `if (!IsActual()) Refresh()` is equivalent to calling `EnsureActual()`
* `CommonZipSource.createSource()` now contains try/catch and retry logic
    * The original try/catch was essentially moved up a level, from `ZipCacher` to `CommonZipSource`
    * If an exception is thrown, the method retries directory extraction
    * If the retry fails, the exception will propagate upward
 * Added `createDirectorySource` helper method in `CommonZipSource`

## Related issues
Improves error messaging for #3053

## Testing
* One unit test checks that when the cache is cleared, the method will retry
* One unit test checks that when a zip file is corrupted, the retry will fail, and an exception will propagate upward

## FirelyTeam Checklist
- [X] **Update the title** of the PR to be succinct and less than 50 characters
- [X] Mark the PR with the label **breaking change** when this PR introduces breaking changes


Narnian12:
Let me know if this should be a breaking change. Now that the try/catch is being removed from `ZipCacher`, it is possible that the [ZipFIle.ExtractToDirectory()](https://learn.microsoft.com/en-us/dotnet/api/system.io.compression.zipfile.extracttodirectory?view=net-10.0) method will throw exceptions now, instead of the try/catch clearing the cache and swallowing the exceptions.

src/Hl7.Fhir.Conformance/Specification/Source/CommonZipSource.cs
```diff
@@ -189,7 +192,26 @@ private CommonDirectorySource createSource()
...
             var zc = new ZipCacher(ZipPath, CacheDirectory);
-            var source = new CommonDirectorySource(_inspector, zc.GetContentDirectory(), _settings);
+
+            try
+            {
+                return createDirectorySource(zc);
+            }
+            catch
+            {
+                zc.Clear();
```
andrzejskowronski: Refresh also calls Clear, so it's not necessary

andrzejskowronski (changes requested):
The CommonZipSource has a separate implementation for STU3 at Hl7.Fhir.STU3 > Specification.Source.ZipSource, which would also need those changes then.

andrzejskowronski (approved):

andrzejskowronski:
Thank you for the contribution!

Narnian12:
Thank you! Happy to contribute :)

PR #3407, opened 2026-02-05
Use PropertyMapping.Name for MemberName if possible to report FHIR properties rather than FHIRPath names

andrzejskowronski:


ewoutkramer:
@copilot - fix the test failures.

ewoutkramer:
Discussion: if the element does not match an existing POCO element, should the `MemberName` be `null` or just the name as found in the deserialization (or overflow)?  I tend to say this should be `null` as MemberName is the name of the POCO element, and if that does not exist, it should be null. But have we documented `MemberName` as such?

@copilot - what would you say?

andrzejskowronski:
> But have we documented `MemberName` as such?

I don't think we have any documentation on it, as original source of the attribute was MS property validation framework.

ewoutkramer:
Right, then I think the current approach is fine.

ewoutkramer (approved):

PR #3430, opened 2026-02-12
Made BaseFhirClient thread safe

JTeeuwissen:
## Description
Updated `extractResourceFromHttpResponse` logic in `BaseFhirClient` to not return values based on the `Last...` result, but instead to use the result directly. To prevent race conditions from parallel queries. 

## Related issues
- None

## Testing
- Not tested

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Base/Rest/BaseFhirClient.cs
+++ b/src/Hl7.Fhir.Base/Rest/BaseFhirClient.cs
@@ -1,4 +1,4 @@
-﻿#nullable enable
+#nullable enable
 
 /* 
  * Copyright (c) 2023, Firely (info@fire.ly) and contributors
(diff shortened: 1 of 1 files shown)


JTeeuwissen:
Is there an indication as to when these changes are released if this pr is merged?

ewoutkramer:
Note that the FhirClient is *not* designed to be used concurrently, and there are probably more places where this will cause problems.

See our documentation here https://docs.fire.ly/projects/Firely-NET-SDK/en/latest/client/setup.html

>Note: the FhirClient is not thread-safe, so you will need to create one for each thread, if necessary. But don’t worry: creating an  instance of a FhirClient is cheap, the connection will not be opened until you start working with it.

ewoutkramer:
Anyway, this looks like a reasonable change, so I'll pull it.

ewoutkramer (approved):

PR #3446, opened 2026-02-18
Add per-request HTTP header callback to FhirClient (#3298)

### LINKED ISSUE #3298: (non default) Request Headers
Currently, whenever we want to modify the headers for a http request, we have to reset and then set the `FhirClient.RequestHeaders`. But these headers (`HttpClient.DefaultRequestHeaders`) are not mend for frequent manipulation and they are not thread-safe.

I'd like each `BaseFhirClient` method to have an overload (or optional parameter, but this would be a breaking change) with a `HttpRequestHeaders` parameter. Which can be passed to
`ToHttpRequestMessage` where it can be used to add headers to the `HttpRequestMessage`

```cs
var message = new HttpRequestMessage(method, uri);
if (headers is not null)
{
    foreach (var httpRequestHeader in headers)
    {
        message.Headers.Add(httpRequestHeader.Key, httpRequestHeader.Value);
    }
}
```

Another option would be using http client middleware. But this doesn't play nice with scoping and doesn't allow for local values to be passed in the headers.

ewoutkramer:
## Summary

Closes #3298.

`FhirClient.RequestHeaders` exposes `HttpClient.DefaultRequestHeaders`, which persists across all requests and is not thread-safe. This PR adds an optional `Action<HttpRequestHeaders>? requestHeaders = null` callback as the last parameter on every public method of `BaseFhirClient` and `FhirClientSearch`. The callback is applied to the per-request `HttpRequestMessage` just before it is sent, leaving `DefaultRequestHeaders` untouched and scoped to a single call.

**Example usage:**
\`\`\`csharp
var patient = await client.ReadAsync<Patient>("Patient/1",
    requestHeaders: h => h.Authorization = new AuthenticationHeaderValue("Bearer", token));
\`\`\`

## Changes

- `BaseFhirClient.cs`: Added `requestHeaders` parameter to both private `executeAsync` overloads (applied after `ToHttpRequestMessage`/before `ExecuteAsync`) and to all ~30 public methods plus private helpers `internalUpdateAsync`, `internalHistoryAsync`, `internalOperationAsync`.
- `FhirClientSearch.cs`: Same pattern for all search and paging methods.
- Parameter is placed after `ct` to avoid breaking existing callers using positional arguments.
- XML doc `<param name="requestHeaders">` tag added to every method that had existing XML documentation.

## Test plan

- [ ] `dotnet build src/Hl7.Fhir.Base` — clean, 0 warnings, 0 errors
- [ ] Existing unit tests pass (`dotnet test` excluding `IntegrationTest` and `LongRunner` categories)
- [ ] Verify existing callers with positional/named args still compile unchanged
- [ ] Spot-check: `client.ReadAsync<Patient>("1", requestHeaders: h => h.Authorization = new AuthenticationHeaderValue("Bearer", "tok"))` compiles and the header appears on the outgoing request

### DIFF
--- a/src/Hl7.Fhir.Base/Rest/BaseFhirClient.cs
+++ b/src/Hl7.Fhir.Base/Rest/BaseFhirClient.cs
@@ -1,6 +1,6 @@
 #nullable enable
 
-/* 
+/*
  * Copyright (c) 2023, Firely (info@fire.ly) and contributors
  * See the file CONTRIBUTORS for details.
  * 
@@ -152,6 +152,7 @@ public partial class BaseFhirClient : IDisposable
     /// <param name="ifNoneMatch">The (weak) ETag to use in a conditional read. Optional.</param>
     /// <param name="ifModifiedSince">Last modified since date in a conditional read. Optional. (refer to spec 2.1.0.5) If this is used, the client will throw an exception you need</param>
     /// <param name="ct" />
+    /// <param name="requestHeaders">Optional. Callback to set additional per-request HTTP headers on the outgoing request.</param>
     /// <typeparam name="TResource">The type of resource to read. Resource or DomainResource is allowed if exact type is unknown</typeparam>
     /// <returns>
     /// The requested resource. This operation will throw an exception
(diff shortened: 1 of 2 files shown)


ewoutkramer:
It's quite a big change, since all methods required new optional parameters.  I really hope we can do this as a parameter to the constructor instead.

ewoutkramer:
I think this is mostly relevant when using a FhirClient in a multi-threaded manner, which is not supported. FhirClients are cheap to create and this is the preferred pattern.

PR #3463, opened 2026-03-10
Merge support/sdk5 into develop (SDK5 → SDK6 sweep)

ewoutkramer:
## Summary

Periodic sweep to merge SDK5 (`support/sdk5`) changes into SDK6 (`develop`), up to release v5.13.3 (commit `5d5dbe619`).

### Background: Messy merge history

A few months ago, `develop` was accidentally merged into `support/sdk5` (commit `aa33d7353`), then immediately reverted (`817e7cb6`). This left a noisy common ancestor that makes a standard `git merge` impractical (the 3-way merge base predates those commits, causing thousands of spurious conflicts).

### Approach

To avoid re-introducing the noise from the accident/revert:
- Used `git merge -s ours` to record the merge point cleanly, keeping SDK6 content as authoritative
- Manually applied only the SDK5 changes that were **not yet present** in SDK6

### Changes applied from SDK5

Most SDK5 post-revert changes were already in SDK6 (cherry-picked previously):
- `ModelInspector` assembly attribute caching + `ValidateSatelliteAssemblyCompatibility`
- `IExtendable.RemoveExtensions` helper method
- `SnapshotGeneratorExtensions` non-inheritable extensions helpers + `HashSet` change
- `SnapshotGenerator` (Conformance + STU3) call to `RemoveAllNonInheritableExtensions` in `ExpandElementAsync`
- `AssemblyInfo` files with version-specific `FhirModelAssembly` attributes
- Snapshot generator test assertions

The one change **not yet in SDK6** that is applied here:

1. **`build/azure-pipelines.yml`** — Add `support/*` branches to CI trigger (from SDK5 PR #3455)

### Not forwarded

- **`ModelInfo.cs`** — The SDK5 PR #3374/#3423 removed the `Lazy<ModelInspector>` wrapper to fix a problematic lazy-initialization issue on SDK5. However, in SDK6 the `Lazy<>` approach is fine: `ModelInspector.ForAssembly()` already uses a `ConcurrentDictionary` for caching, and version mixing is no longer possible. Keeping the `Lazy<>` in SDK6 is correct; the SDK5-specific fix does not need to be forwarded.

### Future sweeps

Because this merge records `support/sdk5` tip (`5d5dbe619`) as merged into `develop`, the next SDK5→SDK6 sweep can use a standard `git merge` starting from this point, without the accidental-merge history interfering.

## Test plan

- [ ] CI pipeline passes on this branch
- [ ] Verify `build/azure-pipelines.yml` trigger includes `support/*`


src/Hl7.Fhir.Shims.STU3AndUp/Model/ModelInfo.cs
```diff
@@ -196,24 +196,25 @@ public static bool IsInstanceTypeFor(FHIRAllTypes superclass, FHIRAllTypes subcl
...
-            }
-            return inspector;
-        });
-
         /// <summary>
         /// Gets the <see cref="ModelInspector"/> providing metadata for the resources and
         /// datatypes in this release of FHIR.
         /// </summary>
-        public static ModelInspector ModelInspector => _modelInspector.Value;
+        public static ModelInspector ModelInspector
```
ewoutkramer: It's prudent of Claude to include this from SDK5, but do we really want this?  Given that you can no longer incorrectly mix versions in ModelInspector, this `Lazy<>` should actually work, right?

andrzejskowronski: Yep, definitely lazy should be fine

andrzejskowronski: Also, shouldn't this be the same between 5 and 6 too?

andrzejskowronski (changes requested):

andrzejskowronski (approved):

PR #3472, opened 2026-03-24
Add direct ISourceNode.ToPoco() implementation

cknaap:
## Description

This pull request introduces a direct conversion from `ISourceNode` to `Poco`. For this, it refactors the implementation of the `NewPocoBuilder` for deserializing FHIR resources from `ISourceNode` (new) and `ITypedElement` (existing) trees, and updates the deserialization benchmarks to use and compare this new approach. The changes improve the flexibility, maintainability, and correctness of POCO construction from various node types, and add more comprehensive benchmarks for deserialization strategies.

**Core NewPocoBuilder Implementation:**

- Added a new partial, internal implementation of `NewPocoBuilder`, split into `NewPocoBuilder.Shared.cs`, `NewPocoBuilder.SourceNode.cs`, and `NewPocoBuilder.TypedElement.cs`, providing robust logic for constructing POCOs from both `ISourceNode` and `ITypedElement` trees, including dynamic type resolution, property assignment, and primitive normalization.
- Fixed a bug in `classMappingForElement` (`NewPocoBuilder.SourceNode.cs`) where `propertyMapping.ImplementingType` was used to look up the class mapping. For properties with an abstract `ImplementingType` but a single concrete `FhirType` (e.g. `Signature.who` in R4, typed as `DataType` but always holding a `Reference`), the code incorrectly fell through to `DynamicDataType`. The fix uses `propertyMapping.GetInstantiableType()` instead, which resolves to the single concrete type when only one is allowed. This fixes the `RoundtripSignature.WorksWithTypedElementSerializers` test failure that surfaced after the initial changes were merged.

**Refactoring and API Improvements:**

- Changed `NewPocoBuilder` from a single file to a partial class, encapsulating logic and making the constructor public for broader usability. Also removed unused usings and cleaned up the main file.

**Benchmark Enhancements:**

- Updated `DeserializationBenchmark.cs` to add new benchmark methods for the new POCO builder, comparing direct and bridge deserialization for both JSON and XML. Removed the `XmlReader` class field (which caused a name collision with `System.Xml.XmlReader`) and replaced it with a local `using var` declaration inside `XmlDictionaryDeserializer()`, also ensuring proper disposal.

**General Code Quality:**

- Improved naming conventions for internal variables and removed unused imports for clarity and maintainability.

These changes collectively modernize and enhance the deserialization pipeline, making it easier to extend and reason about, and providing a strong foundation for future improvements.

## Testing

[...]

PR #3489, opened 2026-04-23
Support for custom resources in `ModelInspector` built from `IStructureDefinitionSummary`

### LINKED ISSUE #2891: Add functionality to construct ClassMappings from IStructureDefinitionSummary for custom resources

andrzejskowronski:
## Summary
- add support for importing `ClassMapping` metadata directly from `IStructureDefinitionSummary`, so custom resources and datatypes can be registered in `ModelInspector` without reflected POCO types
- build `PropertyMapping` instances from summary element metadata, including nested backbone elements, collection metadata, and choice typing
- register custom summary-backed mappings by name/canonical without overwriting the shared runtime-type index for `DynamicResource` and `DynamicDataType`

## Additional fixes
- preserve canonical lookup when a mapping is imported by name first and later imported with a canonical
- resolve summary property references by canonical using absolute URI detection instead of the previous slash-based heuristic, so `urn:` identifiers work too

## Testing
- added unit coverage for importing custom class mappings from synthetic `IStructureDefinitionSummary` inputs
- added integration coverage for building a mapping from a real custom resource `StructureDefinition`
- added regression coverage for delayed canonical registration and `urn:` canonical resolution

## Out of scope
- NewPocoBuilder needs to be reviewed for any changes needed, would be good to have #3472 merged before too

Closes #2891

### DIFF
--- a/src/Hl7.Fhir.Base/Introspection/ClassMappingCollection.cs
+++ b/src/Hl7.Fhir.Base/Introspection/ClassMappingCollection.cs
@@ -41,7 +41,7 @@ internal class ClassMappingCollection : ICollection<ClassMapping>
         // Multiple custom resources/backbones may share the same runtime type
         // (e.g. DynamicResource / DynamicDataType), so indexing them by type would
         // make later imports overwrite earlier ones and would make type-based lookup
-        // ambiguous. Will be solved in follow up task.
+        // ambiguous.
         if (!mapping.IsCustomMapping)
             _byType[mapping.NativeType] = mapping;
 
--- a/src/Hl7.Fhir.Base/Introspection/ClassMappingCollection.cs
+++ b/src/Hl7.Fhir.Base/Introspection/ClassMappingCollection.cs
@@ -84,8 +84,11 @@ internal class ClassMappingCollection : ICollection<ClassMapping>
                 _byType.TryRemove(kvp.Key, out _);
         }
 
-        if (item.Canonical is not null)
-            _byCanonical.TryRemove(item.Canonical, out _);
+        foreach (var kvp in _byCanonical)
+        {
+            if (ReferenceEquals(kvp.Value, item))
+                _byCanonical.TryRemove(kvp.Key, out _);
+        }
 
         return true;
     }
@@ -115,6 +118,17 @@ internal class ClassMappingCollection : ICollection<ClassMapping>
     public void RegisterTypeAlias(Type type, ClassMapping mapping)
     {
         _byType[type] = mapping;
+    }
+
+    /// <summary>
+    /// Registers a canonical alias for an existing <see cref="ClassMapping"/>.
+    /// Unlike <see cref="Add"/>, this only updates the canonical lookup and does not affect the
+    /// name or type dictionaries. This is used when a mapping is first imported by name and later
+    /// associated with a canonical identifier.
+    /// </summary>
+    public void RegisterCanonicalAlias(string canonical, ClassMapping mapping)
+    {
+        _byCanonical[canonical] = mapping;
     }
 
     /// <summary>
--- a/src/Hl7.Fhir.Base/Introspection/ClassMappingCollection.cs
+++ b/src/Hl7.Fhir.Base/Introspection/ClassMappingCollection.cs
@@ -37,7 +37,13 @@ internal class ClassMappingCollection : ICollection<ClassMapping>
         var propKey = mapping.Name;
         _byName[propKey] = mapping;
 
-        _byType[mapping.NativeType] = mapping;
+        // Custom mappings intentionally do not participate in the type index.
+        // Multiple custom resources/backbones may share the same runtime type
+        // (e.g. DynamicResource / DynamicDataType), so indexing them by type would
+        // make later imports overwrite earlier ones and would make type-based lookup
+        // ambiguous. Will be solved in follow up task.
+        if (!mapping.IsCustomMapping)
+            _byType[mapping.NativeType] = mapping;
 
         var canonical = mapping.Canonical;
         if (canonical is not null)
--- a/src/Hl7.Fhir.Base/Introspection/ClassMappingCollection.cs
+++ b/src/Hl7.Fhir.Base/Introspection/ClassMappingCollection.cs
@@ -84,11 +84,8 @@ internal class ClassMappingCollection : ICollection<ClassMapping>
                 _byType.TryRemove(kvp.Key, out _);
         }
 
-        foreach (var kvp in _byCanonical)
-        {
-            if (ReferenceEquals(kvp.Value, item))
-                _byCanonical.TryRemove(kvp.Key, out _);
-        }
+        if (item.Canonical is not null)
+            _byCanonical.TryRemove(item.Canonical, out _);
 
         return true;
     }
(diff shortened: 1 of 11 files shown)


src/Hl7.Fhir.Base/Introspection/PropertyMapping.cs
```diff
@@ -279,6 +296,95 @@ public static bool TryCreate(PropertyInfo prop, [NotNullWhen(true)] out Property
...
+        }
+
+        ResolvedTypeInfo importReferencedType(IStructureDefinitionReference reference)
+        {
+            var mapping = resolveExistingMapping(reference.ReferredType);
+            return new ResolvedTypeInfo(mapping, new PocoTypeReferenceInfo(reference.ReferredType));
+        }
+
+        ClassMapping resolveExistingMapping(string typeName) =>
+            (typeName.Contains('/')
```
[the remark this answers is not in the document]
andrzejskowronski: We probably don't care about urn canonicals, but changed the check to the Uri.TryCreate

src/Hl7.Fhir.Base/Introspection/ModelInspector.cs
```diff
@@ -245,6 +245,32 @@ public IReadOnlyList<ClassMapping> Import(Assembly assembly)
...
+    {
+        if (summary is null) throw Error.ArgumentNull(nameof(summary));
+        if (mappingName is null) throw Error.ArgumentNull(nameof(mappingName));
+
+        if (canonical is not null && FindClassMappingByCanonical(canonical) is { } existingByCanonical)
+            return existingByCanonical;
+
+        if (FindClassMapping(mappingName) is { } existingByName)
+            return existingByName;
+
```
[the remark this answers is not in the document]
andrzejskowronski: Somewhat valid, added logic to link the canonical to the mapping

src/Hl7.Fhir.Base/Introspection/ModelInspector.cs
```diff
@@ -315,13 +323,31 @@ private void extractBackbonesFromClasses(IEnumerable<Type> classTypes)
...
     /// <remarks>The search for the mapping by namem is case-insensitive.</remarks>
     public ClassMapping? FindClassMapping(string fhirTypeName) =>
         _classMappings.ByName.GetValueOrDefault(fhirTypeName);
+    
+    /// <summary>
+    /// Retrieves an already imported <see cref="ClassMapping" /> given an instance of a FHIR type.
+    /// </summary>
+    public ClassMapping? FindClassMapping(Base instance)
+    {
+        if(instance is IDynamicType { DynamicTypeName: {} typeName} && !string.IsNullOrWhiteSpace(typeName))
```
ewoutkramer, line 346: Duplication of the logic in line 207 - I think we need to factor this private function differently.

src/Hl7.Fhir.Base/Introspection/ClassMappingCollection.cs
```diff
@@ -111,6 +120,17 @@ public void RegisterTypeAlias(Type type, ClassMapping mapping)
         _byType[type] = mapping;
     }
 
+    /// <summary>
+    /// Registers a canonical alias for an existing <see cref="ClassMapping"/>.
+    /// Unlike <see cref="Add"/>, this only updates the canonical lookup and does not affect the
+    /// name or type dictionaries. This is used when a mapping is first imported by name and later
+    /// associated with a canonical identifier.
+    /// </summary>
+    public void RegisterCanonicalAlias(string canonical, ClassMapping mapping)
```
ewoutkramer: This can go.

src/Hl7.Fhir.Base/Introspection/ClassMappingCollection.cs
```diff
@@ -78,8 +84,11 @@ public bool Remove(ClassMapping item)
                 _byType.TryRemove(kvp.Key, out _);
         }
 
-        if (item.Canonical is not null)
-            _byCanonical.TryRemove(item.Canonical, out _);
+        foreach (var kvp in _byCanonical)
```
ewoutkramer: This can be reverted to its old state.

src/Hl7.Fhir.Base/Introspection/ModelInspector.cs
```diff
@@ -372,10 +398,13 @@ private void extractBackbonesFromClasses(IEnumerable<Type> classTypes)
 
     /// <inheritdoc cref="IStructureDefinitionSummaryProvider.Provide(string)"/>
     public IStructureDefinitionSummary? Provide(string canonical) =>
-        canonical.Contains('/') ?
+        isCanonical(canonical) ?
```
ewoutkramer, line 401: More watertight, but 20x more expensive. Unnecessary.

src/Hl7.Fhir.Base/Introspection/PropertyMapping.cs
```diff
@@ -71,6 +71,22 @@ public PropertyMapping(ClassMapping declaringClass, string name, Type propertyTy
         _propertyType = propertyType;
     }
 
+    /// <summary>
+    /// Creates a PropertyMapping from pre-computed type information, typically when building
+    /// mappings from a <see cref="Hl7.Fhir.Specification.IElementDefinitionSummary"/>.
+    /// </summary>
+    [SetsRequiredMembers]
+    public PropertyMapping(ClassMapping declaringClass, string name, Type propertyType, ClassMapping propertyTypeMapping,
```
ewoutkramer, line 79: So, the propertyType here can only be a type that implements IDynamicType, right? Otherwise you would not need this overload?

ewoutkramer: No, it can be List<DynamicType?> - so this is the actual property type.

ewoutkramer, line 79: This constructor is too intimate, it should only have `Type` (for the actual propertytype), `IEnumerable<ClassMapping>` for the type/choice types as extra parameters - no external user should be obliged to pass in an `IEnumerable<ITypeSerializationInfo>` - these classes are public, but really they are infrastructure so ClassMapping was able to implement IStructureDefinitionSummary. Also the fact that the constructor sets _type is a smell since this is just a backup field meant as a cache. A constructor that requires half "internal" data to be passed in to set into a cached field is definitely not what we want to do here.

You can also see that in the code in FS that actually calls this function - it goes through hoops to create instances of classes implementing ITypeSerializationInfo to pass in stuff, basically spreading knowledge about how ModelInspector implements IStructureDefinsitionSummary across the two codebases.

TL;DR: The methods on ModelInspector, ClassMapping and PropertyMapping should be in terms of Type, ClassMapping and PropertyMapping. Even though they implement IStructureDefinsitionSummary-related interfaces, this is not what the domain language for these classes is about.

src/Hl7.Fhir.Support.Tests/Introspection/ModelInspectorTest.cs
```diff
@@ -14,6 +14,9 @@
 using Hl7.Fhir.Specification;
```
ewoutkramer, line 14: The tests here should focus on creating a ClassMapping for:

* A custom datatype
* A custom resource, which has both a backbone and an element that has a choice for which one of the choices is this custom datatype.

Then, we should test whether this returns the right information via IStructureDefinitionSummary.

Also, we should create an instance of this type in json (and xml), then parse it and see whether the types are all correct. Then serialize it back out and check whether the roundtrip worked.

ewoutkramer (changes requested):

src/Hl7.Fhir.ElementModel.Shared.Tests/ElementNodeTests.cs
```diff
@@ -120,6 +120,39 @@ public void TestFpNavigateCustomResource()
             Assert.IsFalse((bool?)result.FirstOrDefault().Value);
         }
 
+        [TestMethod]
+        public async Tasks.Task CanBuildClassMappingFromCustomResourceStructureDefinition()
```
ewoutkramer, line 124: This should move to Firely Server and test given a complex ISDSummary (see my example in the other test) whether the new extension methods generate the right ClassMappings.

ewoutkramer (changes requested):

ewoutkramer:
Replaced by #3519.

PR #3493, opened 2026-04-29
 Fix OverflowException when parsing large integer values in JSON

### LINKED ISSUE #3457: OverflowException when parsing large valueQuantity.value due to unconditional long -> int32 conversion in NormalizeValue()
**Describe the bug**
The `NormalizeValue()` method in `FhirJsonNode.cs` unconditionally attempts to convert all `long` values to `int32` without checking whether the value fits within the `int32` range.

This results in an `OverflowException` when parsing FHIR JSON data containing large numeric values in fields that are defined as `decimal` (`FhirDecimal`), such as `valueQuantity.value`.

The issue appears to originate here:
https://github.com/FirelyTeam/firely-net-sdk/blob/9a6e857a6ba91c52566b0d5b24365c6d64315a6c/src/Hl7.Fhir.Base/Serialization/elementmodel/FhirJsonNode.cs#L340-L355

**To Reproduce**

1. Use the `FhirClient` to request `Observation` resources from a FHIR server.
2. The server returns an `Observation` where `valueQuantity.value` contains a numeric value larger than int.MaxValue (2,147,483,647).

Example JSON returned by the server:
``` json
{
  "resourceType": "Observation",
  "valueQuantity": {
    "value": 20231128235900
  }
}
```
3. Retrieve the resource using `TypeOperationAsync`.
4. During parsing of the returned JSON, `NormalizeValue()` attempts to convert the numeric value to `int32`.

**Actual behavior**
An `OverflowException` is thrown because the value exceeds the `int32` range.

**Versions used:**
- FHIR Version: STU3
- FHIR .NET SDK Version: 6.1.0
- .NET Version: .NET 10

MadhukarJukanti:
Fixes #3457

  ## Description
  `FhirJsonNode.NormalizeValue` was calling `Convert.ToInt32()` unconditionally
  on all `long` JSON integers, causing an `OverflowException` for values outside
  the Int32 range (e.g. timestamps like `20231128235900` used as `valueQuantity.value`).

  The fix replaces the try/catch with a range check. Values that fit in Int32 are
  still normalized to `int` for correct JSON serialization. Values outside the
  range are now preserved as `long` without raising an error.

  ## Related issues
  Fixes #3457

  ## Testing
  - Added `FhirJsonParser_LargeIntegerValueQuantity_DoesNotThrow` — regression test for the exact overflow scenario
  - Added `FhirJsonParser_SmallIntegerValue_NormalizesToInt32` — confirms existing int32 normalization is preserved

  ## FirelyTeam Checklist


alexzautke:
Fixed by switching to the new parsing stack in the FHIRClient: https://github.com/FirelyTeam/firely-net-sdk/pull/3480

Issue #3494, opened 2026-04-30
Wrong TypeName for ValueSet.ConceptSetComponent


andrzejskowronski:
**Describe the bug**
`ValueSet.ConceptSetComponent` has `TypeName` = `ValueSet.compose.include`, but it's also used for `ValueSet.compose.exclude`. We should fix our codegen to differentiate those.

memoryfraction:
Hi @FirelyTeam, I submitted this PR two weeks ago and wanted to check if any feedback or additional information is needed from my side: https://github.com/FirelyTeam/firely-net-sdk/pull/3505
Thank you!

alexzautke:
Hi @memoryfraction, thank you for the contribution. Unfortunately it needs to be fixed in the code generation part of the SDK and can’t be set manually. See https://github.com/FirelyTeam/fhir-codegen. We are not able to pull this particular PR.

ewoutkramer:
I don't agree. See my comments on #3505. The type is being reused, on a different path. That does not mean that the name of the type changes.

PR #3505, opened 2026-05-21
fix: Correct TypeName for ValueSet.ConceptSetComponent in include/exc…

memoryfraction:
## Description
Previously, ValueSet.ConceptSetComponent always returned "ValueSet.compose.include" as its TypeName, even when used in the exclude context. This fix ensures the TypeName dynamically reflects the actual usage context:
- Returns "ValueSet.compose.include" when used in Compose.Include
- Returns "ValueSet.compose.exclude" when used in Compose.Exclude

Changes:
- Add _typeName field and SetTypeName() method to ConceptSetComponent
- Update Include/Exclude property getters and setters to automatically set correct TypeName
- Ensure TypeName is updated when components are added, removed, or moved between lists

## Related issues
Resolves [issue [3494](https://github.com/FirelyTeam/firely-net-sdk/issues/3494)].

Testing:
- Add 7 comprehensive unit tests covering all TypeName scenarios
- All tests pass successfully (7/7)
- Backward compatible - default behavior unchanged

Notes:
This is a pragmatic fix within the current code generation structure. For long-term improvement, consider modifying the code generator to create separate IncludeComponent and ExcludeComponent classes for stronger type safety.

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Conformance/Model/Generated/ValueSet.cs
+++ b/src/Hl7.Fhir.Conformance/Model/Generated/ValueSet.cs
@@ -174,7 +174,13 @@ namespace Hl7.Fhir.Model
         {
           if(_Include.InOverflow<List<Hl7.Fhir.Model.ValueSet.ConceptSetComponent>>())
             throw CodedValidationException.FromTypes(typeof(List<Hl7.Fhir.Model.ValueSet.ConceptSetComponent>), Overflow["include"]);
-          return _Include ??= [];
+          var list = _Include ??= [];
+          // Ensure all components in the list have the correct TypeName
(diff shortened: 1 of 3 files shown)


ewoutkramer:
I don't see why this would need to be different types, in fact, the FHIR spec explicitly re-uses the definition of the exclude from the definition of the include element. This is why we name the type after the include element. 

* This means the C# type is the same for both elements
* This is why the TypeName returned should be the same (not the same as the C# type, since that's been "made up", not part of the standard)
* TypeName is supposed to reflect the name of the type, even though it is an instance property. So, I disagree with changing this on getters/setters or make it instance specific in any way.

PR #3508, opened 2026-05-27
Implement Central Package Management (CPM)

granadacoder:
## Description

Implement CPM.  Central Package Management.

https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management

## Related issues

None that I know of.

## Testing
Describe how this change was tested.

dotnet restore
dotnet test

This should be a build improvement, not a functionality improvement.

## FirelyTeam Checklist
- [ ] **Update the title** of the PR to be succinct and less than 50 characters
- [ ] Mark the PR with the label **breaking change** when this PR introduces breaking changes

PR #3509, opened 2026-05-27
Central Package Dependency Management AND LTS/Condition Suggestions

granadacoder:
## Description

Adding Central Package Dependency Management (CPM) ( exactly what https://github.com/FirelyTeam/firely-net-sdk/pull/3508 is )

https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management

AND
(additional)
suggestions for LTS Target Framework Conditions to keep in sync with Microsoft.Extension.*  libraries.

## Related issues
No known requests for CPM

## Testing

dotnet restore

dotnet test

This is a .sln and .csproj improvement.  Not a functional library change.

## FirelyTeam Checklist
- [ x] **Update the title** of the PR to be succinct and less than 50 characters
- [x ] Mark the PR with the label **breaking change** when this PR introduces breaking changes


Directory.Packages.props
```diff
@@ -0,0 +1,61 @@
...
+		<PackageVersion Include="System.ComponentModel.Annotations" Version="5.0.0" />
+		<PackageVersion Include="System.Reflection.Emit.Lightweight" Version="4.7.0" />
+	</ItemGroup>
+
+	<!-- Keep netstandard(s) in sync with the lowest version of the LTS version. https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core -->
+	<ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.1' OR '$(TargetFramework)' == 'net8.0' ">
+		<!-- Keep Microsoft.Extensions.* packages in sync with the lowest LTS version -->
+		
+		<!-- https://www.nuget.org/packages/Microsoft.Extensions.Caching.Memory/#versions-body-tab -->
+		<PackageVersion Include="Microsoft.Extensions.Caching.Memory" Version="8.0.1" />
```
granadacoder, line 35: This is the actual issue that led me to start this PR.

My suggestion:
When targeting net8.0 I think the Microsoft.Extensions.* libaries should match.


The way it is coded (before my suggestions) is that ... 

when targeting net8.0......the version for the below is at 10.0.x

"Microsoft.Extensions.Caching.Memory" Version="10.0.7

and I think net8.0 .. being matched to 10.0.x is a small miscue.

============
I think for:

netstandard2.1 and net8.0
it should be:
"Microsoft.Extensions.Caching.Memory" Version="8.0.x"

(and netstandard2.1 should "map to" the lowest current LTS version)

============

and then for net10.0

it should be:
"Microsoft.Extensions.Caching.Memory" Version="10.0.7

andrzejskowronski: Not sure we agree. 
Since those packages multi-target .NET8 and .NET10 we'd think it's better to use the latest compatible version, with any bug fixes or performance improvements that come along with it.
I'd imagine Microsoft would not multi-target them if the intent was to prevent it?
I understand it is somewhat annoying as a consumer that you're being forced to upgrade the libraries with every new SDK version, so we'll investigate around that pain point :)

granadacoder: Well.  The way around the pain point..is what I proposed. :)

It is not saying you cannot go higher.

My proposal is saying "this is the minimal".
Aka least common denominator.

IMHO, framework code needs to be least common denominator aware.
...

When there is a CVE issue....

The older versions are updated.


Look at System.Text.Json and 8.x versions for example.

granadacoder:
Hi @andrzejskowronski 

Would you add:


https://github.com/GinoCanessa
and
https://github.com/ewoutkramer

to the PR.

We will try to discuss at FHIR d3v days occurring presently.

src/Benchmarks/Benchmarks.csproj
```diff
@@ -24,3 +24,4 @@
 	</ItemGroup>
 
 </Project>
+
```
granadacoder, line 25: I"ll try to clean up these superfluous (unneeded) changes.

granadacoder:
Ok.  I cleaned up the white-space issues.  So it is only the important changes in this PR now.

granadacoder:
I believe I have convinced @ewoutkramer .  Or at least got a "I don't 'not-agree'" from the mini group discussion.      :)

He (e.k.) is going to sleep on it.

Directory.Packages.props
```diff
@@ -20,17 +20,36 @@
...
+	</ItemGroup>
+
+	<!-- Add a new Condition block for all other LTS versions. https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core  -->
+	<ItemGroup Condition=" '$(TargetFramework)' == 'net10.0' ">
+		<!-- Keep Microsoft.Extensions.* packages in sync with this Condition's LTS version -->
+
+		<!-- https://www.nuget.org/packages/Microsoft.Extensions.Caching.Memory/#versions-body-tab -->
+		<PackageVersion Include="Microsoft.Extensions.Caching.Memory" Version="10.0.9" />
```
granadacoder, line 43: bumped to 10.0.9 

as per 

(dependabot)
https://github.com/FirelyTeam/firely-net-sdk/pull/3524

granadacoder:
Ahh.

I see why the PackageReference was introduced:


```
public class CachingTerminologyService : ITerminologyService
{
    public CachingTerminologyService(ITerminologyService terminologyService, IMemoryCache cache, MemoryCacheEntryOptions? entryOptions = null)
    {
        _terminologyService = terminologyService;
        _entryOptions = entryOptions ?? DEFAULT_ENTRY_OPTIONS;
        _cache = cache;
    }

```


CachingTerminologyService is a (Cache-Aside)  Decorator for ITerminologyService.


..

As a "package-reference" PURIST  :)           I would have put the Cache-Aside-Decorator (CachingTerminologyService.cs)  in a different csproj and that csproj would reference Microsoft.Extensions.Caching.Memory

and the raw library would have been Package-Reference "cleaner"  (less dependencies).

But that's me  :)

GinoCanessa:
To my knowledge, the biggest issue with the dependencies of this type is that the .Net team does not do cross-major testing (e.g., `Microsoft.Extensions.Options` version `10.x` is tested against .Net 10 (and not .Net 9, etc.)).

It generally works but may have issues.

Directory.Packages.props
ewoutkramer: Sloan (@granadacoder ) and myself sat down with the Microsoft guys and I think we agreed we'll use Sloan's approach so we're not forcing downstream users to newer versions all the time (while we can still upgrade it ourselves in our downstream stuff, mostly Firely Server).

As a separate discussion: .NET 8 end-of-support is on November 10 this year, so I think we will be switching to .NET 10 soonish. Of course, we could still keep producing 8+10, or let 8 drop back to .netstandard..

ewoutkramer (approved):

ewoutkramer:
> 
> and the raw library would have been Package-Reference "cleaner" (less dependencies).
> 
> But that's me :)

I think we have too many projects already ;-)   Although this CPM helps managing all of them for sure!

PR #3517, opened 2026-06-10
fix: restore collection initializer support on Parameters class

### LINKED ISSUE #3447: Parameters class in SDK 6 cannot be used with collection initializers like in older SDK
Our testing code and application code use collection initializers with Parameters resource which was possible in FHIR SDK 5.x. In SDK 6, this is no longer possible since `Base` no longer implements `IDictionary<string,object>`, and in turn `IEnumerable` which is required for collection initializers.

We feel this is a step back for usability and readability, and since we have other teams consuming our libraries, they may also face this issue once they update to use this version of SDK.

```csharp
// SDK 5
var parameters = new Parameters
{
   { "value1", new FhirString("test") },
   { "value2", new Integer(1) }, 
   { "value3", [ Tuple.Create<string, Base>("value3.1", new FhirString("a")) ] }
};

// SDK 6
var parameters = new Parameters();
parameters.Add("value1", new FhirString("test"));
parameters.Add("value2", new Integer(1));
parameters.Add("value3", [ Tuple.Create<string, Base>("value3.1", new FhirString("a")) ]);
```

In our testing code, we have a workaround for the issue where we add a new class `Parameters2` which approximates the previous functionality with minimal code changes. Production code however would not be able to use this class and would have to go back to the old way of initializing the parameter contents.

```csharp
internal class Parameters2 : IEnumerable
{
    public Parameters Parameters { get; } = new Parameters();

    public void Add(string name, Base value)
    {
        Parameters.Add(name, value);
    }

    public void Add(string name, IEnumerable<Tuple<string, Base>> value)
    {
        Parameters.Add(name, value);
    }

    IEnumerator IEnumerable.GetEnumerator()
        => Parameters.Parameter.GetEnumerator();

    public static implicit operator Parameters(Parameters2 parameters2)
        => parameters2.Parameters;
}
```

My proposed enhancement is simple and just have Parameters implement IEnumerable and expose the enumerator of the Parameters list.

```csharp
public partial class Parameters : IEnumerable
{
    IEnumerator IEnumerable.GetEnumerator()
        => Parameters.Parameter.GetEnumerator();
}
```

_Originally posted by @almostchristian in https://github.com/FirelyTeam/firely-net-sdk/discussions/3411_

MadhukarJukanti:
## Summary
  - SDK 6 removed `IDictionary` from `Parameters`, breaking C# collection initializer syntax that worked in SDK 5
  - The `Add(string, Base?)` method was already present — the only missing piece was `IEnumerable<ParameterComponent>`
  - Added `IEnumerable<ParameterComponent>` to the `Parameters` partial class delegating to the underlying `Parameter` list

  ## Before (broken in SDK 6)
  ```csharp
  var p = new Parameters { { "value1", new FhirString("test") } }; // CS1061

  After (restored)

  var p = new Parameters { { "value1", new FhirString("test") } }; // works

  Changes

  - src/Hl7.Fhir.Base/Model/Parameters.cs — implement IEnumerable<ParameterComponent> + explicit IEnumerable
  - src/Hl7.Fhir.Support.Poco.Tests/Model/ModelTests.cs — two new tests covering collection initializer and enumeration

  Fixes #3447

### DIFF
--- a/src/Hl7.Fhir.Base/Model/Parameters.cs
+++ b/src/Hl7.Fhir.Base/Model/Parameters.cs
@@ -32,6 +32,7 @@
 
 using Hl7.Fhir.Utility;
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
@@ -42,8 +43,13 @@ namespace Hl7.Fhir.Model;
 /// This is the Parameters partial class that adds all the specific functionality of a Parameters to the model
 /// </summary>
 [DebuggerDisplay(@"\{Count={_Parameter != null ? _Parameter.Count : 0}}")]
-public partial class Parameters
+public partial class Parameters : IEnumerable<Parameters.ParameterComponent>
 {
+    /// <inheritdoc/>
+    public IEnumerator<ParameterComponent> GetEnumerator() => Parameter.GetEnumerator();
+
+    IEnumerator IEnumerable.GetEnumerator() => Parameter.GetEnumerator();
+
     public Parameters()
     {
         // Nothing
--- a/src/Hl7.Fhir.Support.Poco.Tests/Model/ModelTests.cs
+++ b/src/Hl7.Fhir.Support.Poco.Tests/Model/ModelTests.cs
@@ -438,6 +438,31 @@ public class ModelTests
         ];
         Assert.IsTrue(expected.SequenceEqual(children));
     }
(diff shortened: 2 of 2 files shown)


andrzejskowronski (approved):

andrzejskowronski:
@MadhukarJukanti thanks for your contribution! :)

Issue #2966, opened 2024-11-14
Avoid calling FindOrImport on ModelInspector


ewoutkramer:
When we will have manually-created classmappings, these will not necessarily be represented using their own POCO anymore (e.g. custom resources, which are represented by DynamicResource) - this means that we will need to be consistent and lookup ClassMappings by their name (which can be a custom type), not their class (which will often by one of the Dynamic types). 

This also means we should avoid FindOrImport, since not everything is based on a class. This then means we cannot "import" types as we find them, but rather we should have them all (lazily) present in the ModelInspector. This is already true for all the main POCOs (since we scan the assemblies), but it might not be true for Backbone types (I am not sure).

Also, this means we will no longer be able to discover satellite types via shared types, if we only load the shared types in the modelinspector initially. I think this is a good thing, since that only causes problems: we don't know the current FHIR version if we are reflecting on a shared type.  But it does place the requirement on the user that all assemblies be loaded into the ModelInspector.  Which I think is only fair.

ewoutkramer:
There is more to this than meets the eye. We still need FindOrImport to create mappings for all the Code<T> as they are encountered. And it is probably still used for the backbone types. That last one is easy to fix, but I am not too sure about the Code<T>.  Also, we're talking about to usecases here:

* We're reflecting on the ACTUAL types in POCO's. We should be able to call FindOrImport, instead of by name, as these types do never have a custom type name, so this won't conflict in the mappings
* We're trying to link from a property to a custom type. We have no facilities for doing this in PropertyMapping currently, but this should go via the lookup by name, not type (as the actual type, DynamicXXX, is ambiguous).

PR #3519, opened 2026-06-11
Support custom resources and datatypes in ClassMapping/ModelInspector

### LINKED ISSUE #2850: Change ClassMappings so it can support custom resources
As part of our effort to support "custom resources" through Pocos (and more specifically through the new DynamicResource/DynamicDataType), we will need a way to provide metadata about these resources so they can be parsed and serialized. Custom resources are useful in general for supporting data that does not have an equivalent in FHIR (yet), but also to add new elements for newer, in-preview versions of FHIR (or even IGs), without having to generate a full new version of the SDK for it.

Currently, the ClassMappings are constructed by reflecting on the POCO's, but by definition DynamicResources will not have any static reflectable type information, so the information should be supplied by the user.  In fact, we have an existing mechanism for that,  `IStructureDefinitionSummary` which was designed to support the `ITypedElement` subsystem. The current ClassMappings implement `IStructureDefinitionSummary`, so there is already a strong link between the two abstractions. Whatever we do, the new capabilities shoul;d still support `IStructureDefinitionSummary`, so the ITypedElement stack can be retained.

I have done a quick survey over the metadata that the serializers & parsers need, and it looks to me that none of it is inherently tied to reflection, even the generated getters/setters could be rewritten to use IDictionary (if we so choose), and thus be re-useable across resources and dynamic resources. 

Note that the POCO's should be prepared for this change to use something akin `IDictionary` (not only `IReadonlyDictionary`) so the (de)serializers can treat poco's and dynamic resources uniformly and so they can deal with overflow elements. This is a separate task (TODO: add a link to that separate task once created).

ewoutkramer:
## Summary

Enables registering `ClassMapping`s for custom resources and datatypes (backed by `DynamicResource`/`DynamicDataType`) with the `ModelInspector`, expressed purely in the domain language of the mapping classes (`Type`, `ClassMapping`, `PropertyMapping`). Supersedes #3489, which leaked `ITypeSerializationInfo` into the public mapping API.

The core modeling change: with custom types, a .NET `Type` no longer identifies a FHIR type (all custom types share the same dynamic .NET type), so:

- `PropertyMapping` gains **`FhirTypeMappings`**, the authoritative (lazily resolved) list of the element''s types as `ClassMapping`s. The legacy `FhirType` (`Type[]`) remains as a compatibility facade, but degrades to the native dynamic types for custom choices (documented on the property). `buildTypes`/`GetInstantiableType`/`PromoteToList` now work off `FhirTypeMappings`.
- A new constructor builds a custom `PropertyMapping` from the property `Type` and the `ClassMapping`s of its (choice) types - no serialization-infrastructure types in the signature, and the `IElementDefinitionSummary.Type` cache stays a cache.
- Custom mappings register by **name/canonical only**, never in the by-type index. `ModelInspector` gains instance-aware `FindClassMapping(Base)`/`FindOrImportClassMapping(Base)` honoring `IDynamicType.DynamicTypeName`.
- `ClassMapping.IsAbstract` can be set explicitly for mappings whose native type does not reflect the FHIR type''s abstractness.

## Consumer fixes (so custom type identity survives end-to-end)

- The JSON/XML (de)serializers, `PocoElementNode`, `PocoNode` and the poco validator resolve mappings instance-aware, or directly from the declared `PropertyTypeMapping` instead of roundtripping through the .NET type (which silently degraded custom types to their generic dynamic mapping).
- `IElementDefinitionSummary.Type` references are always by type *name* (never canonical), since consumers match them against choice-element type suffixes.

## Validation

- `AllowedTypesAttribute` gains name-based matching (`TypeNames`), since `Type[]` cannot express custom choice types; `Types` and `TypeNames` act as a union when combined. Choice `PropertyMapping`s built from custom types synthesize this validation automatically, so the metadata and the validation rule cannot drift apart.
- The attribute validator now validates custom properties normally and accepts registered custom resources and open-choice custom datatypes, while still reporting unknown elements (including unrecognized elements *on* a custom class) and unregistered dynamic resources.

## Testing

- New tests build a custom datatype plus a custom resource with a backbone and a choice element (one choice being the custom datatype) through the public API only, then assert the `IStructureDefinitionSummary` surface, full JSON **and** XML parse/serialize roundtrips (including dynamic type identity of parsed values), typed-element navigation, and a validation matrix (allowed/disallowed/unregistered choice types, open choices, cardinality, unknown elements, unknown resource types, mixed `Types`+`TypeNames`).
- Regression sweeps green: Hl7.Fhir.Support.Tests (500), Hl7.Fhir.Support.Poco.Tests (552), Hl7.Fhir.ElementModel.STU3.Tests (70), R4 validation tests (41), full Hl7.Fhir.Serialization.STU3.Tests (13,404).

## Out of scope

- The StructureDefinition-summary -> ClassMapping translation lives downstream (Firely Server) on top of this API. Note for that implementation: set `Order` on every custom `PropertyMapping`, since XML serialization orders elements by it.
- Type hierarchies for custom types (`baseDefinition`) and `urn:` canonicals.

Closes #2850

### DIFF
--- a/src/Hl7.Fhir.Base/ElementModel/PocoElementNode.cs
+++ b/src/Hl7.Fhir.Base/ElementModel/PocoElementNode.cs
@@ -31,7 +31,7 @@ namespace Hl7.Fhir.ElementModel
         {
             Current = root;
             _inspector = inspector;
-            _myClassMapping = _inspector.FindOrImportClassMapping(root.GetType())!;
+            _myClassMapping = _inspector.FindOrImportClassMapping(root)!;
 
             InstanceType = ((IStructureDefinitionSummary)_myClassMapping).TypeName;
             Definition = ElementDefinitionSummary.ForRoot(_myClassMapping, rootName ?? root.TypeName);
@@ -46,10 +46,7 @@ namespace Hl7.Fhir.ElementModel
             Current = instance;
             _inspector = inspector;
 
-            var instanceType = definition.Choice != ChoiceType.None
-                ? instance.GetType()
-                : determineInstanceType(definition);
-            _myClassMapping = _inspector.FindOrImportClassMapping(instanceType)!;
+            _myClassMapping = determineInstanceMapping(instance, definition);
             InstanceType = ((IStructureDefinitionSummary)_myClassMapping).TypeName;
             Definition = definition;
 
@@ -58,9 +55,15 @@ namespace Hl7.Fhir.ElementModel
             ShortPath = shortPath;
         }
 
-        private Type determineInstanceType(PropertyMapping definition)
+        private ClassMapping determineInstanceMapping(Base instance, PropertyMapping definition)
         {
-            if (!definition.IsPrimitive) return definition.PropertyTypeMapping.NativeType;
+            // For a choice element, the instance determines the type. For other elements we use the
+            // declared type, which we already have as a mapping - note that going through the .NET
+            // type here instead would lose the identity of custom types, since these all share
+            // the same (dynamic) .NET type.
+            if (definition.Choice != ChoiceType.None)
+                return _inspector.FindOrImportClassMapping(instance)!;
+            if (!definition.IsPrimitive) return definition.PropertyTypeMapping;
             throw new NotSupportedException(
                 $"Encountered unexpected primitive type {Name} for PocoElementNode.InstanceType.");
         }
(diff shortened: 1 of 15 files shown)


ewoutkramer:
Closes #2892. Closes #2966.

andrzejskowronski (approved):

PR #3525, opened 2026-06-17
Optimize serialization locking

### LINKED ISSUE #3453: Excessive locking while deserializing
NCQA reported that if you parrallelize deserialization, there is lock contention on the global lock on `SetAnnotation`. This raises some questions:

* Why is the deserializer calling SetAnnotation?  There is certainly an option to annotate the deserialized instance with encountered errors and position information, but I thought this was off by default. Let's check.
* A global lock is not necessary, as we only need to make sure each instance's annotation list is protected, so this could be more localized. Or maybe replaced by a ConcurrentDictionary/CuncurrentBag/ConcurrentList.

andrzejskowronski:
## Description
Specializes AnnotationList with non-locking update implementation on top of the existing ConcurrentDictionary, provides default locking implementation on the instance itself.

## Related issues
Fixes #3453

### DIFF
--- a/src/Hl7.Fhir.Base/Utility/AnnotationList.cs
+++ b/src/Hl7.Fhir.Base/Utility/AnnotationList.cs
@@ -38,7 +38,7 @@ public class AnnotationList : IAnnotatable, IEnumerable<object>
     }
 
     /// <summary>
-    /// Add an annotation to the list, possibly replacing an existing one.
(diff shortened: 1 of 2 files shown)


cknaap (approved):
Looks good. Maybe @ewoutkramer still wants to comment on the new IAnnotatable member. I like it.

ewoutkramer (approved):

PR #3531, opened 2026-06-24
Fix serialization base64 size limits

### LINKED ISSUE #3501: System.Text.Json serialization limit hit when serializing large base64 document content
We're encountering System.Text.Json's 125MB serialization limit when working with FHIR resources that contain large base64-encoded document content.
                                                                                                                                                                                                                                         
### **Problem**                                                                                                                                                                                                                                
                                                                                                                                                                                                                                         
The Firely SDK uses System.Text.Json for serialization and converts base64 byte arrays to strings. System.Text.Json has a hard limit on string value length during serialization, which causes failures when document content is large.
                  
###   **Background**                                                                                                                                                                                                                             
                  
System.Text.Json removed the base64 length limit for byte arrays in this PR: https://github.com/dotnet/runtime/pull/85334                                                                                                              
                  
However, since the Firely SDK converts base64 bytes to strings before serialization, this fix doesn't apply. The string length limit still triggers.                                                                                   
                  
We've added context about this issue affecting production workloads in the original System.Text.Json issue: https://github.com/dotnet/runtime/issues/39953                                                                             
                  
###   **Request**                                                                                                                                                                                                                                
                  
Could the SDK be updated to leverage the byte array serialization path (which no longer has the limit) rather than converting to strings? Or alternatively, is there another approach that would avoid hitting this limit?             
                  
###   **Environment**                                                                                                                                                                                                                            
  - Firely SDK version: **6.1.1**
  - .NET version: **8**

andrzejskowronski:
## Description

This pull request addresses an issue with serializing large Base64 strings in FHIR primitives by improving how binary data is handled during JSON serialization. The main focus is to avoid limitations in `System.Text.Json` when serializing large Base64 strings.

**Improvements to FHIR primitive serialization:**

* Updated `serializeFhirPrimitive` in `BaseFhirJsonSerializer.cs` to pass the underlying byte array of `Base64Binary` values directly to the serializer, bypassing the 125MB limitation of `System.Text.Json` for strings.
* Enhanced `SerializePrimitiveValue` to handle `byte[]` values by writing them as Base64 strings using `WriteBase64StringValue`, ensuring proper serialization of binary data.

## Related issues
Fixes #3501

### DIFF
--- a/src/Hl7.Fhir.Base/Serialization/BaseFhirJsonSerializer.cs
+++ b/src/Hl7.Fhir.Base/Serialization/BaseFhirJsonSerializer.cs
@@ -14,8 +14,10 @@ using Hl7.Fhir.Model;
 using Hl7.Fhir.Utility;
 using System;
 using System.Buffers;
+using System.Buffers.Text;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using System.Text.Json;
 
 namespace Hl7.Fhir.Serialization;
@@ -187,7 +189,7 @@ public class BaseFhirJsonSerializer(ModelInspector inspector)
                     writeStartArray(elementName, numNullsMissed, writer);
                 }
 
-                SerializePrimitiveValue(value.JsonValue, writer);
+                SerializePrimitiveValue(value, writer);
             }
             else
             {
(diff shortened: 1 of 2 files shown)


src/Hl7.Fhir.Base/Serialization/BaseFhirJsonSerializer.cs
```diff
@@ -253,13 +254,23 @@ private void serializeFhirPrimitive(string elementName, PrimitiveType value, Utf
         {
             // Write a property with 'elementName'
             writer.WritePropertyName(elementName);
-            SerializePrimitiveValue(value.JsonValue, writer);
+            SerializePrimitiveInternal(value, writer);
```
ewoutkramer: This evades calling SerializePrimitive(). This is an overloadable function so subclasses can customize serialization. By circumventing this for Base64Binary we (unexpectedly) take away a customization path.

What I *should* have done is have SerializePrimitive take a `PrimitiveType` as a paramter, instead of object.  Maybe we can still do that -> a new overload that (for the non-special case) still calls the older overload.

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonSerializer.cs
andrzejskowronski: Yes, that was the issue I was having, don't mind making this one public and virtual to make it overridable too

ewoutkramer: rite!

ewoutkramer (approved):

PR #3537, opened 2026-06-29
Ensure PropertyMapping gets Code<T> for the PropertyTypeMapping

andrzejskowronski:
## Description
This pull request updates how `Code<T>` types are handled in property mapping and serialization/deserialization, ensuring that the exact generic type is used rather than the base `Code` type.

**Property Mapping Improvements:**

* Updated `determineMappingType` in `PropertyMapping.cs` to return the exact generic type for `Code<T>` instead of the base `Code` type, ensuring proper class mapping and instance creation.

**Test Updates and Additions:**

* Added a new unit test `TestDeserializeCodeOfTIntoPoco` in `FhirJsonDeserializationTests.cs` to verify that deserialization of a resource with a `Code<T>` property results in the correct generic type and value.
* Updated `GetClassMappingForProperty` test in `PropertyMappingTest.cs` to expect the `PropertyTypeMapping.Name` for `ContactPoint.System` to be the specific generic type (`codeOfT<Hl7.Fhir.Model.ContactPoint+ContactPointSystem>`) instead of just `code`

### DIFF
--- a/src/Hl7.Fhir.Base/Introspection/PropertyMapping.cs
+++ b/src/Hl7.Fhir.Base/Introspection/PropertyMapping.cs
@@ -405,9 +405,11 @@ public class PropertyMapping : IElementDefinitionSummary
         if (typeof(Enum).IsAssignableFrom(implementingType))
             return typeof(Enum);
 
+        // To ensure CreateInstance indeed creates pocos that fit the expected property values, we need to return the
+        // implementing type, so PropertyMapping.PropertyTypeMapping will link to the exact ClassMapping needed
         // For all Code<T>, we use the mapping for Coding
-        if (implementingType.IsConstructedGenericType && implementingType.GetGenericTypeDefinition() == typeof(Code<>))
-            return typeof(Code);
+        // if (implementingType.IsConstructedGenericType && implementingType.GetGenericTypeDefinition() == typeof(Code<>))
+        //     return typeof(Code);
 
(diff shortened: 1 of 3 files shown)


src/Hl7.Fhir.Base/Introspection/PropertyMapping.cs
```diff
@@ -405,9 +405,11 @@ private static Type determineMappingType(Type[]? overridingTypes, Type implement
         if (typeof(Enum).IsAssignableFrom(implementingType))
             return typeof(Enum);
 
+        // To ensure CreateInstance indeed creates pocos that fit the expected property values, we need to return the
```
ewoutkramer, line 408: This changes the external behaviour of PropertyTypeMapping. This property has returned `Code` and not `Code<T>` for years - and 3rd party tools (like Brian P.'s tooling) will almost certainly depend on the behaviour.

Indeed, we ourselves, depend on this in the PocoNode:

```
           _myClassMapping = determineInstanceMapping(instance, definition);
            InstanceType = ((IStructureDefinitionSummary)_myClassMapping).TypeName;
```

So, even PocoNode.InstanceType will change name (maybe this was the effect you wanted, but still).

You can see that public effect clearly in the changed unit test below.

All in all, I think this is a breaking change that we cannot make.

ewoutkramer (changes requested):

andrzejskowronski:
We'll shelve it for now to not break the expectations of current behavior

PR #3540, opened 2026-07-01
Fix #3502: Detect wrong-cased element and resource type names at the model level

### LINKED ISSUE #3502: Standardize case-sensitivity handling in JSON & XML parsers
[Specification](https://hl7.org/fhir/R4/formats.html) defines serialized elements as case sensitive.
Our current de/serialization uses `ClassMapping.PropertyMappings` with `OrdinalIgnoreCase`, which will lead to us normalizing sensitivity for known properties.

We should emit the error, preferably only during `Strict` parsing/serializing.

**Additional context**
This is most likely quite big breaking change in the serialization, as it was not considered as error before.

ewoutkramer:
## Summary

FHIR names are case-sensitive per spec, but the SDK's lookup dictionaries use `OrdinalIgnoreCase`, silently normalizing wrong-case input. This PR detects and reports these violations — and after review feedback, the detection now lives **in the model/validation layer rather than in the parsers**, since a wrong-cased name is a model violation, not a syntax error.

The rule is now a single sentence: **lenient modes bind every wrong-cased name to the element it nearly matches and stay silent; strict mode binds none, preserves the wire form in the overflow, and fails with exactly one wrong-case error.**

### Design

- **One detection primitive**: `ClassMapping.TryFindElement(name)` (internal) performs the case-insensitive lookup and reports the mapping, the canonical (correctly-cased) name — including a normalized choice-type suffix — and whether the match is case-exact. Both parsers and the validator consume it; the previously duplicated per-parser detection blocks are gone.
- **Two model-level error codes** replace the three format-specific ones: `PVAL131` (`WRONG_CASED_ELEMENT`) and `PVAL132` (`WRONG_CASED_RESOURCE_TYPE`). They apply to JSON and XML alike (a wrong-cased XML root element is a resource type violation, so it raises PVAL132).
- **Exactly one error per failed name**: `PVAL131` when the name differs from a defined element only by casing, `PVAL128` (`UNKNOWN_ELEMENT`) only when it matches nothing at all. Never both.
- **No new settings — binding policy is derived from the `ExceptionFilter`** (see the documented `DeserializerSettings.UsesStrictCaseBinding`):
  - If the active filter *ignores* wrong-case errors (the default, `Recoverable`, `BackwardsCompatible`, `NoOverflow`, `Ostrich`), wrong-cased names are bound to the element they nearly match — data stays in the typed properties.
  - If the filter *reports* them (`Strict`, or any filter via `Enforcing([PVAL131])`), a wrong-cased element is treated as unknown: the data is preserved under its original name in the overflow (and roundtrips), and validation reports `PVAL131`. A failing parse never silently corrects the input.
  - This also guarantees no filter combination can silently strand data in the overflow: data only reaches the overflow under strict binding, which by construction only happens when the filter reports `PVAL131` — so such a parse always fails. This is why `PVAL131` is not in `ISSUES_CAUSED_BY_OVERFLOW` (reasoning recorded in a comment there).
  - `SyntaxOnly`/`Ostrich` set `Validator = null`: case checks are model validation, so nothing is reported and binding stays lenient.
- **The dictionary interface is covered too**: `patient.SetValue("Active", ...)` puts the value in the overflow; `Validate()` now reports `PVAL131` with the correct-casing suggestion — for plain names and choice-suffixed keys alike.

### Behaviour changes vs 6.2 (deliberate)

- **Wrong-cased choice prefixes now bind in lenient modes.** ≤6.2 matched plain names and choice *suffixes* case-insensitively but choice *prefixes* ordinally, so `"DeceasedBoolean"` was an unknown element (default mode: parse failed; `Recoverable`: data hidden in overflow). We consider that asymmetry a bug: it now binds to `Deceased` like any other near miss. This is the only lenient-mode change, and it is in the lenient direction (previously-rejected input now parses).
- **Strict mode reports new errors** — covered by the `<remarks>` on `DeserializationMode.Strict` stating the issue set may expand.
- **A wrong-cased `resourceType` still binds to the matching POCO type, even in Strict** (reported as `PVAL132`). Falling back to `DynamicResource` would degrade the entire resource to overflow data, which seems disproportionate. Candidate to revisit in SDK7.

### Backwards compatibility

- All other lenient-mode behaviour is unchanged from 6.2; correctly-cased input is unaffected everywhere.
- No public API removed relative to `develop`; `FindMappedElementByChoiceName(string, bool)` remains.

## Test plan
- [x] `dotnet build Hl7.Fhir.sln` — clean
- [x] `Hl7.Fhir.Support.Poco.Tests` — 595/595 passed (incl. new tests: strict overflow preservation with single-error assertion, lenient choice-prefix binding, resourceType strict binding, `Enforcing()` opt-in strict binding in Recoverable, dictionary-key detection for plain and choice names)
- [x] `Hl7.Fhir.Serialization.R4.Tests` — 16234/16234 passed (roundtrip corpus)

### DIFF
--- a/src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
+++ b/src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
@@ -280,6 +280,52 @@ public class ClassMapping(
         }
     }
 
+    /// <summary>
+    /// Looks up the element that a (possibly choice-suffixed) serialized name refers to, matching
+    /// names case-insensitively, and reports how well the given name matches the definition.
+    /// </summary>
+    /// <param name="name">The name for the element as encountered in the serialized form.</param>
+    /// <returns>The result of the lookup, or <c>null</c> when the name does not match any element
+    /// of this class, not even when compared case-insensitively.</returns>
+    /// <remarks>This is the single source of truth for detecting case-sensitivity violations in
+    /// element names: FHIR names are case-sensitive, but this lookup will also find "near misses"
+    /// that differ from a defined element name only by casing. Callers decide, based on
+    /// <see cref="ElementLookupResult.IsExactCase"/> and <see cref="ElementLookupResult.MatchedByLegacyRules"/>,
+    /// whether to bind the data to the found element and/or to report an error.</remarks>
+    internal ElementLookupResult? TryFindElement(string name)
+    {
+        if (name == null) throw Error.ArgumentNull(nameof(name));
+
+        // Unsuffixed names: found via the (case-insensitive) name dictionary.
+        if (FindMappedElementByName(name) is { } byName)
+            return new ElementLookupResult(
+                byName,
+                byName.Name,
+                IsExactCase: string.Equals(name, byName.Name, StringComparison.Ordinal),
+                MatchedByLegacyRules: true);
+
+        // Choice elements: the name is the element name plus a type suffix.
+        if (FindMappedElementByChoiceName(name, ignoreCase: true) is { } byChoice)
+        {
+            var canonicalName = byChoice.Name + normalizeChoiceSuffix(name[byChoice.Name.Length..]);
+            return new ElementLookupResult(
+                byChoice,
+                canonicalName,
+                IsExactCase: string.Equals(name, canonicalName, StringComparison.Ordinal),
+                MatchedByLegacyRules: name.StartsWith(byChoice.Name, StringComparison.Ordinal));
+        }
+
+        return null;
+
+        // Resolves the suffix to its actual datatype mapping (case-insensitively) so the canonical
+        // name we suggest is correct even when the suffix casing is wrong. Suffixes that do not
+        // resolve to a known type carry no case information, so they are kept as-is.
+        string normalizeChoiceSuffix(string suffix) =>
+            suffix.Length > 0 && Inspector.FindClassMapping(suffix) is { } typeMapping
+                ? char.ToUpperInvariant(typeMapping.Name[0]) + typeMapping.Name[1..]
+                : suffix;
+    }
+
     #region IStructureDefinitionSummary members
     /// <inheritdoc />
     string IStructureDefinitionSummary.TypeName =>
@@ -422,3 +468,22 @@ public class ClassMapping(
     internal static ClassMapping DynamicDataType => _dynDataMapping ??= ModelInspector.Base.FindOrImportClassMapping(typeof(DynamicDataType)) ??
                                                     throw new InvalidOperationException($"{nameof(DynamicDataType)} mapping not found in Base.");
 }
+
+/// <summary>
+/// The result of looking up an element by a serialized name using <see cref="ClassMapping.TryFindElement(string)"/>.
+/// </summary>
+/// <param name="Mapping">The element that the name refers to (for choice elements: the unsuffixed choice element).</param>
+/// <param name="CanonicalName">The correctly-cased form of the given name, including a correctly-cased
+/// type suffix for choice elements. This is the name that should have been used in the serialized form.</param>
+/// <param name="IsExactCase">Whether the given name is exactly equal to <paramref name="CanonicalName"/>.
+/// When <c>false</c>, the name is a case-sensitivity violation ("near miss").</param>
+/// <param name="MatchedByLegacyRules">Whether this match would also have been found by the lookup rules
+/// used by the SDK before case-sensitivity violations were detected (SDK 6.2 and earlier): element names
+/// were compared case-insensitively, but the element-name part of a choice name was compared
+/// case-sensitively (only its type suffix was case-insensitive). Used to reproduce that legacy binding
+/// behaviour exactly when case-sensitivity violations are not treated as errors.</param>
+internal sealed record ElementLookupResult(
+    PropertyMapping Mapping,
+    string CanonicalName,
+    bool IsExactCase,
+    bool MatchedByLegacyRules);
--- a/src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
+++ b/src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
@@ -243,7 +243,19 @@ public class ClassMapping(
     /// so for where there is no suffix. In this case, however, <see cref="FindMappedElementByName(string)"/>
     /// is faster.
     /// </remarks>
-    public PropertyMapping? FindMappedElementByChoiceName(string name, bool ignoreCase = false)
+    public PropertyMapping? FindMappedElementByChoiceName(string name) => FindMappedElementByChoiceName(name, ignoreCase: false);
+
+    /// <summary>
+    /// Returns the mapping for an element of this class by a name that
+    /// might be suffixed by a type name (e.g. for choice elements).
+    /// </summary>
+    /// <param name="name">The (possibly suffixed) name to look up.</param>
+    /// <param name="ignoreCase">If <c>true</c>, the choice prefix is matched case-insensitively.</param>
+    /// <remarks>Will also return properties for which the name is exactly the same,
+    /// so for where there is no suffix. In this case, however, <see cref="FindMappedElementByName(string)"/>
+    /// is faster.
+    /// </remarks>
+    public PropertyMapping? FindMappedElementByChoiceName(string name, bool ignoreCase)
     {
         if (name == null) throw Error.ArgumentNull(nameof(name));
 
--- a/src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
+++ b/src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
@@ -282,7 +282,7 @@ public class ClassMapping(
 
     /// <summary>
     /// Looks up the element that a (possibly choice-suffixed) serialized name refers to, matching
-    /// names case-insensitively, and reports how well the given name matches the definition.
+    /// names case-insensitively, and reports whether the given name matches the definition exactly.
     /// </summary>
     /// <param name="name">The name for the element as encountered in the serialized form.</param>
     /// <returns>The result of the lookup, or <c>null</c> when the name does not match any element
@@ -290,8 +290,8 @@ public class ClassMapping(
     /// <remarks>This is the single source of truth for detecting case-sensitivity violations in
     /// element names: FHIR names are case-sensitive, but this lookup will also find "near misses"
     /// that differ from a defined element name only by casing. Callers decide, based on
-    /// <see cref="ElementLookupResult.IsExactCase"/> and <see cref="ElementLookupResult.MatchedByLegacyRules"/>,
-    /// whether to bind the data to the found element and/or to report an error.</remarks>
+    /// <see cref="ElementLookupResult.IsExactCase"/>, whether to bind the data to the found
+    /// element and/or to report an error.</remarks>
     internal ElementLookupResult? TryFindElement(string name)
     {
         if (name == null) throw Error.ArgumentNull(nameof(name));
@@ -301,8 +301,7 @@ public class ClassMapping(
             return new ElementLookupResult(
                 byName,
                 byName.Name,
-                IsExactCase: string.Equals(name, byName.Name, StringComparison.Ordinal),
-                MatchedByLegacyRules: true);
+                IsExactCase: string.Equals(name, byName.Name, StringComparison.Ordinal));
 
         // Choice elements: the name is the element name plus a type suffix.
         if (FindMappedElementByChoiceName(name, ignoreCase: true) is { } byChoice)
@@ -311,8 +310,7 @@ public class ClassMapping(
             return new ElementLookupResult(
                 byChoice,
                 canonicalName,
-                IsExactCase: string.Equals(name, canonicalName, StringComparison.Ordinal),
-                MatchedByLegacyRules: name.StartsWith(byChoice.Name, StringComparison.Ordinal));
+                IsExactCase: string.Equals(name, canonicalName, StringComparison.Ordinal));
         }
 
         return null;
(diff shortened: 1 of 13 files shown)


ewoutkramer:
Fixes #3502

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
```diff
@@ -723,13 +726,36 @@ ref Utf8JsonReader reader
...
+            byNameMapping = parentMapping.FindMappedElementByName(propNameWithoutUnderscore);
+            if (byNameMapping is not null && !string.Equals(byNameMapping.Name, propNameWithoutUnderscore, StringComparison.Ordinal))
+                caseMismatchExpectedName = byNameMapping.Name;
+            else
+            {
+                byChoiceMapping = parentMapping.FindMappedElementByChoiceName(propNameWithoutUnderscore, ignoreCase: true);
+                if (byChoiceMapping is not null &&
+                    !string.Equals(byChoiceMapping.Name, propNameWithoutUnderscore[..byChoiceMapping.Name.Length], StringComparison.Ordinal))
+                    caseMismatchExpectedName = byChoiceMapping.Name + propNameWithoutUnderscore[byChoiceMapping.Name.Length..];
+            }
```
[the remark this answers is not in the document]
ewoutkramer: Fixed in d3fcaf817: when a choice prefix has wrong casing, the suffix is now resolved against the actual datatype mapping (via ClassMapping.FindClassMapping) so the suggested name is fully corrected even if the suffix casing is also wrong (e.g. ValueDatetime now suggests valueDateTime, not valueDatetime). Added a regression test (WrongCase_ChoicePrefixAndSuffix_SuggestsFullyCorrectedName) in both JSON and XML test suites.

src/Hl7.Fhir.Base/Serialization/BaseFhirXmlDeserializer.cs
```diff
@@ -584,9 +587,26 @@ private PropertyValueMapping getMappingForElement(
...
+            if (!string.Equals(byNameMapping.Name, elementName, StringComparison.Ordinal))
+                caseMismatchExpectedName = byNameMapping.Name;
+        }
+        else
+        {
+            byChoiceMapping = parentMapping.FindMappedElementByChoiceName(elementName, ignoreCase: true);
+            if (byChoiceMapping is not null &&
+                !string.Equals(byChoiceMapping.Name, elementName[..byChoiceMapping.Name.Length], StringComparison.Ordinal))
+                caseMismatchExpectedName = byChoiceMapping.Name + elementName[byChoiceMapping.Name.Length..];
+        }
```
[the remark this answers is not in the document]
ewoutkramer: Fixed in d3fcaf817: the suffix is now resolved against the actual datatype mapping so the suggested name is fully corrected (e.g. ValueDatetime now suggests valueDateTime, not valueDatetime). Same fix applied symmetrically in the JSON deserializer. Added a regression test (WrongCase_ChoicePrefixAndSuffix_SuggestsFullyCorrectedName).

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
```diff
@@ -652,11 +669,20 @@ private ClassMapping determineResourceClassMappingFromInstance(ref Utf8JsonReade
...
+        // The inspector finds resource types case-insensitively, so the data can still be parsed
+        // into the correct POCO, but a wrong-cased resource type name violates the spec and is
+        // reported. This is a model-level validation, so it is skipped when the validator is off.
+        if (Settings.Validator is not null && !string.Equals(resourceMapping.Name, resourceType, StringComparison.Ordinal))
         {
-            null or { IsResource: false } => new ClassMapping(_inspector, resourceType, typeof(DynamicResource)),
-            { } resourceMapping => resourceMapping,
-        };
+            var (line, pos) = reader.GetLocation();
+            state.Errors.Add(COVE.WRONG_CASED_RESOURCE_TYPE(state.Path.GetInstancePath(), line, pos, resourceType, resourceMapping.Name));
```
ewoutkramer, line 682: It feels a bit weird we're emitting a COVE ("Model") error from the parser - is this the compromise where we stick to using resources instead of DynamicResource so we don't see the misspelled name anymore?

If we REALLY wanted to have this during model validation:
* Could we use an Annotation to annotate the original name, so we can still detect it at model validation? It's true that a model-only user would never set that annotation ofcourse, but a model-only user could not easily create a miscased resource, unless by creating a DynamicResource with a wrongly capitalized existing resource name - this sounds rather explicit if you do that manually.
* What happens if I create a DynamicResource with the name "patient" now?

src/Hl7.Fhir.Support.Poco.Tests/PocoValidationTests/ValidatableObjectTests.cs
```diff
@@ -44,6 +44,41 @@ public void TestCodeOfT()
...
+    public void WrongCasedChoiceDictionaryKey_IsReportedByValidation()
+    {
+        // Same for a suffixed choice-element key: it is unknown to the POCO, but validation
+        // detects that it only differs from a correct choice name by casing.
+        var patient = new Patient();
+        patient.SetValue("DeceasedBoolean", new FhirBoolean(true));
+
+        var errors = patient.Validate();
+        errors.Should().Contain(e => e.ErrorCode == COVE.WRONG_CASED_ELEMENT_CODE && e.Message.Contains("'deceasedBoolean'"));
+        errors.Should().Contain(e => e.ErrorCode == COVE.UNKNOWN_ELEMENT_CODE);
```
ewoutkramer: Why does this return two errors and the previous unit test returns only one?

src/Hl7.Fhir.Support.Poco.Tests/PocoValidationTests/ValidatableObjectTests.cs
```diff
@@ -44,6 +44,41 @@ public void TestCodeOfT()
...
+    [TestMethod]
+    public void WrongCasedChoiceDictionaryKey_IsReportedByValidation()
+    {
+        // Same for a suffixed choice-element key: it is unknown to the POCO, but validation
+        // detects that it only differs from a correct choice name by casing.
+        var patient = new Patient();
+        patient.SetValue("DeceasedBoolean", new FhirBoolean(true));
+
+        var errors = patient.Validate();
+        errors.Should().Contain(e => e.ErrorCode == COVE.WRONG_CASED_ELEMENT_CODE && e.Message.Contains("'deceasedBoolean'"));
```
ewoutkramer, line 69: Why two errors?

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
ewoutkramer: Yes — that's exactly the compromise. When the parser sees `resourceType: "patient"` it binds to the real `Patient` POCO (the decision noted in the PR description: falling back to `DynamicResource` would demote the *entire* resource to overflow, which is disproportionate for a casing slip). But once it's a `Patient`, its `TypeName` is `"Patient"` — the original `"patient"` casing exists nowhere in the POCO. So the parser is the only place that still holds the wire name; by model-validation time it's gone. That's why the error has to be raised there.

To keep the parser layering-honest I made it a `COVE`/PVAL132 rather than a `FhirJsonException`/`FhirXmlException`, so both parsers share one code and it filters uniformly with the other case errors. Emitting a "model" code from the parser is admittedly the unusual bit, but it's the pragmatic consequence of binding to the real POCO.

On the two alternatives:

**Annotate the original name and recover it at model-validation time** — doable, but it only helps the *parse* path (a model-only user never sets the annotation), so we'd be adding a parse-time side channel *plus* a model-validation rule to relocate a check the parser can already do directly with the info in hand. More machinery for the same outcome.

**What happens if you `new DynamicResource("patient")` today** — you already get an error, just a different one. `FindOrImportClassMapping` resolves `"patient"` case-insensitively to the real `Patient` mapping (`ClassMappingCollection._byName` is `OrdinalIgnoreCase` — the same root cause as this whole issue), and `ValidateObject` flags any `DynamicResource` that resolves to a real, non-custom mapping as `UNKNOWN_RESOURCE_TYPE` (PVAL130): *"Encountered unknown resource type 'patient'"*. So the "model-only user hand-builds a miscased resource" scenario you describe is already covered — under the generic unknown-resource-type code rather than a wrong-case-specific one. Which is itself an argument that we *don't* need the annotation path: the only route that both binds to the real POCO **and** knows the miscasing is the parser; the manual-`DynamicResource` route is already caught.

If you'd rather have symmetry with elements — wrong-cased resource type → `DynamicResource` in Strict, detected purely at model level — it's a one-line change in each parser. The cost is that the whole resource then lands in overflow instead of a typed `Patient`. I went with bind-and-report as the lesser evil, but happy to flip it if you prefer the cleaner layering.

src/Hl7.Fhir.Support.Poco.Tests/PocoValidationTests/ValidatableObjectTests.cs
ewoutkramer: It comes down to whether the wrong-cased name resolves through the case-insensitive name dictionary.

Validation walks the elements and does `classMapping.FindMappedElementByName(name)` (`PocoValidationExtensions.doObjectValidation`, line 70). That dictionary (`_byName`, `OrdinalIgnoreCase`) holds **simple element names**, but for choice elements it only holds the *unsuffixed* stem (`deceased`) — never the serialized `deceasedBoolean`.

- `"Active"` → `FindMappedElementByName("Active")` hits the `active` mapping case-insensitively → `ValidateProperty` gets a non-null mapping → it's a bound near-miss → **1 error** (`WRONG_CASED_ELEMENT`).
- `"DeceasedBoolean"` → `FindMappedElementByName` misses (no suffixed key in `_byName`) → `ValidateProperty` gets `null` → it goes down the unknown-element branch, where `TryFindElement` recognizes the case-insensitive near-miss → **2 errors** (`WRONG_CASED_ELEMENT` + `UNKNOWN_ELEMENT`).

So the invariant is: `WRONG_CASED_ELEMENT` is *always* reported; `UNKNOWN_ELEMENT` accompanies it whenever the value couldn't be bound and genuinely sits in overflow.

One inconsistency this exposes, worth a decision: in the **parser** strict path a wrong-cased *simple* element also yields both errors — it's routed to an ad-hoc/unknown mapping, so its data really is in overflow (see `WrongCaseNames_StrictMode_PreservesDataInOverflow`). But in the standalone `Validate()` path here, that same `"Active"` resolves case-insensitively to the real mapping and reports only one, even though its value is likewise sitting in the overflow. If we want the two paths to agree — always emit the pair when the data is in overflow — the fix is small (have `ValidateProperty` also emit `UNKNOWN_ELEMENT` when the near-miss value isn't actually in the typed property). I left it as-is for now because detection is never lost, only the companion `UNKNOWN_ELEMENT` differs. Let me know which way you'd like it.

src/Hl7.Fhir.Support.Poco.Tests/PocoValidationTests/ValidatableObjectTests.cs
ewoutkramer: Short version (full mechanism in the `UNKNOWN_ELEMENT`-assertion thread on the next line): `"DeceasedBoolean"` is a *choice* name, and the validator's `FindMappedElementByName` only indexes the unsuffixed stem (`deceased`), so it can't bind the value and routes it through the unknown-element branch — which emits `WRONG_CASED_ELEMENT` **and** `UNKNOWN_ELEMENT`. A wrong-cased *simple* name like `Active` resolves case-insensitively to its mapping and reports only the single `WRONG_CASED_ELEMENT`.

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
ewoutkramer: I agree with your reasoning.

src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
```diff
@@ -243,16 +243,29 @@ PropertyMappingCollection createCollection()
     /// so for where there is no suffix. In this case, however, <see cref="FindMappedElementByName(string)"/>
     /// is faster.
     /// </remarks>
-    public PropertyMapping? FindMappedElementByChoiceName(string name)
+    public PropertyMapping? FindMappedElementByChoiceName(string name) => FindMappedElementByChoiceName(name, ignoreCase: false);
```
andrzejskowronski: That's keeping old error scenario we discussed, right?

src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
```diff
@@ -267,6 +280,50 @@ PropertyMappingCollection createCollection()
...
+    /// element names: FHIR names are case-sensitive, but this lookup will also find "near misses"
+    /// that differ from a defined element name only by casing. Callers decide, based on
+    /// <see cref="ElementLookupResult.IsExactCase"/>, whether to bind the data to the found
+    /// element and/or to report an error.</remarks>
+    internal ElementLookupResult? TryFindElement(string name)
+    {
+        if (name == null) throw Error.ArgumentNull(nameof(name));
+
+        // Unsuffixed names: found via the (case-insensitive) name dictionary.
+        if (FindMappedElementByName(name) is { } byName)
```
andrzejskowronski, line 293: A bit wasteful as `ByChoiceName` will also call `ByName` first. Maybe we could unify it?

andrzejskowronski (changes requested):

src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
ewoutkramer: I am wondering whether the ignoreCase overload is still needed.

ewoutkramer: Removed in a40f235fa. The overload's only consumer was `TryFindElement`, so it's gone: the choice-prefix scan is now a private `findChoiceMatch(name, comparison)` helper shared by the public (case-sensitive) `FindMappedElementByChoiceName(string)` — restored to its exact released signature and semantics — and `TryFindElement`, which passes `OrdinalIgnoreCase` directly.

ewoutkramer: Not quite — this single-arg overload is the *released* public API (its case-sensitive prefix matching restored exactly for binary/behavioural compat; it's still used by `PocoNode`/`NewPocoBuilder` and external callers). It is no longer part of the wrong-case detection path at all: the parsers and validator go through the internal `TryFindElement`, which matches case-insensitively and reports exactness. As of a40f235fa the `ignoreCase` overload that sat between the two is gone.

src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
ewoutkramer: Unified in a40f235fa: `TryFindElement` no longer calls `FindMappedElementByChoiceName`. Both it and the public lookup now share a private `findChoiceMatch(name, comparison)` that does only the choice-prefix scan, so the name dictionary is consulted exactly once per lookup.

andrzejskowronski (approved):

PR #3541, opened 2026-07-03
Remove redundant null-checks and initialization logic from `setLink` method

andrzejskowronski:
## Description
Stumbled on the logic when using the `Bundle.SelfLink` etc. helpers. Property claims the value is nullable, but it would throw `ArgumentNullException`. Then later down the code we had logic for actually removing the entry if the value is null.

### DIFF
--- a/src/Hl7.Fhir.Base/Model/Bundle.cs
+++ b/src/Hl7.Fhir.Base/Model/Bundle.cs
@@ -118,10 +118,6 @@ namespace Hl7.Fhir.Model
 
         private void setLink(string rel, Uri? uri)
         {
-            if (uri is null) throw new ArgumentNullException(nameof(uri));
-
-            if (Link == null) Link = new List<LinkComponent>();
-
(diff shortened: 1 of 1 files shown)


src/Hl7.Fhir.Base/Model/Bundle.cs
```diff
@@ -118,10 +118,6 @@ public Uri? Alternate
 
         private void setLink(string rel, Uri? uri)
         {
-            if (uri is null) throw new ArgumentNullException(nameof(uri));
-
-            if (Link == null) Link = new List<LinkComponent>();
```
cknaap, line 123: Should Link be nullified as well when the last entry is removed from it?

cknaap (approved):
Approved for the improvement, but with a question

src/Hl7.Fhir.Base/Model/Bundle.cs
andrzejskowronski: I think that would be more wasteful as we always initialize an empty list on the property access if it is null

PR #3562, opened 2026-07-27
Retain XML comments in the POCO-based parser

### LINKED ISSUE #3561: Support retaining XML comments in the POCO-based parser

The legacy ElementModel-based XML parser retains the comments of the source document: `FhirXmlNode` exposes them as a `SourceComments` annotation (`CommentsBefore` / `ClosingComments` / `DocumentEndComments`), and `FhirXmlBuilder` writes them back out again.

The new POCO-based XML parser drops them entirely:

- `XmlReaderExtensions.ReadToContent` skips `XmlNodeType.Comment` without recording it;
- the `MoveToContent()` calls in `BaseFhirXmlDeserializer` and in `readXhtml` skip comments silently;
- the string-based entry points in `PocoDeserializationExtensions` build their `XmlReader` with `IgnoreComments = true`, so comments never even reach the deserializer;
- `BaseFhirXmlSerializer` never writes comments, even when the POCO carries a `SourceComments` annotation.

The consequence is that migrating from the legacy parser to the POCO parser silently loses every comment in the document. For users who keep authored comments in their FHIR XML (conformance resources, test fixtures, hand-maintained examples), a parse/serialize roundtrip destroys them with no warning.

Support comment retention in the POCO-based parser and serializer, reusing the existing public `SourceComments` annotation so both parsers produce the same shape:

- a new opt-in `DeserializerSettings.RetainComments` (default `false`);
- the deserializer collects comments and annotates them onto the POCO of the element they precede, the element they close, and the root for comments trailing the document;
- `BaseFhirXmlSerializer` writes the annotated comments back out.

- *Always on, like the legacy parser.* Rejected: it adds an annotation to every commented element, so it costs memory on a hot path that is otherwise annotation-free. The legacy parser could afford this because it computed the annotation lazily from the backing `XObject` tree; the POCO parser has no tree to go back to and must store it eagerly.
- *A separate XML-specific settings class.* Rejected: `DeserializerSettings` already carries XML-only settings such as `DisallowXsiAttributesOnRoot`.

**Additional context**

Known limitations worth documenting rather than solving, since the POCO model has nowhere to hang these:

- a comment preceding an element that FHIR writes as an XML *attribute* (e.g. `Extension.url`) cannot be written back, because a comment cannot live inside a start tag;
- comments between a resource container (e.g. `contained`) and the resource element inside it move outside the container on roundtrip;
- comments are lost when serializing with a summary filter, since the subsetted clone made for it does not carry annotations;
- when the caller supplies their own `XmlReader`, it must not be configured with `IgnoreComments = true`, or the setting silently does nothing.

Kasdejong:
## Description

The POCO-based XML parser discarded the comments in the source document, where the legacy ElementModel parser retains them as a `SourceComments` annotation and `FhirXmlBuilder` writes them back out. Moving from the legacy parser to the POCO parser therefore lost every comment silently.

This adds an opt-in `DeserializerSettings.RetainComments` (default `false`, since it adds an annotation per commented element) that reuses the existing public `SourceComments` annotation, so both parsers produce the same shape.

## Related issues

Fixes #3561.

### DIFF
--- a/src/Hl7.Fhir.Base/Serialization/BaseFhirXmlDeserializer.cs
+++ b/src/Hl7.Fhir.Base/Serialization/BaseFhirXmlDeserializer.cs
@@ -83,7 +83,7 @@ public class BaseFhirXmlDeserializer
     /// <remarks>The <see cref="ParserSettings.ExceptionFilter"/> influences which issues are returned.</remarks>
     public bool TryDeserializeResource(XmlReader reader, [NotNullWhen(true)] out Resource? instance, out IEnumerable<CodedException> issues)
     {
-        PocoDeserializerState state = new() { Comments = Settings.RetainComments ? new() : null };
+        var state = createState();
 
         // If the stream has just been opened, move to the first token. (skip processing instructions, comments, whitespaces etc.)
         reader.MoveToContent(state);
@@ -95,7 +95,7 @@ public class BaseFhirXmlDeserializer
 
         // Whatever is left after the root element has been read are the comments trailing the document.
         if (instance is not null)
-            addSourceComments(instance, documentEnd: state.Comments?.Take());
+            addSourceComments(instance, documentEnd: state.Comments?.Consume());
 
         issues = Settings.ExceptionFilter is { } filter
             ? state.Errors.Remove(filter)
@@ -115,7 +115,7 @@ public class BaseFhirXmlDeserializer
     /// <remarks>The <see cref="ParserSettings.ExceptionFilter"/> influences which issues are returned.</remarks>
     public bool TryDeserializeElement(Type targetType, XmlReader reader, [NotNullWhen(true)] out Base? instance, out IEnumerable<CodedException> issues)
     {
-        PocoDeserializerState state = new() { Comments = Settings.RetainComments ? new() : null };
+        var state = createState();
 
         // If the stream has just been opened, move to the first token. (skip processing instructions, comments, whitespaces etc.)
         reader.MoveToContent(state);
@@ -126,7 +126,7 @@ public class BaseFhirXmlDeserializer
         instance = DeserializeElementInternal(targetType, reader, state);
 
         // Whatever is left after the root element has been read are the comments trailing the document.
-        addSourceComments(instance, documentEnd: state.Comments?.Take());
+        addSourceComments(instance, documentEnd: state.Comments?.Consume());
 
         issues = Settings.ExceptionFilter is { } filter
             ? state.Errors.Remove(filter)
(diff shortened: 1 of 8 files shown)


Kasdejong:
Pushed a follow-up commit consolidating the comment-buffering state (see thread below for context): `PocoDeserializerState.RetainComments`/`AddComment`/`TakePendingComments` are replaced by a single nullable `SourceCommentCollector? Comments` field, so every call site reads `state.Comments?.Add(...)` / `state.Comments?.Take()` — a null-conditional no-op when the setting is off, rather than a bool check hidden inside a method body. Also renamed `MoveToContentCapturingComments` to `MoveToContent(reader, state)` to match the existing `ReadToContent(reader, state)` naming, and collapsed the four duplicated `XmlReader` construction lines in `PocoDeserializationExtensions` into one helper.

No behavior change — full `Support.Poco.Tests` (603) and `Serialization.R4.Tests` (16234) pass unchanged before and after.

src/Hl7.Fhir.Base/Serialization/SourceCommentCollector.cs
```diff
@@ -0,0 +1,42 @@
...
+{
+    private List<string>? _pending;
+
+    public void Add(string comment) => (_pending ??= []).Add(comment);
+
+    /// <summary>
+    /// Returns the comments buffered since the previous call and clears the buffer, or <c>null</c> when
+    /// no comments were encountered.
+    /// </summary>
+    public string[]? Take()
```
andrzejskowronski: Slightly confusing naming when reading the code, maybe something like `Consume` or `Flush`?

src/Hl7.Fhir.Base/Serialization/BaseFhirXmlDeserializer.cs
```diff
@@ -83,15 +83,20 @@ private bool usesStrictCaseBinding()
     /// <remarks>The <see cref="ParserSettings.ExceptionFilter"/> influences which issues are returned.</remarks>
     public bool TryDeserializeResource(XmlReader reader, [NotNullWhen(true)] out Resource? instance, out IEnumerable<CodedException> issues)
     {
-        PocoDeserializerState state = new();
+        PocoDeserializerState state = new() { Comments = Settings.RetainComments ? new() : null };
```
andrzejskowronski: It's just 2 occurences right now so might be a nitpick, but I'd prefer an extension or something in case we add something more

andrzejskowronski (changes requested):

andrzejskowronski (approved):

PR #3564, opened 2026-08-02
Remove allocations from ClassMapping hot path

### LINKED ISSUE #3563: ClassMapping hot paths: per-access delegate allocation and repeated PropertyMappings snapshots during deserialization
## Summary

Profiling FHIR-JSON deserialization under a workload that parses ~900 patient bundles (Firely SDK 6.3.0; the affected files are byte-identical on current `develop`) shows two avoidable costs in `Hl7.Fhir.Base/Introspection/`, both on the per-element deserialization path:

1. **`ClassMapping.PrimitiveValueProperty` (line 218)** recomputes `PropertyMappings.SingleOrDefault(pm => pm.RepresentsValueElement)` on every access, for a value that is a pure function of the (already lazily built) mapping collection.

   The cost is larger than a LINQ scan. Enumerating `PropertyMappingCollection` goes to `ConcurrentDictionary.Values`, and that property **acquires every one of the dictionary's bucket locks and copies all entries into a fresh `List` wrapped in a `ReadOnlyCollection`** on each read. So each element parsed took a full, locked snapshot of the declaring class's property collection — twice over, since `HasPrimitiveValueMember` does the same thing with `Any`. In a thread-time profile of the bundle-parsing workload `get_PrimitiveValueProperty` was the single largest frame at **26.8% of active thread time**, and it accounted for ~93% of observed lock contention at 4-way parallelism. Those bucket locks — not any lazy-initialization check — are the contention: `LazyInitializer` takes no lock in this overload.

2. **`LazyInitializer.EnsureInitialized(ref field, factory)` allocates the factory delegate on every access**, including the warm path where the field has long been initialized. This applies wherever the factory captures state or is an instance method group, which is the case at `ClassMapping.PropertyMappingsInternal` (line 195, a `this`-capturing local function), `ClassMapping.CreateInstance`/`CreateList` (lines 361/375, method groups — so one delegate per created element and per created list), `PropertyMappingCollection.ByOrder`/`ChoiceProperties`, and `PropertyMapping.FhirTypeMappings` (read through `GetInstantiableType()` for every element whose property type is abstract).

   From the allocation census of the workload (15.55 GB sampled), the delegates alone are ~1.10 GB, **7.1% of everything the workload allocates**:

   | delegate type | site | allocated | share |
   |---|---|---|---|
   | `Func<PropertyMappingCollection>` | `PropertyMappingsInternal` | 602.1 MB | 3.87% |
   | `Func<Func<object>>` | `CreateInstance` | 384.3 MB | 2.47% |
   | `Func<Func<IList>>` | `CreateList` | 58.4 MB | 0.38% |
   | `Func<ClassMapping[]>` | `FhirTypeMappings` | 35.9 MB | 0.23% |
   | `Func<List<PropertyMapping>>` | `ByOrder` + `ChoiceProperties` | 17.4 MB | 0.11% |

   These are also the *only* `LazyInitializer` factories that appear in the census at all, which is a natural scope boundary: the other calls in the assembly either pass a non-capturing lambda (`() => []`), which the compiler caches in a static field so it costs nothing per access, or are not on the per-element parse path.

## Proposed fix (small, semantics-preserving)

- Cache the value-element scan inside `PropertyMappingCollection`, next to the `ByOrder`/`ChoiceProperties` caches and invalidated with them, and have `PrimitiveValueProperty`/`HasPrimitiveValueMember` read that. The value is a pure function of the collection's contents.
- Replace the `EnsureInitialized` calls with the allocation-free fast path
  `return Volatile.Read(ref _field) ?? create();`, where `create()` publishes via
  `Interlocked.CompareExchange(ref _field, built, null)`. This is the same shape `LazyInitializer.EnsureInitialized` itself has (`Volatile.Read(ref target) ?? EnsureInitializedCore(...)`), so it preserves its semantics exactly — the factory may race, one instance is published, the read has acquire semantics — while making the warm path a plain field read with zero allocation.

## Measured effect

Drop-in A/B on the workload above: **A** = `Hl7.Fhir.Base.dll` built from the `v6.3.0` tag, **B** = the same tree plus this fix. The three affected files are byte-identical between `v6.3.0` and current `develop`, so the port is exact and the result carries over; measuring against 6.3.0 is necessary because the consumer pins 6.3.0 and a `develop` build is not binary compatible with it. Both DLLs come from the same toolchain and are dropped into otherwise identical copies of the harness output directory.

Workload: 900 CMS measure cases from the Firely CQL SDK integration corpus, bundle deserialization **in** the loop (bundle cache off), single-threaded (dop 1, the deserialization path is serial), 100 reps per run, medians over reps 86–100. Four order-balanced rounds (A first in rounds 1 and 3, B first in 2 and 4); the pooled figure is the mean of the four round medians, which cancels the first-run-in-a-round penalty visible in the raw numbers.

| metric (per 900 cases) | A (6.3.0 stock) | B (+ fix) | delta | delta % |
|---|---|---|---|---|
| wall | 1635 ms | 1546 ms | −89 ms | **−5.4%** |
| cpu | 1731 ms | 1625 ms | −106 ms | **−6.2%** |
| allocated | 778.6 MB | 711.5 MB | −67.2 MB | **−8.6%** |
| allocated per case | 886.0 KB | 809.8 KB | −76.2 KB | **−8.6%** |
| gen0 collections | 47 | 43 | −4 | −8.5% |
| GC pause | 43.1 ms | 38.5 ms | −4.6 ms | −11% |
| gen1 / gen2 / lock contention | ~0 / 0 / 0 | 0 / 0 / 0 | — | — |

Round agreement:

* **Allocation and gen0 are reproducible**: −8.3% / −8.9% / −8.6% / −8.7% (allocation) and −8.5% in
  all four rounds (gen0). This is the part of the result to rely on.
* **Wall and CPU do not agree across rounds**: −15.4% / −3.2% / −0.1% / −2.0% (wall). The scatter
  tracks run order and first-run effects on this 4-core box, not the variant — round 1's stock run
  was simply a slow one (median 1769 ms against 1536–1635 ms in the other three rounds). −5.4% is
  the order-balanced mean, and it is worth exactly as much as that spread allows: the direction is
  consistent (every round favours the patch) but the magnitude is not resolved better than
  "a few percent".

  One cross-check: this A/B was run twice, before and after the review revisions (which added the
  `FhirTypeMappings` conversion and the `Volatile.Read`s). Four order-balanced rounds each time,
  eight in total. Pooled wall came out at −5.40% the first time and −5.44% the second, and pooled
  allocation at −8.3% then −8.6% — the extra ~0.3 points of allocation being the `FhirTypeMappings`
  delegate, consistent with the 0.23% share the census attributed to it.

### Microbenchmark of the paths themselves

Bytes allocated per call, measured with `GC.GetAllocatedBytesForCurrentThread()` on a warmed-up `boolean` mapping (200k iterations), `develop` vs. patched:

| path | develop | patched |
|---|---|---|
| `PropertyMappings` + `PrimitiveValueProperty` + `HasPrimitiveValueMember` + `FindMappedElementByName`, per iteration | 464 B | **0 B** |
| `FhirTypeMappings`, per read | 64 B | **0 B** |
| `CreateInstance`, per call | 128 B | **64 B** (just the `FhirBoolean`) |
| `CreateList`, per call | 96 B | **32 B** (just the `List`) |
| `FindMappedElementByChoiceName`, per call *(not addressed here)* | 392 B | 200 B |

The last row is a separate, untouched cost: `findChoiceMatch` does `ChoiceProperties.Where(...).ToList()` per call, which allocates a closure, an iterator and a list. It drops only because it no longer pays for the two delegates underneath it. Worth its own issue.

## Notes

- The workload is the Firely CQL SDK's CMS integration corpus; the effect is on *deserialization* (bundle parsing). CQL evaluation itself does not touch `ClassMapping` (verified by profiling with parsing removed from the loop).
- No public API change and no intended behavioural change; see the PR for the exact list of observable deltas.
- While implementing this, one latent bug turned up next door: `PropertyMappingCollection.Remove(x)` updated its cached lists by removing `x` from them, but the name dictionary is case-insensitive, so `Remove` can evict a *different* mapping that happens to share `x`'s name — leaving the evicted mapping behind in `ByOrder`/`ChoiceProperties`. The PR fixes that by dropping the caches instead.

alexzautke:
## Description

Two patterns in `ClassMapping`, `PropertyMappingCollection` and `PropertyMapping` cost an allocation
— and in one case a full, locked snapshot of a collection — on *every* access, while all of them sit
on the per-element path of (de)serialization. This PR removes both.

**1. `LazyInitializer.EnsureInitialized(ref field, factory)` allocates its factory on every call**,
including the warm path where the field has long been initialized. The local functions passed in
capture `this` (so the compiler cannot cache them in a static), and the method-group arguments
(`NativeType.BuildFactoryMethod`, `NativeType.BuildListFactoryMethod`, `buildFhirTypeMappings`)
allocate too — for `CreateInstance`/`CreateList`, once per element and per list created during a
parse.

Six sites are converted: `ClassMapping.PropertyMappingsInternal`, `ClassMapping.CreateInstance`,
`ClassMapping.CreateList`, `PropertyMappingCollection.ByOrder`, `PropertyMappingCollection.ChoiceProperties`
and `PropertyMapping.FhirTypeMappings` (reached via `GetInstantiableType()` for every element whose
property type is abstract). They now use:

```csharp
_field ?? LazyInitializer.EnsureInitialized(ref _field, factory)
```

The `??` short-circuits, so once the field is initialized an access is an ordinary read that never
reaches `EnsureInitialized` or the factory delegate it allocates per call; the delegate is only
created on the once-only cold path, which stays the standard BCL idiom. The ordinary read on the
warm path is safe per the documented .NET memory model
([Memory-model.md](https://github.com/dotnet/runtime/blob/main/docs/design/specs/Memory-model.md)):
object reference stores are release stores ("*Object assignment to a location potentially accessible
by other threads is a release with respect to accesses to the instance's fields/elements and
metadata*") and "*memory ordering honors data dependency*" on the reading side, so a thread that
observes the field non-null also observes the fully initialized object. The factory bodies are
unchanged, including the `DeclaringClass` validation performed when building the property mappings.
(Earlier revisions of this PR used a hand-rolled `Volatile.Read`/`Interlocked.CompareExchange`
pattern here; it was reworked to the shape above after review, for maintainability — see the review
thread.)

`ClassMapping.PropertyMappingsInternal` additionally uses the `EnsureInitialized` overload with a
`syncLock`, which double-checks under a lock so its factory runs *exactly once* (also a review
outcome): that member runs a possibly user-supplied `PropertyMapper` and builds the mutable
collection that callers manipulate through `PropertyMappings`, where a concurrently repeated factory
run would not be obviously harmless. The lock object itself is allocated on demand by
`LazyInitializer`. The other five sites use the lock-free two-argument overload: their factories are
idempotent and produce values that are never mutated afterwards, so its
publish-once-but-may-run-twice semantics are unobservable there. No deadlock exposure: building a
class's property mappings never reads another mapping's `PropertyMappings` (type resolution via
`FhirTypeMappings` is itself lazy — that is what breaks the type cycles).

**Scope — why these six and not the other eleven `LazyInitializer` calls in the assembly.** Only a
factory that captures state or is an instance method group allocates per access; a non-capturing
lambda such as `() => []` is cached by the compiler in a static field and is free. Of the remaining
eleven calls, five are non-capturing (`Base`, `DomNode`, `PocoNodeOrList` annotations/overflow), four
capture but are not on the per-element POCO parse path (`EnumMapping.PropertyMappings`,
`PropertyMapping`'s `IElementDefinitionSummary.Type`, `ElementNode.ChildDefinitions`,
`TypedElementOnSourceNode.Value`), and two (`PropertyMapping.Getter`/`Setter`) are inside a
`#if USE_GETTER_SETTER_AND_CODEGEN` that is not defined. An allocation census of the parsing workload
below confirms this empirically: the only `LazyInitializer` factories that allocate anything at all in
that workload are exactly the six converted here.

**2. `ClassMapping.PrimitiveValueProperty` re-ran a scan of the whole collection on every read.**
`PropertyMappings.SingleOrDefault(pm => pm.RepresentsValueElement)` looks like a cheap LINQ call, but
enumerating `PropertyMappingCollection` reads `ConcurrentDictionary.Values` — which takes *all* of the
dictionary's bucket locks and copies every entry into a fresh `List` inside a `ReadOnlyCollection`.
So every element parsed took a full, locked snapshot of its declaring class's property collection,
and `HasPrimitiveValueMember` (`Any`, same enumeration) took another. That was 26.8% of active thread
time in our profile, and ~93% of the lock contention at 4-way parallelism.

Both now read a list that `PropertyMappingCollection` computes once, next to the existing `ByOrder`
and `ChoiceProperties` caches and invalidated in the same places, so mutating a mapping's property
list is still reflected. `null` stays a valid, cached answer for classes without a value element, and
`SingleOrDefault` still resolves the value, so a malformed mapping with two value elements still
throws.

**3. Drive-by bug fix: `PropertyMappingCollection.Remove` left stale cache entries.** Mappings are
keyed by name *case-insensitively*, so `Remove(x)` evicts whatever mapping carries `x`'s name, which
need not be `x` itself. Updating the cached lists with `_byOrder?.Remove(x)` then removes nothing and
leaves the evicted mapping in the cache. This was already wrong for `ByOrder`/`ChoiceProperties`, and
caching the value element would have inherited it. `Remove` now calls `clearCaches()` instead, so the
lists are rebuilt from the name dictionary on the next read. Covered by a new test that fails without
this change.

## Observable behavioural deltas

No public API change and no *intended* behavioural change, but for completeness, everything that is
observably different:

* **Exception message in the duplicate case.** A (malformed) class declaring two value elements still
  throws `InvalidOperationException` from `PrimitiveValueProperty`, but the message now comes from
  `SingleOrDefault()` over the cached list ("Sequence contains more than one element") rather than
  from `SingleOrDefault(predicate)` ("…more than one matching element"). Pinned by a new test.
* **The property mapper runs exactly once.** Before this PR, racing first accesses to
  `PropertyMappings` could each run the property mapper, with only one result ever published. Now
  the mapper runs exactly once — a strictly narrower behaviour that only shows up for mappers with
  side effects.
* **Concurrent read-during-mutation of the property collection.** Previously each read of
  `PrimitiveValueProperty` took a point-in-time snapshot under the `ConcurrentDictionary`'s bucket
  locks; now it reads an unsynchronised cached list that is dropped on mutation, so a read racing a
  mutation may observe either side of it. This is exactly the guarantee `ByOrder` and
  `ChoiceProperties` have always had, and the collection was never safe for concurrent mutation anyway
  (`Add`/`AddRange`/`Clear`/`Remove` update the dictionary and clear the caches non-atomically).
* **Stale caches after `Remove` are fixed**, as described above — a behaviour change in the sense that
  a bug stopped happening.

## Related issues

Closes #3563 — ClassMapping hot paths: per-access delegate allocation and repeated
PropertyMappings snapshots during deserialization.

## Testing

Suites re-run after the change (including the review revisions):

* `Hl7.Fhir.Support.Tests` — **508 passed, 0 failed** (503 before, plus the new tests below).
* `Hl7.Fhir.Support.Poco.Tests` — **603 passed, 0 failed** (POCO JSON/XML (de)serialization).
* `Hl7.Fhir.Serialization.R4.Tests` — **16 234 passed, 0 failed**.

Seven tests added to `ClassMappingTest`, all of which exercise the changed paths directly:

| test | what it guards |
|---|---|
| `PrimitiveValuePropertyIsStableAndTracksChanges` | `FhirBoolean` → the `value` element; repeated reads return the same instance; remove/re-add is reflected; a class with no value element keeps returning `null` (the cached-`null` case) |
| `PrimitiveValuePropertyThrowsOnDuplicateValueElements` | a mapping with two value elements still throws, on the cached read as well as the first one, while `HasPrimitiveValueMember` still answers `true` |
| `PrimitiveValuePropertyIsInvalidatedByClearAndAddRange` | `Clear()` and `AddRange()` invalidate the value-element cache |
| `RemovingAMappingByNameInvalidatesThePrimitiveValueProperty` | the case-insensitive-`Remove` bug above — **fails without the `clearCaches()` fix** |
| `LazyMembersAreSafeToReadConcurrently` | 64 threads racing `PropertyMappings`, `PrimitiveValueProperty`, `CreateInstance` and `CreateList` on a freshly created, uninitialised mapping: all threads must observe the one published collection and the one value element, while the factories must still hand each of them a fresh instance/list |
| `PropertyMapperRunsExactlyOnceUnderConcurrentFirstAccess` | 64 threads racing first access against a counting property mapper: the mapper runs exactly once and every thread observes the same collection |
| `CreateInstanceAndCreateListReturnNewInstances` | the cached factories still hand out fresh instances and fresh lists |

(The existing `GetMappingsInParrallel` regression test for #556 races `ClassMapping.TryCreate` only and
never reads these members, and `CanManipulatePropertyMappingsList` checks `FindMappedElementByName`
rather than the caches — hence the new concurrency and invalidation tests.)

Additionally verified, outside the test suite:

* **Runtime allocation** on a warmed-up mapping, 200 000 iterations (re-verified after each review
  revision — the `??` short-circuit leaves the warm path at 0 B):

  | path | develop | this PR |
  |---|---|---|
  | `PropertyMappings` + `PrimitiveValueProperty` + `HasPrimitiveValueMember` + `FindMappedElementByName`, per iteration | 464 B | **0 B** |
  | `FhirTypeMappings`, per read | 64 B | **0 B** |
  | `CreateInstance`, per call | 128 B | **64 B** (just the `FhirBoolean`) |
  | `CreateList`, per call | 96 B | **32 B** (just the `List`) |

* **IL:** on `develop`, the six converted `EnsureInitialized` sites each contain an *unconditional*
  `ldftn; newobj` pair — a delegate allocated on every call — while
  `ClassMapping.PrimitiveValueProperty`/`HasPrimitiveValueMember` allocate their (non-capturing)
  predicate only once behind the usual `ldsfld; brtrue` cache, and pay per access through the calls
  underneath instead: `PropertyMappings` (one of the six delegates) plus the
  `ConcurrentDictionary.Values` snapshot. With this PR, the allocating instructions sit behind the
  `??`'s null check, on the branch that only executes while the field is still uninitialized; the
  0 B runtime figures above are the warm path's actual cost.

## Measured effect

Measured as a drop-in A/B in a downstream consumer (the Firely CQL SDK's CMS measure integration
corpus), because that is where the hot path was found. Note the caveat: the consumer pins **6.3.0**
and a `develop` build is not binary compatible with it, so the measurement uses `Hl7.Fhir.Base.dll`
built from the `v6.3.0` tag with and without this change. All three touched files are byte-identical
between `v6.3.0` and current `develop` (`git diff v6.3.0 develop --` is empty for them), and the
commit cherry-picks onto the tag with an identical diff, so the measurement carries over.

Workload: 900 measure cases, FHIR-JSON bundle deserialization **in** the loop, single-threaded (the
deserialization path is serial), 100 reps per run, medians over reps 86–100, four order-balanced
rounds.

| metric (per 900 cases) | A (6.3.0 stock) | B (+ fix) | delta | delta % |
|---|---|---|---|---|
| wall | 1635 ms | 1546 ms | −89 ms | **−5.4%** |
| cpu | 1731 ms | 1625 ms | −106 ms | **−6.2%** |
| allocated | 778.6 MB | 711.5 MB | −67.2 MB | **−8.6%** |
| allocated per case | 886.0 KB | 809.8 KB | −76.2 KB | **−8.6%** |
| gen0 collections | 47 | 43 | −4 | −8.5% |
| GC pause | 43.1 ms | 38.5 ms | −4.6 ms | −11% |
| gen1 / gen2 / lock contention | ~0 / 0 / 0 | 0 / 0 / 0 | — | — |

Round agreement:

* **Allocation and gen0 are reproducible**: −8.3% / −8.9% / −8.6% / −8.7% (allocation) and −8.5% in
  all four rounds (gen0). This is the part of the result to rely on.
* **Wall and CPU do not agree across rounds**: −15.4% / −3.2% / −0.1% / −2.0% (wall). The scatter
  tracks run order and first-run effects on this 4-core box, not the variant — round 1's stock run
  was simply a slow one (median 1769 ms against 1536–1635 ms in the other three rounds). −5.4% is
  the order-balanced mean, and it is worth exactly as much as that spread allows: the direction is
  consistent (every round favours the patch) but the magnitude is not resolved better than
  "a few percent".

  One cross-check: this A/B was run twice, before and after the first round of review revisions
  (which added the `FhirTypeMappings` conversion). Four order-balanced rounds each time, eight in
  total. Pooled wall came out at −5.40% the first time and −5.44% the second, and pooled allocation
  at −8.3% then −8.6% — the extra ~0.3 points of allocation being the `FhirTypeMappings` delegate,
  consistent with the 0.23% share the census attributed to it.

Correctness gate on the same corpus: all 900 cases across 20 measures complete with exit code 0 and
zero exceptions, both with and without the patch.

## FirelyTeam Checklist
      (not applicable — no public API change; see "Observable behavioural deltas" above)

---
_Generated by [Claude Code](https://claude.ai/code/session_01Kj6AseHiRtRbAEN6yRzK1t)_

### DIFF
--- a/src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
+++ b/src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
@@ -187,54 +187,35 @@ public class ClassMapping(
     // This list is created lazily. This not only improves initial startup time of
     // applications but also ensures circular references between types will not cause loops.
     private PropertyMappingCollection? _mappings;
-    private readonly object _mappingsLock = new();
+    private object? _mappingsLock;
 
-    // Note: this member - like the other lazily initialized members below, and the equivalent
-    // members in PropertyMapping and PropertyMappingCollection - deliberately does not use
-    // LazyInitializer.EnsureInitialized(). Passing that a factory allocates a delegate on *every*
-    // access, including the warm path where the value has long been initialized, and these members
-    // are read per element on the (de)serialization path. (This only applies to factories that
-    // capture state or are instance method groups: a non-capturing lambda such as () => [] is
-    // cached by the compiler in a static field, so the LazyInitializer calls elsewhere in this
-    // assembly cost nothing per access and are left alone.)
+    // Note on the `_field ?? LazyInitializer.EnsureInitialized(...)` shape of this member, the
+    // other lazily initialized members below and the equivalents in PropertyMapping and
+    // PropertyMappingCollection: calling EnsureInitialized unconditionally would allocate its
+    // factory delegate on *every* access, including the warm path where the field has long been
+    // initialized, and these members are read per element on the (de)serialization path. The
+    // null-coalescing read short-circuits that: once the field is initialized, an access is an
+    // ordinary read that allocates nothing, and the factory delegate is only created on the
+    // once-only cold path. The ordinary read is enough because the runtime guarantees that
+    // object reference stores are release stores and that data-dependent reads are ordered, so a
+    // thread observing the field non-null also observes the fully initialized object; see
+    // https://github.com/dotnet/runtime/blob/main/docs/design/specs/Memory-model.md.
     //
-    // On the warm path all of these members are a single Volatile.Read, whose acquire semantics
-    // guarantee that a caller observing the field also observes the fully constructed object on
-    // weakly ordered architectures. The cold paths differ:
-    //
-    // This member builds the mutable collection that callers may manipulate through
-    // PropertyMappings, and it runs the (possibly user-supplied) property mapper to do so. Racing
-    // threads must therefore not each run the mapper and build their own copy - the lock, taken
-    // only until the field is first published, makes initialization run exactly once.
-    //
-    // The other lazy members (CreateInstance/CreateList below, and the equivalents in
-    // PropertyMapping and PropertyMappingCollection) have idempotent factories producing values
-    // that are never mutated afterwards, so they keep LazyInitializer's lock-free semantics
-    // instead: when threads race, the factory may run more than once, but only a single result is
-    // ever published (via Interlocked.CompareExchange) and every caller receives that one
-    // instance - a duplicate run is unobservable there.
-    private PropertyMappingCollection PropertyMappingsInternal
+    // This member uses the EnsureInitialized overload with a syncLock, so that the (possibly
+    // user-supplied) property mapper building this mutable collection runs exactly once. The other
+    // lazy members have idempotent factories whose results are never mutated afterwards, so they
+    // use the lock-free overload, where racing factories may run concurrently but only a single
+    // result is ever published and handed out.
+    private PropertyMappingCollection PropertyMappingsInternal =>
+        _mappings ?? LazyInitializer.EnsureInitialized(ref _mappings, ref _mappingsLock, createPropertyMappings)!;
+
+    private PropertyMappingCollection createPropertyMappings()
     {
-        get
-        {
-            return Volatile.Read(ref _mappings) ?? createCollection();
+        var properties = propertyMapper(this).ToList();
+        if(properties.FirstOrDefault(m => m.DeclaringClass != this) is {} errorMapping)
+            throw new InvalidOperationException($"PropertyMapping '{errorMapping.Name}' is already used for another ClassMapping '{errorMapping.DeclaringClass.Name}'.");
 
-            PropertyMappingCollection createCollection()
-            {
-                lock (_mappingsLock)
-                {
-                    if (_mappings is { } existing) return existing;
-
-                    var properties = propertyMapper(this).ToList();
-                    if(properties.FirstOrDefault(m => m.DeclaringClass != this) is {} errorMapping)
-                        throw new InvalidOperationException($"PropertyMapping '{errorMapping.Name}' is already used for another ClassMapping '{errorMapping.DeclaringClass.Name}'.");
-
-                    var created = new PropertyMappingCollection(properties);
-                    Volatile.Write(ref _mappings, created);
-                    return created;
-                }
-            }
-        }
+        return new PropertyMappingCollection(properties);
     }
 
     /// <summary>
@@ -390,16 +371,10 @@ public class ClassMapping(
     /// <remarks>If not set, the default constructor for the <see cref="NativeType"/> will be used.</remarks>
     public Base CreateInstance()
     {
-        var factory = Volatile.Read(ref _factory) ?? createFactory();
+        var factory = _factory ?? LazyInitializer.EnsureInitialized(ref _factory, NativeType.BuildFactoryMethod)!;
         var newInstance = factory();
         if (newInstance is IDynamicType idt) idt.DynamicTypeName = Name;
         return (Base)newInstance;
-
-        Func<object> createFactory()
-        {
-            var created = NativeType.BuildFactoryMethod();
-            return Interlocked.CompareExchange(ref _factory, created, null) ?? created;
-        }
     }
 
     private Func<object>? _factory;
@@ -410,14 +385,8 @@ public class ClassMapping(
     /// <remarks>If not set, the default List constructor for the <see cref="NativeType"/> will be used.</remarks>
     public IList CreateList()
     {
-        var factory = Volatile.Read(ref _listFactory) ?? createFactory();
+        var factory = _listFactory ?? LazyInitializer.EnsureInitialized(ref _listFactory, NativeType.BuildListFactoryMethod)!;
         return factory();
-
-        Func<IList> createFactory()
-        {
-            var created = NativeType.BuildListFactoryMethod();
-            return Interlocked.CompareExchange(ref _listFactory, created, null) ?? created;
-        }
     }
 
     private Func<IList>? _listFactory;
--- a/src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
+++ b/src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
@@ -187,6 +187,7 @@ public class ClassMapping(
     // This list is created lazily. This not only improves initial startup time of
     // applications but also ensures circular references between types will not cause loops.
     private PropertyMappingCollection? _mappings;
+    private readonly object _mappingsLock = new();
 
     // Note: this member - like the other lazily initialized members below, and the equivalent
     // members in PropertyMapping and PropertyMappingCollection - deliberately does not use
@@ -197,11 +198,21 @@ public class ClassMapping(
     // cached by the compiler in a static field, so the LazyInitializer calls elsewhere in this
     // assembly cost nothing per access and are left alone.)
     //
-    // The pattern used instead has the same semantics as LazyInitializer for reference types: the
-    // factory may run more than once when threads race, but only a single instance is ever
-    // published and every caller receives that one instance. The Volatile.Read on the fast path
-    // provides the same acquire semantics, so a caller that observes the field also observes the
-    // fully constructed object on weakly ordered architectures.
+    // On the warm path all of these members are a single Volatile.Read, whose acquire semantics
+    // guarantee that a caller observing the field also observes the fully constructed object on
+    // weakly ordered architectures. The cold paths differ:
+    //
+    // This member builds the mutable collection that callers may manipulate through
+    // PropertyMappings, and it runs the (possibly user-supplied) property mapper to do so. Racing
+    // threads must therefore not each run the mapper and build their own copy - the lock, taken
+    // only until the field is first published, makes initialization run exactly once.
+    //
+    // The other lazy members (CreateInstance/CreateList below, and the equivalents in
+    // PropertyMapping and PropertyMappingCollection) have idempotent factories producing values
+    // that are never mutated afterwards, so they keep LazyInitializer's lock-free semantics
+    // instead: when threads race, the factory may run more than once, but only a single result is
+    // ever published (via Interlocked.CompareExchange) and every caller receives that one
+    // instance - a duplicate run is unobservable there.
     private PropertyMappingCollection PropertyMappingsInternal
     {
         get
@@ -210,12 +221,18 @@ public class ClassMapping(
 
             PropertyMappingCollection createCollection()
             {
-                var properties = propertyMapper(this).ToList();
-                if(properties.FirstOrDefault(m => m.DeclaringClass != this) is {} errorMapping)
-                    throw new InvalidOperationException($"PropertyMapping '{errorMapping.Name}' is already used for another ClassMapping '{errorMapping.DeclaringClass.Name}'.");
+                lock (_mappingsLock)
+                {
+                    if (_mappings is { } existing) return existing;
 
-                var created = new PropertyMappingCollection(properties);
-                return Interlocked.CompareExchange(ref _mappings, created, null) ?? created;
+                    var properties = propertyMapper(this).ToList();
+                    if(properties.FirstOrDefault(m => m.DeclaringClass != this) is {} errorMapping)
+                        throw new InvalidOperationException($"PropertyMapping '{errorMapping.Name}' is already used for another ClassMapping '{errorMapping.DeclaringClass.Name}'.");
+
+                    var created = new PropertyMappingCollection(properties);
+                    Volatile.Write(ref _mappings, created);
+                    return created;
+                }
             }
         }
     }
--- a/src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
+++ b/src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
@@ -188,11 +188,25 @@ public class ClassMapping(
     // applications but also ensures circular references between types will not cause loops.
     private PropertyMappingCollection? _mappings;
 
+    // Note: this member - like the other lazily initialized members below, and the equivalent
+    // members in PropertyMapping and PropertyMappingCollection - deliberately does not use
+    // LazyInitializer.EnsureInitialized(). Passing that a factory allocates a delegate on *every*
+    // access, including the warm path where the value has long been initialized, and these members
+    // are read per element on the (de)serialization path. (This only applies to factories that
+    // capture state or are instance method groups: a non-capturing lambda such as () => [] is
+    // cached by the compiler in a static field, so the LazyInitializer calls elsewhere in this
+    // assembly cost nothing per access and are left alone.)
+    //
+    // The pattern used instead has the same semantics as LazyInitializer for reference types: the
+    // factory may run more than once when threads race, but only a single instance is ever
+    // published and every caller receives that one instance. The Volatile.Read on the fast path
+    // provides the same acquire semantics, so a caller that observes the field also observes the
+    // fully constructed object on weakly ordered architectures.
     private PropertyMappingCollection PropertyMappingsInternal
     {
         get
         {
-            return LazyInitializer.EnsureInitialized(ref _mappings, createCollection)!;
+            return Volatile.Read(ref _mappings) ?? createCollection();
 
             PropertyMappingCollection createCollection()
             {
@@ -200,7 +214,8 @@ public class ClassMapping(
                 if(properties.FirstOrDefault(m => m.DeclaringClass != this) is {} errorMapping)
                     throw new InvalidOperationException($"PropertyMapping '{errorMapping.Name}' is already used for another ClassMapping '{errorMapping.DeclaringClass.Name}'.");
 
-                return new PropertyMappingCollection(properties);
+                var created = new PropertyMappingCollection(properties);
+                return Interlocked.CompareExchange(ref _mappings, created, null) ?? created;
             }
         }
     }
@@ -215,7 +230,7 @@ public class ClassMapping(
     /// property will also be present in the PropertyMappings collection. If this class has
     /// no such property, it is null.
     /// </summary>
-    public PropertyMapping? PrimitiveValueProperty => PropertyMappings.SingleOrDefault(pm => pm.RepresentsValueElement);
+    public PropertyMapping? PrimitiveValueProperty => PropertyMappingsInternal.PrimitiveValueProperty;
 
     /// <summary>
     /// This indicates that this class is representing the Patient data (and implements <see cref="IPatient"/>).
@@ -225,7 +240,7 @@ public class ClassMapping(
     /// <summary>
     /// Whether the reflected type has a member that represent a primitive value.
     /// </summary>
-    public bool HasPrimitiveValueMember => PropertyMappings.Any(pm => pm.RepresentsValueElement);
+    public bool HasPrimitiveValueMember => PropertyMappingsInternal.HasPrimitiveValueMember;
 
     /// <summary>
     /// Returns the mapping for an element of this class by its name.
@@ -358,10 +373,16 @@ public class ClassMapping(
     /// <remarks>If not set, the default constructor for the <see cref="NativeType"/> will be used.</remarks>
     public Base CreateInstance()
     {
-        var factory = LazyInitializer.EnsureInitialized(ref _factory, NativeType.BuildFactoryMethod)!;
+        var factory = Volatile.Read(ref _factory) ?? createFactory();
         var newInstance = factory();
         if (newInstance is IDynamicType idt) idt.DynamicTypeName = Name;
         return (Base)newInstance;
+
+        Func<object> createFactory()
+        {
+            var created = NativeType.BuildFactoryMethod();
+            return Interlocked.CompareExchange(ref _factory, created, null) ?? created;
+        }
     }
 
     private Func<object>? _factory;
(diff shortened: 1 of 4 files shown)


src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
```diff
@@ -188,19 +188,34 @@ internal static bool TryCreate(ModelInspector parent, Type type, [NotNullWhen(tr
...
+    // capture state or are instance method groups: a non-capturing lambda such as () => [] is
+    // cached by the compiler in a static field, so the LazyInitializer calls elsewhere in this
+    // assembly cost nothing per access and are left alone.)
+    //
+    // The pattern used instead has the same semantics as LazyInitializer for reference types: the
+    // factory may run more than once when threads race, but only a single instance is ever
+    // published and every caller receives that one instance. The Volatile.Read on the fast path
+    // provides the same acquire semantics, so a caller that observes the field also observes the
+    // fully constructed object on weakly ordered architectures.
     private PropertyMappingCollection PropertyMappingsInternal
```
andrzejskowronski: Not opposed to change in general, but this one is particularly dangerous I think, in this unlikely case where two threads access it while it's not yet initialized, both of them might modify the collection in some way but only one will propagate the changes in the end.
I don't think that differs from the previous behavior either though...
I'd hope to get @ewoutkramer thoughts on that ;)

alexzautke: Good point — addressed in c181443.

Strictly speaking, the scenario where two threads end up modifying *different* collections can't occur with the CAS pattern: the losing thread's freshly built collection is discarded before it is ever returned, so every caller receives the single published instance (the same semantics `LazyInitializer.EnsureInitialized` had here before). But the underlying discomfort is legitimate: the factory could run twice concurrently, and this particular member runs a possibly **user-supplied** `PropertyMapper` and builds the mutable collection that callers manipulate through `PropertyMappings`. Better to make it obviously safe than subtly safe.

`PropertyMappingsInternal` now uses double-checked locking: the warm path is unchanged (a single `Volatile.Read`, no lock, no allocation — re-verified at 0 B per access on a warmed-up mapping), and only the one-time cold path takes a lock, so the property mapper runs exactly once. No deadlock exposure: building a class's property mappings never reads another mapping's `PropertyMappings` (type resolution via `FhirTypeMappings` is itself lazy — that's what breaks the type cycles). The only added cost is one lock object per `ClassMapping` instance.

The other lazy members deliberately keep the lock-free pattern: their factories are idempotent and produce values that are never mutated afterwards (delegates, read-only lists), so a duplicate run is unobservable there. The comment on the member now spells out this distinction.

Also added `PropertyMapperRunsExactlyOnceUnderConcurrentFirstAccess`, which races 64 threads on first access against a counting mapper and asserts the mapper ran exactly once and all threads observed the same collection — it fails on the previous CAS version.

---
_Generated by [Claude Code](https://claude.ai/code)_

ewoutkramer:
I am wondering whether that whole Volatile.Read stuff is worth doing. It feels like a micro-optimalization that introduces concepts that we are not familiar with, replacing accepted best practices like LazyInitializer use. And yes, there's a small delegate allocation there, but I can't believe that's worth doing. What say you @andrzejskowronski ?

ewoutkramer:
@copilot - can you comment on whether these changes provide performance benefits that outweigh the added complexity of dealing with locks?

ewoutkramer:
So, copilot, GPT and Alexander say this is worth doing. Anyway, GPT has a suggestion to make this far more readable by removing the manual locking:

 > So I would propose: retain the initial Volatile.Read, but let LazyInitializer handle the cold path. That delivers 
 > essentially the same allocation saving with much less bespoke concurrency logic. The Volatile.Read is still needed for safe
 > lock-free publication, but Interlocked.CompareExchange and the handwritten lock can disappear from most or all call sites:
 >
 >   return Volatile.Read(ref _field)
      ?? LazyInitializer.EnsureInitialized(ref _field, createValue)!;

src/Hl7.Fhir.Base/Introspection/ClassMapping.cs
alexzautke: Following up on the maintainability discussion: 4ac4fdf replaces the hand-rolled `Volatile.Read`/`Interlocked.CompareExchange` pattern at all six sites with the suggested shape

```csharp
_field ?? LazyInitializer.EnsureInitialized(ref _field, factory)
```

This keeps both goals. The `??` short-circuits, so once the field is initialized an access is an ordinary read that never reaches `EnsureInitialized` or its per-call factory delegate — the warm path re-measures at 0 B per access. The cold path is the standard BCL idiom, and there is no `Volatile`, `Interlocked` or local-function CAS left in the diff.

The ordinary read on the warm path is safe per the documented .NET memory model ([Memory-model.md](https://github.com/dotnet/runtime/blob/main/docs/design/specs/Memory-model.md)): *"Object assignment to a location potentially accessible by other threads is a release with respect to accesses to the instance's fields/elements and metadata"*, and *"memory ordering honors data dependency"* on the reading side — so a thread that observes the field non-null also observes the fully initialized object, without needing `Volatile.Read`.

The exactly-once concern from this thread stays addressed, now via the BCL itself: `PropertyMappingsInternal` uses the `EnsureInitialized(ref target, ref syncLock, factory)` overload, which double-checks under a lock so the (possibly user-supplied) property mapper runs exactly once. That overload even allocates the lock object on demand, so the per-`ClassMapping` lock field is just a null reference until first use. The other five members use the lock-free two-argument overload since their factories are idempotent. `PropertyMapperRunsExactlyOnceUnderConcurrentFirstAccess` (64 threads racing first access) still pins the exactly-once behaviour.

Two footnotes: the netstandard2.1 reference assembly for `LazyInitializer` predates nullable annotations, so the calls carry a `!` to compile with warnings-as-errors; and a null-returning factory now throws `InvalidOperationException` again, exactly like the original `LazyInitializer` code — one less behavioural delta.

Re-verified on the new commit: both target frameworks build clean; `Support.Tests` 508, `Support.Poco.Tests` 603, `Serialization.R4.Tests` 16 234 — all passed.

---
_Generated by [Claude Code](https://claude.ai/code)_

alexzautke: To confirm how 4ac4fdf relates to the "retain the initial `Volatile.Read`, but let `LazyInitializer` handle the cold path" proposal: structurally that is exactly what it does — `LazyInitializer.EnsureInitialized` is the cold path at all six sites, and `Interlocked.CompareExchange` and the handwritten lock are gone from every call site (the exactly-once guarantee for `PropertyMappingsInternal` comes from `EnsureInitialized`'s own `syncLock` overload, not from a lock of ours).

The one deliberate difference: the warm path is an ordinary read (`_field ?? ...`), not `Volatile.Read(ref _field) ?? ...`. The proposal's premise that `Volatile.Read` is still needed for safe lock-free publication doesn't hold on the reader side under the documented .NET memory model ([Memory-model.md](https://github.com/dotnet/runtime/blob/main/docs/design/specs/Memory-model.md)): publication safety comes from the writer side — object reference stores are release stores ("*object assignment to a location potentially accessible by other threads is a release*"; inside `EnsureInitialized` it is a `Volatile.Write` anyway) — combined with "*memory ordering honors data dependency*" when reading through the obtained reference. So a thread that reads the field non-null, even with a plain read, is guaranteed to observe the fully initialized object. The doc's "publication via ordinary assignment" example is this exact pattern.

That said, `Volatile.Read` on the warm path would be equally correct, merely more conservative — it costs nothing on x64 and a single acquire-load on ARM64. If you prefer it for explicitness (it documents the cross-thread handoff at the read site rather than relying on the memory-model note in the comment), it is a one-word change per site — happy to switch.

---
_Generated by [Claude Code](https://claude.ai/code)_

ewoutkramer (approved):

alexzautke:
Regarding the proposal to "retain the initial `Volatile.Read`, but let `LazyInitializer` handle the cold path": structurally that is exactly what 4ac4fdf does — `LazyInitializer.EnsureInitialized` is the cold path at all six sites, and `Interlocked.CompareExchange` and the handwritten lock are gone from every call site (the exactly-once guarantee for `PropertyMappingsInternal` comes from `EnsureInitialized`'s own `syncLock` overload, not from a lock of ours).

The one deliberate difference: the warm path is an ordinary read (`_field ?? ...`), not `Volatile.Read(ref _field) ?? ...`. The proposal's premise that `Volatile.Read` is still needed for safe lock-free publication doesn't hold on the reader side under the documented .NET memory model ([Memory-model.md](https://github.com/dotnet/runtime/blob/main/docs/design/specs/Memory-model.md)): publication safety comes from the writer side — object reference stores are release stores ("*object assignment to a location potentially accessible by other threads is a release*"; inside `EnsureInitialized` it is a `Volatile.Write` anyway) — combined with "*memory ordering honors data dependency*" when reading through the obtained reference. So a thread that reads the field non-null, even with a plain read, is guaranteed to observe the fully initialized object. The doc's "publication via ordinary assignment" example is this exact pattern.

That said, `Volatile.Read` on the warm path would be equally correct, merely more conservative — it costs nothing on x64 and a single acquire-load on ARM64. If you prefer it for explicitness (it documents the cross-thread handoff at the read site rather than relying on the memory-model note in the comment), it is a one-word change per site — happy to switch.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #3568, opened 2026-08-05
Omit empty JSON and XML structures without buffering (GPT)

ewoutkramer:
## Problem

The serializers can discover that a complex value has no serializable children only after filtering has begun. At that point the underlying forward-only writer may already have emitted the opening JSON object/array token or XML start element, leaving an empty structure that should have been omitted.

PR #3557 solved this by buffering serialized output so the opening token could effectively be undone. That works, but makes filtering, serialization, and buffer management interact inside the serializer and obscures which layer owns each decision.

## Conclusion

Full payload buffering is not necessary. The only output that must be reversible is structural output that has not yet been proven necessary. This PR encapsulates that concern in small deferred writers:

- the serializer still decides what to traverse and filter;
- the normal JSON and XML writers still perform actual serialization;
- the deferred writer owns only whether pending structural tokens must be committed.

The deferred state is proportional to nesting depth, not payload size. A container/element is committed when real serializable content requires it; otherwise its pending start is discarded when its scope closes.

### JSON

- Adds a `DeferredJsonWriter` that defers object/array starts and property names.
- Omits complex values made empty by filtering.
- Preserves sparse primitive-array positions where JSON/FHIR semantics require placeholders.
- Fixes the indentation regression by keeping indentation decisions in the actual writer rather than replaying buffered JSON.

### XML

- Adds a `DeferredXmlWriter` with the same structural responsibility.
- Commits an element when it receives an attribute, a surviving child, or raw XHTML content.
- Keeps the document resource root required while allowing empty nested resources and complex elements to disappear.
- Treats retained source comments as metadata: comments follow a surviving element but do not materialize an otherwise empty element.

## Behavioral impact

- Empty JSON objects/arrays and XML elements created only because all children were filtered are omitted.
- Primitive values, XML attributes, XHTML, and required roots continue to serialize.
- JSON indentation remains correct.
- Invalid-resource snapshots now omit empty invalid XML containers/elements; their captured diagnostics are unchanged.

## Validation

- Full POCO support test suite: 610 passed.
- Added JSON and XML regressions for filtered complex values and content-preserving cases.
- Added XML coverage for attributes, raw content, roots, nested resources, and retained comments.
- `Hl7.Fhir.Base` builds for both `net8.0` and `netstandard2.1`.


ewoutkramer:
I prefer Claude's, closing this one.

PR #3572, opened 2026-08-10
Halve the allocations of parse-time validation

cknaap:
## Description

Model validation runs on every property and every object of a parse, and is on in every deserialization mode but `SyntaxOnly`/`Ostrich`, so it sits on the hot path of every consumer of the SDK.

Profiling a StructureDefinition-heavy workload showed a parse allocating **50-65x the payload** with the validator on, against **18-23x** with it off, while the POCO graph it returns is only ~4x the payload. In other words, most of what a parse allocates is transient garbage produced by validating - not the graph the caller asked for. Roughly two thirds of the total came from a handful of per-element allocations in the validation path.

This removes that per-element churn. **No validation outcome changes** - no error is added, removed or reworded, and no public API changes.

- The Json deserializer dispatches validation without a closure. An immediate validation calls a method; a delayed one - every FHIR primitive, which is the bulk of the elements of a resource - stores a `DelayedValidation` struct in the object context. Capturing the arguments instead put them in a display class, which the compiler allocates *on entry* to the two validation methods - before the check whether a validator is configured at all. See the note under the table: that unconditional allocation, not the delegate per delayed validation, is what this actually saves.
- Resolving the mapping for an element no longer pushes a path part just so a diagnostic *could* name the element; the path is built in the error branches that need it.
- `FhirAttributeValidator.ValidateObject` only visits the elements that have a minimum cardinality, cached per class (`ClassMapping.MandatoryElements`), instead of running an `OfType`/`SingleOrDefault` over the validation attributes of *every* element of the class, per validated instance. This is the largest single win.
- Attribute validation no longer allocates enumerators and a result array per property/object, and the error lists are only created once there is an error to report.
- The Json object context creates its dictionaries lazily.

`PocoValidationContext` (one allocation per property and per object) is deliberately left alone: it is public, a `record`, mutated through `MemberName ??=` and used as `PocoValidationContext?`, so turning it into a `readonly struct` would be a breaking change - for a good deal less than the above is worth.

An earlier revision also memoized the `Func<string>` that `PathPart` hands to the validators, so the delegate was created once per position instead of once per validation. That was worth 115.8 KB of the SD parse (4.3%), but its safety rested on the parts being immutable and on `SetIndex` not copying the cache onto a part representing a different position - too much subtlety to maintain for the gain, so it was reverted on review (403f7c3). `PathPart.cs` and `BaseFhirXmlDeserializer.cs` are unchanged from `develop`.

## Related issues

No issue; this came out of an allocation analysis of the parse phase done while parallelizing StructureDefinition compilation in another repo. The same 65% is paid per request, concurrently, on Vonk's request path (`FhirXmlJsonParser.Build` always installs a validator), which is why this is worth doing upstream rather than locally.

## Testing

Added `ValidatingDeserializationBenchmarks`, which parses a small CodeSystem and a Patient StructureDefinition with the validator on and off, so the cost of validation is visible as the difference between the two. Both payloads are taken from the shipped `specification.zip` and the setup refuses to run on data that does not parse cleanly, so neither variant pays for constructing errors.

Measured against `develop` with the same benchmark (R5, net8.0, Release, BenchmarkDotNet), runs repeated interleaved:

| | develop | this PR | Δ |
|---|---|---|---|
| Patient SD (88 KB), validating | 5,237.8 KB / 3.33-3.42 ms | 2,724.6 KB / 1.64-1.96 ms | **-48% alloc, -45% time** |
| Patient SD, validator off | 1,414.2 KB / 0.84-0.93 ms | 979.3 KB / too noisy to call | -31% alloc |
| CodeSystem (2.8 KB), validating | 157.7 KB / 88.9 µs | 88.0 KB / 55-63 µs | **-44% alloc, -29% time** |
| CodeSystem, validator off | 46.0 KB / 21.7 µs | 33.1 KB / 17.7-19.8 µs | -28% alloc |

The validator's own marginal allocation - the thing this PR targets - drops **54%** (3,823 KB to 1,745 KB on the SD). The SD parse went from 59x to 31x its payload. Allocation figures were bit-identical across every run; the time win is far outside the 5-10% run-to-run spread. The validator-off *timing* is not callable either way: that path is dominated by gen2 collections of the 88 KB LOH buffer that `Utf8JsonReaderFromJsonText` allocates, and it swung both ways between runs.

The allocation figures in both columns were re-measured head to head after the revert described above, in one session and with the same toolchain; the `develop` column reproduces to the byte. The timings span the runs done before and after that revert, which is why they are given as ranges - the revert costs about 4% of the allocation and nothing measurable in time.

### Why the validator-off rows improve at all

Switching the validator off should make this PR a no-op, and it does not, because part of the churn was never conditional on there being a validator. `doPropertyValidation` and `doObjectValidation` captured their arguments in a closure, and the compiler allocates that display class on entry to the method - before the `if (Settings.Validator is null) return;` guard - so a parse with validation switched off still allocated one per property and per object.

Reverting only that hunk on top of the rest of the branch puts the SD's validator-off figure back at 1,240.2 KB, so it accounts for 260.8 KB of the 434.9 KB that row gains; the remainder is the lazily created dictionaries and the path parts no longer pushed for diagnostics that are never produced.

The same isolation shows this change is close to a wash on the *validating* path (2,654.5 KB to 2,608.8 KB, -1.7%): `DelayedValidation` is stored inline in the scheduling dictionary, and the larger entries offset most of what the display class and the delegate cost. It is the unconditional allocation that makes it worth doing, which is the opposite of what one would guess.

Test suites passing on this change: `Hl7.Fhir.Serialization.R5.Tests` (16,630), `Hl7.Fhir.Serialization.R4.Tests` (16,234), `Hl7.Fhir.Serialization.STU3.Tests` (13,405), `Hl7.Fhir.Support.Poco.Tests` (619), `Hl7.Fhir.Support.Tests` (508). The failures in `Hl7.Fhir.R4.Tests` (3-4) and `Hl7.Fhir.Specification.R4.Tests` (8) are environmental and unrelated - SSL failures against public test servers, external terminology services, and one test with a hardcoded path from another machine.

## FirelyTeam Checklist

### DIFF
--- a/src/Hl7.Fhir.Base/Introspection/PropertyMapping.cs
+++ b/src/Hl7.Fhir.Base/Introspection/PropertyMapping.cs
@@ -288,26 +288,38 @@ public class PropertyMapping : IElementDefinitionSummary
     /// The collection of zero or more <see cref="ValidationAttribute"/> (or subclasses) declared
     /// on this property.
     /// </summary>
-    public ValidatingFhirModelAttribute[] ValidationAttributes { get; init; } = [];
+    /// <remarks>Like the other metadata on a mapping, these attributes are a snapshot: the derived
+    /// <see cref="IsMandatoryElement"/> and <see cref="MandatoryCardinality"/> are determined when the
+    /// mapping is constructed, so mutating an attribute afterwards - e.g. setting
+    /// <see cref="CardinalityAttribute.Min"/> on an attribute that was handed to a mapping, or replacing
+    /// an entry of the array - is not picked up. Build a mapping with the attributes it should have.</remarks>
+    public ValidatingFhirModelAttribute[] ValidationAttributes
+    {
+        get => _validationAttributes;
+        init
+        {
+            _validationAttributes = value;
+
+            // Validating an object means checking, for every one of its elements, whether a mandatory
+            // element is missing - so this scan would otherwise be done per element per validated
+            // instance. It is done here instead, at the same moment IsMandatoryElement is derived from
+            // the very same attribute (see TryCreate).
+            _mandatoryCardinality = value.OfType<CardinalityAttribute>().SingleOrDefault() is { Min: > 0 } cardinality
+                ? [cardinality]
+                : [];
+        }
+    }
+
+    private ValidatingFhirModelAttribute[] _validationAttributes = [];
 
     /// <summary>
     /// The <see cref="CardinalityAttribute"/> declared on this element when it has a minimum cardinality
     /// higher than 0, as a (single-entry) list of attributes to validate; an empty list otherwise.
     /// </summary>
-    /// <remarks>Validating an object means checking, for every one of its elements, whether a mandatory
-    /// element is missing - so this scan of <see cref="ValidationAttributes"/> would otherwise be done
-    /// per element per validated instance. Since <see cref="ValidationAttributes"/> is init-only, the
-    /// outcome is stable and computed only once. See the note on
-    /// <see cref="ClassMapping.PropertyMappingsInternal"/> for the lazy initialization idiom used here.</remarks>
-    internal ValidatingFhirModelAttribute[] MandatoryCardinality =>
-        _mandatoryCardinality ?? LazyInitializer.EnsureInitialized(ref _mandatoryCardinality, computeMandatoryCardinality)!;
+    /// <remarks>Determined when <see cref="ValidationAttributes"/> is set - see the remarks there.</remarks>
+    internal ValidatingFhirModelAttribute[] MandatoryCardinality => _mandatoryCardinality;
 
-    private ValidatingFhirModelAttribute[]? _mandatoryCardinality;
-
-    private ValidatingFhirModelAttribute[] computeMandatoryCardinality() =>
-        ValidationAttributes.OfType<CardinalityAttribute>().SingleOrDefault() is { Min: > 0 } cardinality
-            ? [cardinality]
-            : [];
+    private ValidatingFhirModelAttribute[] _mandatoryCardinality = [];
 
     /// <summary>
     /// For a bound element, this is the name of the binding.
--- a/src/Hl7.Fhir.Base/Introspection/PropertyMapping.cs
+++ b/src/Hl7.Fhir.Base/Introspection/PropertyMapping.cs
@@ -290,6 +290,24 @@ public class PropertyMapping : IElementDefinitionSummary
     /// </summary>
     public ValidatingFhirModelAttribute[] ValidationAttributes { get; init; } = [];
 
+    /// <summary>
+    /// The <see cref="CardinalityAttribute"/> declared on this element when it has a minimum cardinality
+    /// higher than 0, as a (single-entry) list of attributes to validate; an empty list otherwise.
+    /// </summary>
+    /// <remarks>Validating an object means checking, for every one of its elements, whether a mandatory
+    /// element is missing - so this scan of <see cref="ValidationAttributes"/> would otherwise be done
+    /// per element per validated instance. Since <see cref="ValidationAttributes"/> is init-only, the
+    /// outcome is stable and computed only once. See the note on
+    /// <see cref="ClassMapping.PropertyMappingsInternal"/> for the lazy initialization idiom used here.</remarks>
+    internal ValidatingFhirModelAttribute[] MandatoryCardinality =>
+        _mandatoryCardinality ?? LazyInitializer.EnsureInitialized(ref _mandatoryCardinality, computeMandatoryCardinality)!;
+
+    private ValidatingFhirModelAttribute[]? _mandatoryCardinality;
+
+    private ValidatingFhirModelAttribute[] computeMandatoryCardinality() =>
+        ValidationAttributes.OfType<CardinalityAttribute>().SingleOrDefault() is { Min: > 0 } cardinality
+            ? [cardinality]
+            : [];
 
     /// <summary>
     /// For a bound element, this is the name of the binding.
--- a/src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
+++ b/src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
@@ -279,7 +279,7 @@ public class BaseFhirJsonDeserializer
     {
         if (Settings.Validator is null) return;
 
-        var context = new PocoValidationContext(target, _inspector, state.Path.PathProducer,
+        var context = new PocoValidationContext(target, _inspector, state.Path.GetInstancePath,
                 line, pos, Settings.NarrativeValidation)
         {
             MemberName = metadata.PropertyMapping.NativeProperty?.Name
@@ -303,7 +303,7 @@ public class BaseFhirJsonDeserializer
         if(Settings.Validator is null) return;
 
         var context =
-            new PocoValidationContext(poco, _inspector, state.Path.PathProducer, line, pos, Settings.NarrativeValidation)
+            new PocoValidationContext(poco, _inspector, state.Path.GetInstancePath, line, pos, Settings.NarrativeValidation)
             {
                 MemberName = memberName
             };
--- a/src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
+++ b/src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
@@ -188,7 +188,7 @@ public class BaseFhirJsonDeserializer
         // postponed until after all properties have been seen (e.g. Instance and Property validations for
         // primitive properties, since they may be composed from two properties `name` and `_name` in json
         // and should only be validated when both have been processed, even if megabytes apart in the json file).
-        state.GetObjectContext().RunDelayedValidation();
+        runDelayedValidations(state);
         state.LeaveObjectContext();
 
         // read past object, unless this is the last EndObject in the top-level Deserialize call
@@ -279,10 +279,7 @@ public class BaseFhirJsonDeserializer
     {
         if (Settings.Validator is null) return;
 
-        var elementName = metadata.PropertyMapping.Name;
-        var runDelayed = metadata.ValueMapping.IsFhirPrimitive;
-
-        var context = new PocoValidationContext(target, _inspector, state.Path.GetInstancePath,
+        var context = new PocoValidationContext(target, _inspector, state.Path.PathProducer,
                 line, pos, Settings.NarrativeValidation)
         {
             MemberName = metadata.PropertyMapping.NativeProperty?.Name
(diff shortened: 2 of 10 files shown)


src/Hl7.Fhir.Base/Introspection/PropertyMapping.cs
```diff
@@ -290,6 +290,24 @@ ClassMapping resolve(Type ft) =>
...
+    /// The <see cref="CardinalityAttribute"/> declared on this element when it has a minimum cardinality
+    /// higher than 0, as a (single-entry) list of attributes to validate; an empty list otherwise.
+    /// </summary>
+    /// <remarks>Validating an object means checking, for every one of its elements, whether a mandatory
+    /// element is missing - so this scan of <see cref="ValidationAttributes"/> would otherwise be done
+    /// per element per validated instance. Since <see cref="ValidationAttributes"/> is init-only, the
+    /// outcome is stable and computed only once. See the note on
+    /// <see cref="ClassMapping.PropertyMappingsInternal"/> for the lazy initialization idiom used here.</remarks>
+    internal ValidatingFhirModelAttribute[] MandatoryCardinality =>
+        _mandatoryCardinality ?? LazyInitializer.EnsureInitialized(ref _mandatoryCardinality, computeMandatoryCardinality)!;
```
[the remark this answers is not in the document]
cknaap: Addressed in ac8156c, by taking the third option: the snapshot is made explicit, deterministic and tested.

You are right that the doc comment overclaimed - init-only makes the *reference* stable, not the contents, so `Min` can be mutated afterwards. What it does not do is introduce a new class of staleness: `IsMandatoryElement` is public, init-only, and has always been computed once in `TryCreate` from the very same `CardinalityAttribute` (`IsMandatoryElement = cardinalityAttr?.Min > 0`). It drives `IElementDefinitionSummary.IsRequired` and `ElementMetadataFilter`, so on `develop` a post-hoc `Min = 1` already gives you an element that the summary API and the serialization filters call optional while the validator calls it mandatory. Mapping metadata is a snapshot of the attributes it was built from; this element was simply the one place still reading through to the attribute.

What was genuinely worth fixing is *when* the value froze. The lazy version froze it at the first validation, which is arbitrary. It is now derived in the `init` accessor of `ValidationAttributes`, so it is fixed at mapping construction - the same moment `IsMandatoryElement` is - for every route that sets the attributes, including the custom-mapping constructor and hand-built mappings. The `ClassMapping.MandatoryElements` list layered on top of it is a pure function of the frozen mappings and is already invalidated by `PropertyMappingCollection.clearCaches()` whenever the collection changes.

The remarks on `ValidationAttributes` now state the snapshot semantics rather than claiming immutability, and two tests in `CustomClassMappingTests` pin the contract: that the mandatory elements are derived from the attributes the mapping was built with (and are what the validator reports on), and that mutating a `CardinalityAttribute` handed to a mapping afterwards is not observed - consistent with `IsMandatoryElement`, which is asserted alongside it.

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
```diff
@@ -279,68 +279,116 @@ private void doPropertyValidation(Base target, string serializedName, object pro
...
 
-            state.Errors.Add(Settings.Validator.ValidateObject(poco, classMapping, context));
+    /// <summary>
+    /// A validation that has been scheduled to run once the enclosing object has been deserialized
+    /// completely (see <see cref="ObjectParsingState.ScheduleDelayedValidation"/>).
+    /// </summary>
+    /// <remarks>This is a struct, and the validations are dispatched by <see cref="Run"/> rather than by a
+    /// delegate, so that scheduling one costs nothing: FHIR primitives - which is what delayed validation is
+    /// for - make up the bulk of the elements of a resource, and every one of them would otherwise allocate a
+    /// closure plus a delegate for a validation that (nearly) always succeeds.</remarks>
```
ewoutkramer: @copilot - But now, instead of allocating a compiler-optimized closure, we're allocating a DelayedValidation structure? what's the difference?  Isn't this structure just an explicitly declared closure?

src/Hl7.Fhir.Base/Serialization/PathPart.cs
```diff
@@ -63,19 +74,25 @@ public PathPart ExitElement()
...
     public abstract string GetInstancePath();
 
+    /// <summary>
+    /// A delegate producing <see cref="GetInstancePath"/> for this position, for use by the validators.
+    /// </summary>
+    /// <remarks>A path is only ever materialized when a diagnostic is actually produced, so validation is
+    /// handed this producer rather than a path. Creating it once per part - instead of once per validation -
+    /// is what keeps it out of the per-element allocation budget of a parse: every validation done at this
+    /// position (at least a property and an object validation) shares this single delegate.</remarks>
+    public Func<string> PathProducer => _pathProducer ??= GetInstancePath;
```
ewoutkramer: I don't see how copying a function's address into a member speeds up anything at all. At max this caches a vtable lookup, but that's sounds like a micromirco optimalization, offset by the extra property access?

src/Hl7.Fhir.Base/Validation/FhirAttributeValidator.cs
```diff
@@ -91,50 +91,72 @@ public virtual IReadOnlyCollection<CodedValidationException> ValidateProperty(
...
+    /// common case is a handful of attributes that all pass, so this avoids allocating anything at all
+    /// (no enumerators, no result collection) unless a validation actually fails.</remarks>
     private static IReadOnlyCollection<CodedValidationException> runAttributeValidation(
         object? candidateValue,
         ValidatingFhirModelAttribute[] attributes,
-        PocoValidationContext validationContext) =>
-        attributes.SelectMany(vfma => vfma.Validate(candidateValue, validationContext)).ToArray();
+        PocoValidationContext validationContext)
+    {
+        List<CodedValidationException>? errors = null;
```
ewoutkramer, line 141: Can you double check this optimization is worth it? It's by far less readable than the original SelectMany.

src/Hl7.Fhir.Base/Serialization/PathPart.cs
cknaap: In short: capturing the `GetInstancePath` instance method into a `Function<string>` is where the cost is. And that happens _per validation_, potentially multiple times per element.

The long explanation by Claude:

The premise is the part worth correcting: nothing here is a function address, and no vtable lookup is being cached. `Func<string>` is a heap object, and what the field caches is an allocation.

`state.Path.GetInstancePath` - the old call sites - is a method group conversion. `GetInstancePath` is an *instance* method, so the delegate has to carry the receiver, which means the compiler cannot hoist it into a static cache: Roslyn's method-group caching only fires for *static* method groups. Every execution of that expression emits a `newobj` for a fresh `MulticastDelegate`, 64 bytes on x64 (header + `_target`, `_methodBase`, `_methodPtr`, `_methodPtrAux`, `_invocationList`, `_invocationCount`). It sat in `doPropertyValidation` and `doObjectValidation`, so that was one delegate per property and one per object of the parse - for a path that is only ever materialized if a diagnostic is actually produced, which I checked: every remaining `GetInstancePath()` call is inside an error branch, and `COVE.Initialize` invokes the producer only while constructing an exception.

For a scalar element, `EnterElement` allocates one `ElementPathPart`, and both the object validation of the primitive value and the property validation run at that same part: 2 x 64 B becomes 1 x 64 B, against +8 B for the field. Net -56 B per scalar element. And where the validator is off, or a part is only used to build an error string, the field stays null and we allocate nothing where `develop` allocated one delegate per call.

On the extra property access: `_pathProducer ??= GetInstancePath` on the hit path is a field load, a null test and an inlined return, with a branch the predictor gets right after the first pass. Weighed against a 64-byte gen0 allocation it isn't close. You're right that creating a delegate from a virtual method emits `ldvirtftn` - but the parts are `sealed` now so the JIT devirtualizes it, and that was never the reason for the change.

Two things I'll concede rather than defend:

- **The win is uneven.** It is 2:1 on scalar elements. On repeating elements `SetIndex` creates a fresh `IndexPathPart` per item, so each item gets one validation on its own part and the cache saves nothing there while still paying the 8 bytes. It's a win because scalars dominate, not uniformly.
- **This is not where the numbers come from.** The 57% drop in the validator's marginal allocation is carried by `MandatoryElements`, `DelayedValidation` and the enumerator removal. This is a two-line change that removes a real per-validation allocation and is provably not slower, but it isn't the headline.

The record->class conversion that came with it is not optional, though: `SetIndex` used the record copy-constructor, which would have copied a cached producer onto a part representing a *different* position. That's a correctness constraint on the caching rather than a style preference.

The obvious alternative - have `PocoValidationContext` hold the `PathPart` and drop the delegate entirely - is closed off: `PathProducer` is public API on a public record, and `PathPart` is internal to `Hl7.Fhir.Serialization`.

src/Hl7.Fhir.Base/Validation/FhirAttributeValidator.cs
cknaap: Fair ask - so I measured it in isolation. Put the `SelectMany` back on top of the rest of the branch and re-ran `ValidatingDeserializationBenchmarks` (R5, net8.0, Release):

| Patient SD (88 KB), validating | `SelectMany` | `foreach` | Δ |
|---|---|---|---|
| allocated | 3,344.8 KB | 2,608.8 KB | **-736 KB** |
| validator's marginal cost | 2,365.5 KB | 1,629.5 KB | **-31%** |

The non-validating variants came out bit-identical in both runs, so only the validation path moved. That is ~28% of everything this PR saves, out of one method - it runs on every property *and* every object, and the old line allocated a display class, a delegate, a `SelectMany` iterator and an enumerator per attribute (~192 B fixed) before a single error existed. `Validate` already returns `IReadOnlyCollection<>`, so all of that was spent rediscovering a `Count` that was already there.

You are right about the readability, though, and it was free to fix. `ValidateObject` already had a local `addRange` helper for exactly this pattern, so I hoisted it to a private method and both now read the same way (29e419e):

```csharp
foreach (var attribute in attributes)
    addRange(ref errors, attribute.Validate(candidateValue, validationContext));
```

Allocations unchanged to the byte (2,608.8 KB), `Hl7.Fhir.Support.Poco.Tests` green (619).

src/Hl7.Fhir.Base/Serialization/BaseFhirJsonDeserializer.cs
cknaap: Conceptually yes - the same captured data, explicitly declared. But I measured it, and the interesting part is not where I expected. I put the closure/`Action` version back on top of the rest of the branch (keeping the lazy dictionary, so only this one thing differs) and re-ran `ValidatingDeserializationBenchmarks`:

| Patient SD (88 KB) | closure | struct | Δ |
|---|---|---|---|
| validating | 2,654.5 KB | 2,608.8 KB | -45.7 KB (-1.7%) |
| validator off | 1,240.2 KB | 979.3 KB | **-260.8 KB (-21%)** |

On the validating path it is nearly a wash, and your scepticism is half right: the struct is 7 fields stored inline in the dictionary (~56 B per entry) against an 8-byte `Action` reference, and those fatter entries claw back most of what the display class and the delegate saved.

The real win is on the path where the validator is switched off entirely - which at first looks impossible, since nothing is scheduled and the dictionary is never even created. The cause is that the display class is allocated at *method entry*, before the `if (Settings.Validator is null) return;` guard, because captured parameters have to be copied into it on entry. So on `develop` every property and every object of a parse allocated a closure even when nobody had asked for validation at all. That also accounts for a good part of the "-31% with the validator off" line in the description.

Net: worth keeping, but for a different reason than the remark on the struct claimed - corrected in cc4556a.

src/Hl7.Fhir.Base/Serialization/PathPart.cs
ewoutkramer: > This is not where the numbers come from. The 57% drop in the validator's marginal allocation is carried by MandatoryElements, DelayedValidation and the enumerator removal. This is a two-line change that removes a real per-validation allocation and is provably not slower, but it isn't the headline.

Right, we agree on that. I honestly think this optimization is so far beyond the comprehension of those who will have to maintain this code, that I want to challenge if it is  worth keeping - if the gain is marginal I prefer maintainability here.

src/Hl7.Fhir.Base/Validation/FhirAttributeValidator.cs
ewoutkramer: Brilliant, thanks.

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/Serialization/PathPart.cs
cknaap: Agreed - reverted in 403f7c3.

I measured what it cost before pulling it, so the trade is on the record: the Patient StructureDefinition goes from 2,608.8 KB to 2,724.6 KB, and the CodeSystem from 84.9 KB to 88.0 KB. So 4.3% of the parse, against a construct whose safety argument is "the parts are immutable, and `SetIndex` must not copy the cache onto a part that represents a different position". That is a lot of reasoning to carry for 4%, and the failure mode if someone gets it wrong later is a wrong path in a diagnostic - quiet, and not something a test would obviously catch. Your call is the right one.

`PathPart.cs` and `BaseFhirXmlDeserializer.cs` are byte-identical to `develop` again. The record-to-class conversion went with it, since making the cache safe was its only purpose.

The rest of the PR is unaffected: `MandatoryElements`, `DelayedValidation`, the attribute-validation loop and the lazy dictionaries all stand, and the validator-off figure does not move at all (979.3 KB either way). I will update the numbers in the description once a full run finishes.

Test suites still green on the revert: `Hl7.Fhir.Serialization.R5.Tests` (16,630) and `Hl7.Fhir.Support.Poco.Tests` (619).

ewoutkramer (approved):

PR #3575, opened 2026-08-12
Tweak dynamic type inference logic for NewPocoBuilder

andrzejskowronski:
This pull request improves the logic for determining whether a FHIR type should be treated as a primitive or complex type, especially for dynamically built POCOs such as logical models, and adds new tests to verify this behavior. The main changes are grouped below:

### Improved Type Determination Logic

* Refactored the `determineBestDynamicMappingForElement` method in `NewPocoBuilder.TypedElement.cs` to use FHIR definitions (when available) to distinguish between primitive and complex types, rather than relying solely on type name heuristics. This helps correctly handle logical models identified by canonical URLs and prevents misclassification of complex types as primitives.
* Added helper methods `isPrimitiveType` (checks for "value" child with XML attribute representation) and `isPrimitiveTypeName` (uses lowercase naming heuristic, but excludes canonical URLs) to improve primitive type detection.
* Updated imports in `NewPocoBuilder.TypedElement.cs` to include `Hl7.Fhir.Specification` for new logic.

### New Tests for Logical Model Handling

* Added a new test class `LogicalModelToPocoTests` to verify that logical model roots (identified by canonical URL or type name) are always built as complex types and not as primitives, and that FhirPath evaluation works correctly on these dynamically built POCOs.

Encountered when working with CDS hooks being parsed by legacy stack, FHIRPath of `%resource.hook` would return the value correctly, but `%resource.hook = 'order-sign'` would fail.

Poco parsers did not exhibit the issue as they go over the json property type directly rather than working with just `ISourceNode.Text`

### DIFF
--- a/src/Hl7.Fhir.Base/ElementModel/NewPocoBuilder.TypedElement.cs
+++ b/src/Hl7.Fhir.Base/ElementModel/NewPocoBuilder.TypedElement.cs
@@ -191,17 +191,16 @@ internal partial class NewPocoBuilder
         if (node.Definition?.Type is [IStructureDefinitionSummary summary])
             return isPrimitiveType(summary)
                 ? DetermineBestPrimitiveMapping()
-                : isResource(summary.IsResource) ? ClassMapping.DynamicResource : ClassMapping.DynamicDataType;
+                : summary.IsResource || isResource(node) ? ClassMapping.DynamicResource : ClassMapping.DynamicDataType;
 
         if (node.InstanceType is { } it && isPrimitiveTypeName(it))
             return DetermineBestPrimitiveMapping();
 
-        return isResource(false) ? ClassMapping.DynamicResource : ClassMapping.DynamicDataType;
+        return isResource(node) ? ClassMapping.DynamicResource : ClassMapping.DynamicDataType;
 
-        bool isResource(bool typeIsResource) =>
-            typeIsResource
-            || node.Annotation<IResourceTypeSupplier>()?.ResourceType is not null
-            || node.Definition?.IsResource is true;
+        static bool isResource(ITypedElement node) =>
+            node.Definition?.IsResource is true
+            || node.Annotation<IResourceTypeSupplier>()?.ResourceType is not null;
 
         ClassMapping DetermineBestPrimitiveMapping()
         {
--- a/src/Hl7.Fhir.Base/ElementModel/NewPocoBuilder.TypedElement.cs
+++ b/src/Hl7.Fhir.Base/ElementModel/NewPocoBuilder.TypedElement.cs
@@ -139,15 +139,34 @@ internal partial class NewPocoBuilder
     /// Determines whether the type described by this summary is a primitive: in FHIR, those are the types
     /// with a "value" child, which is represented as an attribute in Xml.
     /// </summary>
+    /// <remarks>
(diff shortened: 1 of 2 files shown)


src/Hl7.Fhir.Base/ElementModel/NewPocoBuilder.TypedElement.cs
```diff
@@ -134,18 +135,54 @@ private ClassMapping classMappingForElement(ITypedElement node, PropertyMapping?
...
+    /// <summary>
+    /// Determines whether the type described by this summary is a primitive: in FHIR, those are the types
+    /// with a "value" child, which is represented as an attribute in Xml.
+    /// </summary>
+    private static bool isPrimitiveType(IStructureDefinitionSummary summary)
+    {
+        foreach (var element in summary.GetElements())
+        {
+            if (element.ElementName == "value" && element.Representation == XmlRepresentation.XmlAttr)
+                return true;
```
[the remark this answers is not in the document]
andrzejskowronski: Adjusted to presence of .value with XmlAttr is still definitive primitive, but absence still leads to type checks

src/Hl7.Fhir.Base/ElementModel/NewPocoBuilder.TypedElement.cs
```diff
@@ -134,18 +135,73 @@ private ClassMapping classMappingForElement(ITypedElement node, PropertyMapping?
...
+
+        if (node.InstanceType is { } it && isPrimitiveTypeName(it))
             return DetermineBestPrimitiveMapping();
 
-        if (node.Annotation<IResourceTypeSupplier>()?.ResourceType is not null || node.Definition?.IsResource is true)
-            return ClassMapping.DynamicResource;
+        return isResource(false) ? ClassMapping.DynamicResource : ClassMapping.DynamicDataType;
 
-        return ClassMapping.DynamicDataType;
+        bool isResource(bool typeIsResource) =>
```
ewoutkramer: The parameter appears to be over-engineered for symmetry. The author was probably thinking: "I'll extract this into a helper that can handle both the 'we know from the definition' case and the 'we don't know' case." But in practice:

When there's a definition (line 194), you already know if it's a resource from summary.IsResource — passing it as a parameter and returning early is just shortcutting what you'd determine anyway from the other checks.
When there's no definition (line 199), passing false does nothing.
A simpler version would be:

C#
bool isResource() =>
    node.Annotation<IResourceTypeSupplier>()?.ResourceType is not null
    || node.Definition?.IsResource is true;
And then at line 194:

C#
: isResource() || summary.IsResource ? ClassMapping.DynamicResource : ClassMapping.DynamicDataType;
This removes the parameter and makes the intent clearer: "either we determined it's a resource, or the definition told us it is."

ewoutkramer (changes requested):

src/Hl7.Fhir.Base/ElementModel/NewPocoBuilder.TypedElement.cs
andrzejskowronski: Fair, tweaked with the order of checks inverted as asking summary/definition will be a bit less costly than checking annotations collection

ewoutkramer (approved):

